PR #5, opened 2020-11-08
[FEAT](UNIRIS): BeaconChain P2P availability sampling

bacharif:
- setting up the ping function


bacharif:
Issue with the test

lib/uniris/beacon_chain/subset.ex
```diff
@@ -84,9 +84,9 @@ defmodule Uniris.BeaconChain.Subset do
       ) do
     listNodes =
       Enum.filter(P2P.list_nodes(), fn x -> :binary.part(x.first_public_key, 0, 1) == subset end)
-      Logger.info("List Node : ", listNodes)
 
-    _p2p_view_available = Enum.map(listNodes, fn x -> GenICMP.ping(x) end)
+    _p2p_view_available = Enum.map(listNodes, fn x -> GenICMP.ping(x.ip) end)
```
samuel-uniris, line 88: I think you should use Task.async_stream to make the ping requests concurrently.
Otherwise it will do it sequentially, one after another.

test/uniris/beacon_chain/gen_icmp_test.exs
```diff
@@ -0,0 +1,53 @@
...
+  alias Uniris.P2P
+  alias Uniris.P2P.Node
+
+  doctest GenICMP
+
+  test "ping/1 should ping the destination node" do
+    assert {:ok, %{data: <<222, 173, 190, 239>>, id: 0, seq: 0}} = GenICMP.ping({127, 0, 0, 1})
+  end
+
+  test "should fetch the transaction locally when the current node is a storage node" do
```
samuel-uniris, line 14: You should change the test name

samuel-uniris (changes requested):

lib/uniris/beacon_chain/subset.ex
```diff
@@ -85,11 +85,11 @@ defmodule Uniris.BeaconChain.Subset do
     listNodes =
       Enum.filter(P2P.list_nodes(), fn x -> :binary.part(x.first_public_key, 0, 1) == subset end)
 
-    _p2p_view_available = Enum.map(listNodes, fn x -> GenICMP.ping(x.ip) end)
+    p2p_view_available = Enum.map(listNodes, fn x -> GenICMP.ping(x.ip) end)
 
     new_registry = SlotRegistry.seal_current_slot(slot_registry, slot_time)
 
-    {:noreply, %{state | slot_registry: new_registry}}
+    {:noreply, %{state | slot_registry: new_registry, p2p_view_available: p2p_view_available}}
```
samuel-uniris, line 92: You have to use `Map.put/3` to create the new state, as the `p2p_view_available` field is not existing yet inside the state map. This notation will not work.
This will work:


```elixir
new_state = state
    |> Map.put(:slot_registry, new_registry)
    |> Map.put(:p2p_view_available, p2p_view_available)
```

samuel-uniris (changes requested):
Note: ICMP calls require the host be running on sudo or to use `setcap` to allow the erlang VM to use this protocol. Moreover some routers or firewalls may block ICMP port.

Another approach could be to leverage the P2P opening port of the targeting node and its protocol by using for example TCP to create a healthcheck endpoint. What do you think ?

PR #11, opened 2020-12-14
Fix types

88maomao:
Introduce dialyzer for type verification


samuel-uniris:
Thanks for the PR.
Seems good to have those checks.

Do you think it will possible to add it inside the git hooks ? So only successful type specifications will make a valid git commit


(config.exs)
```
use Mix.Config

config :git_hooks,
  auto_install: true,
  verbose: true,
  hooks: [
    pre_commit: [
      tasks: [
        "mix format",
        "mix clean",
        "mix compile --warnings-as-errors",
        "mix credo --strict",
        "mix test --trace"
      ]
    ]
  ]
```

pvmart:
It would be as easy as adding "mix dialyzer" to git_hooks configuration. But it is not free. First time it runs it generates [PLT](https://github.com/jeremyjh/dialyxir#plt) files which takes around 15 minutes on 6 core 2.2GHz linux box. After that each run takes around 1 to 3 minutes.

samuel-uniris:
Ok thanks, I will keep in mind.
Maybe we could later on try to reduce the analysis time by ignoring some app maybe. To see....

PR #12, opened 2020-12-17
Update and clean

pvmart:
An attempt to simplify dev workflow


samuel-uniris:
Thanks.
It should improve the workflow and make the trick for the CI proposal job.

PR #14, opened 2021-01-14
Created Networking module; upnpv1, v2, natpmp; basic metrics reporter

dmytronasyrov:
add upnpv1, upnpv2, natpmp, statsd reporter;
refactor Networking module


samuel-uniris:
Could you make different PR for networking modules and metrics ?

lib/uniris/db/cassandra_impl.ex
```diff
@@ -23,7 +23,7 @@ defmodule Uniris.DB.CassandraImpl do
   """
   @spec start_link(Keyword.t()) :: {:ok, pid()}
   def start_link(opts \\ []) do
-    nodes = Keyword.get(opts, :nodes, ["127.0.0.1:9042"])
+    nodes = Keyword.get(opts, :nodes, ["cassandra.gambitstream.com:31300"])
```
samuel-uniris, line 26: I think you forgot to undo this change

lib/uniris/networking/networking.ex
```diff
@@ -12,19 +12,23 @@ defmodule Uniris.Networking do
...
-  def get_p2p_port do
-    Config.load_from_sys_env?
-    |> case do
-      {:error, reason} -> {:error, reason}
-      {:ok, false} -> Config.p2p_port_from_config()
-      {:ok, true} -> Config.p2p_port_from_env()
-    end
-  end
+  @spec get_p2p_port() :: {:ok, pos_integer} | {:error, :invalid_port}
+  defdelegate get_p2p_port, to: Config
```
samuel-uniris, line 33: I think it should publish the port on UPnP (ie. port mapping) to make it available if it's supported, and just get it from Application config

mix.exs
```diff
@@ -14,7 +14,7 @@ defmodule Uniris.MixProject do
       deps: deps(),
       compilers: [:elixir_make, :phoenix] ++ Mix.compilers(),
       elixirc_paths: elixirc_paths(Mix.env()),
-      elixirc_options: [warnings_as_errors: true],
+      elixirc_options: [warnings_as_errors: false],
```
samuel-uniris, line 17: I think you forgot to undo this change

lib/uniris/networking/ip_lookup/ip_lookup.ex
```diff
@@ -1,16 +1,16 @@
...
 
-  alias __MODULE__.Config
-
   # Public
 
-  @spec get_node_ip() :: {:ok, :inet.ip_address()} | {:error, binary}
+  @spec get_node_ip() :: {:ok, :inet.ip_address()} | {:error, :invalid_ip_provider | :not_recognizable_ip | :ip_discovery_error}
   def get_node_ip do
-    with {:ok, ip_provider} <- Config.ip_provider() do
+    with config <- Application.get_env(:uniris, Uniris.Networking),
```
samuel-uniris, line 8: One suggestion could be if the provider is not the Static one (defined IP address), try to use Nat(UpnP) otherwise fallback to IPFY maybe.

samuel-uniris (changes requested):

config/dev.exs
```diff
@@ -72,7 +69,8 @@ config :uniris, Uniris.Crypto.Keystore, impl: Uniris.Crypto.SoftwareKeystore
 config :uniris, Uniris.Crypto.SoftwareKeystore,
   seed: System.get_env("UNIRIS_CRYPTO_SEED", "node1")
 
-config :uniris, Uniris.DB, impl: Uniris.DB.KeyValueImpl
+config :uniris, Uniris.DB, impl: Uniris.DB.CassandraImpl
```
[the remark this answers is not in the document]
dmytronasyrov: true

lib/uniris/db/cassandra_impl.ex
dmytronasyrov: true

mix.exs
dmytronasyrov: true

PR #16, opened 2021-01-17
Add UPnP v1, v2 and NAT PMP to networking module

dmytronasyrov:
Add UPnP v1, v2 and NAT PMP to networking module


lib/uniris/networking/networking.ex
```diff
@@ -0,0 +1,29 @@
...
+  Provides current host IP address.
+  """
+  @spec get_node_ip() :: {:ok, :inet.ip_address()} | {:error, :invalid_ip_provider | :not_recognizable_ip | :ip_discovery_error}
+  defdelegate get_node_ip, to: IPLookup
+
+  @doc """
+  Provides P2P port number.
+  """
+  @spec get_p2p_port() :: {:ok, pos_integer} | {:error, :invalid_port | :port_unassigned}
+  def get_p2p_port do
```
samuel-uniris, line 20: I think the best should be to:
1. Get port from config if possible
2. Then try to publish it
3. If not, use random port and try to publish

lib/uniris/networking/ip_lookup/ip_lookup.ex
```diff
@@ -0,0 +1,28 @@
+defmodule Uniris.Networking.IPLookup do
+  @moduledoc false
+
+  alias __MODULE__.{Nat, Ipify}
+
+  # Public
+
+  @spec get_node_ip() :: {:ok, :inet.ip_address()} | {:error, :invalid_ip_provider | :not_recognizable_ip | :ip_discovery_error}
+  def get_node_ip do
+    with config when not is_nil(config) <- Application.get_env(:uniris, Uniris.Networking),
```
samuel-uniris, line 10: I propose to change a bit the logic. 

Instead of getting external IP and support switch when there is not config, get the provider implementation (default to NAT), if NAT try to fetch it, otherwise switch to IPFY. 
For instance through pattern matching.

So with this scheme, each ip provider could have its fallback resolution.

samuel-uniris (changes requested):

lib/uniris/networking/networking.ex
dmytronasyrov: ok

lib/uniris/networking/ip_lookup/ip_lookup.ex
dmytronasyrov: Not sure I understand the difference with the current logic.

lib/uniris/networking/networking.ex
dmytronasyrov: I'm not sure it will work in a cloud. Where there is no UPnP available and you need just a strict port. In such a case steps 2 and 3 will mess with the security group/ACL settings of the instance.

samuel-uniris: We could use switch to identify the strategy to do given the ip provider.
If NAT, try the UPnP port publication (default to the one inside the conf)
if Static, just use the port
if IPFY, just use the port (maybe to test if the port is opened otherwise it's an error)

Those are just example of adaptative behaviors

lib/uniris/networking/ip_lookup/ip_lookup.ex
samuel-uniris: How I understand the current logic is like:
- If conf is nil -> get external ip
- If conf is present, fetch from the provider

 A proposition could be:
 
 ```elixir
def get_node_ip do
  conf = Application.get_env(:uniris, Uniris.Networking)
  ip_provider = Keyword.get(config, :ip_provider, Nat)
  do_get_node_ip(ip_provider)
end
 
 defp do_get_node_ip(Nat) do
    case Nat.get_node_ip do
       {:ok, ip} ->
          {:ok, ip}
      {:error, _} ->
          IPFY.get_node_ip()
    end
end

defp do_get_node_ip(IPFY), do: IPFY.get_node_ip()
defp do_get_node_ip(Static), do: Static.get_node_ip()
 ```
 
 Where each provider can define, if needed,  a fallback method

lib/uniris/networking/networking.ex
dmytronasyrov: I propose the next algorithm:
  1. No port in config && UPnP or NAT PMP is available -> get random port from the pool.
  2. Port in config && UPnP or NAT PMP is available - try to publish port from config.
  3. Port in config && Unable to publish port from config && UPnP or NAT PMP is available - get random port from the pool.
  4. Port in config && UPnP or NAT PMP not available -> return port from config.
  5. No port in config && UPnP or NAT PMP not available -> find open port.
  6. No port in config && open port not found && UPnP or NAT PMP not available -> return error.

lib/uniris/networking/ip_lookup/ip_lookup.ex
dmytronasyrov: I propose the next algorithm:
  1. No provider in config -> use NAT discovery.
  2. No provider in config && NAT discovery failed -> use IPIFY.
  3. No provider in config && NAT discovery failed && IPIFY failed (for any reason) -> use hostname from config.
  4. No provider in config && NAT discovery failed && IPIFY failed (for any reason) && no hostname in config -> use localhost.
  5. Provider in config -> use provider.
  6. Provider in config failed -> Goto 1.

lib/uniris/networking/networking.ex
samuel-uniris: Some of those points are good but I don't think the case when the port is missing in the conf will really happen. 

The default conf will rarely change in production. 

So we can assume than the port should be present otherwise it will crash (as a specification in the let's crash philosophy)

lib/uniris/networking/ip_lookup/ip_lookup.ex
samuel-uniris: 1) I think it should not really happen as the default will be NAT
2) I think it should be: NAT discovery failed -> use IPFY
3) not sure if the right fallback from IPFY should be an hostname from config. As it's not the expected behavior of auto IP lookup provided by NAT/IPFY implementation
4) I think the localhost should be used only for dev, or private network but this should be the responsibility of a configuration from the Static provider

lib/uniris/networking/networking.ex
samuel-uniris: We could also log the fact than the port is not available and request the user to open manually the port if not possible

lib/uniris/networking/networking.ex
```diff
@@ -9,21 +9,54 @@ defmodule Uniris.Networking do
...
+  2. Port in config && Unable to publish port from config && UPnP or NAT PMP is available - get random port from the pool.
+  3. Port in config && UPnP or NAT PMP not available -> return error :port_unassigned.
   """
   @spec get_p2p_port() :: {:ok, pos_integer} | {:error, :invalid_port | :port_unassigned}
   def get_p2p_port do
-    with {:error, _reason} <- IPLookup.Nat.get_random_port(),
-    {:error, :invalid_port} <- Config.get_p2p_port do
-      port = Enum.random(31000..33000)
+    with {:ok, port_to_open} <- Config.get_p2p_port,
+    {:ok, port} <- IPLookup.Nat.open_port(port_to_open) do
```
samuel-uniris, line 39: I think to use the UPnP port publication, it should assume the protocol is the one desired.
If the `Static` provider is given in the conf, we may not want to use UPnP and the port should come from conf.

config/dev.exs
```diff
@@ -14,8 +14,10 @@ config :phoenix, :plug_init_mode, :runtime
 # hostname(string) - (for Static) provides a constant IP address for Static (ex. "127.0.0.1")
 # port(pos_int) - (for Static) provides a P2P port number (ex. 3002)
 #
-# config :uniris, Uniris.Networking, 
-#   port: 4000
+config :uniris, Uniris.Networking, 
+  # ip_provider: Uniris.Networking.IPLookup.Static,
+  # hostname: "127.0.0.1",
+  port: 5454
```
samuel-uniris, line 20: Why did you change the default port ?

samuel-uniris (changes requested):

lib/uniris/networking/ip_lookup/nat.ex
```diff
@@ -31,13 +37,14 @@ defmodule Uniris.Networking.IPLookup.Nat do
...
-  @spec assign_port(list(atom)) :: {:ok, pos_integer} | {:error, any()}
-  defp assign_port([]), do: {:error, :port_unassigned}
-  defp assign_port([protocol_module | protocol_modules]) do
+  @spec assign_port(list(atom), pos_integer) :: {:ok, pos_integer} | {:error, any()}
+  defp assign_port([], _port), do: {:error, :port_unassigned}
+  defp assign_port([protocol_module | protocol_modules], port) do
     with {:ok, router_ip} <- protocol_module.discover(),
-    {:ok, _since, port, _external_port, _} <- protocol_module.add_port_mapping(router_ip, :tcp, 0, 0, 0) do
-      {:ok, port}
+    {:ok, _since, internal_port, _external_port, _} <- protocol_module.add_port_mapping(router_ip, :tcp, port, port, 0) do
```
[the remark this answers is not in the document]
dmytronasyrov: it's the same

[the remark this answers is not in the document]
dmytronasyrov: I mean same numbers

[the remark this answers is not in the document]
samuel-uniris: ok

config/dev.exs
dmytronasyrov: just tested. Config is reverted to 3002 now

samuel-uniris (approved):

samuel-uniris:
Could you use credo to cleanup using `mix credo --strict` to create consistency with the current codebase

samuel-uniris:
Please can you check the remaining conflicts and fix some warnings otherwise the compilation will not passed (due to `--warnings-as-errors`)

samuel-uniris (changes requested):
I still got some issues from credo

image

dmytronasyrov:
I know. But I like "|> case do" construction. I prefer not to use it rather than follow credo's suggestions

samuel-uniris (approved):

PR #17, opened 2021-01-18
add telemetry

dmytronasyrov:
add telemetry


lib/uniris/application.ex
```diff
@@ -43,7 +45,12 @@ defmodule Uniris.Application do
     ]
 
     opts = [strategy: :rest_for_one, name: Uniris.Supervisor]
-    Supervisor.start_link(Utils.configurable_children(children), opts)
+    res = Supervisor.start_link(Utils.configurable_children(children), opts)
+
+    :telemetry.execute([:uniris, :run_app, :success], %{"latency" => 1.2}, %{"hello" => "uniris"})
```
samuel-uniris, line 50: I think those measurements were here for tests . Maybe a cleanup ..

lib/telemetry/instrumenter.ex
```diff
@@ -0,0 +1,20 @@
+defmodule Uniris.Telemetry.Instrumenter do
```
samuel-uniris, line 1: I think we can leverage TelemetryMetrics here, with a given supervisor to support the right reporter and the list of metrics to support.

samuel-uniris (changes requested):

lib/uniris/application.ex
dmytronasyrov: You mentioned you want a simple example of telemetry usage. That's it.

samuel-uniris: I wanted a simple setup but ok will change later

lib/telemetry/instrumenter.ex
dmytronasyrov: Not sure exactly what you mean, but I think I did it in the correct fashion. I added a polling telemetry events example into the supervision tree. Take a look.

lib/uniris/application.ex
dmytronasyrov: Let me clean these rows. We have a setup and it works as expected.

lib/telemetry/instrumenter.ex
samuel-uniris: Ok good now

samuel-uniris (approved):

PR #22, opened 2021-02-02
Add Coingecko Oracle & Oracle cron

dmytronasyrov:
Add Coingecko Oracle & Oracle cron


lib/uniris/p2p/transport/tcp_impl.ex
```diff
@@ -37,6 +37,8 @@ defmodule Uniris.P2P.Transport.TCPImpl do
   defp recv_loop(socket) do
     case :gen_tcp.recv(socket, 0) do
       {:ok, data} ->
+        IO.puts("MESSAGE RECEIVED: #{inspect(data)}")
```
samuel-uniris, line 40: Forgot to clean

config/dev.exs
```diff
@@ -67,7 +67,9 @@ config :uniris, Uniris.Crypto.Keystore, impl: Uniris.Crypto.SoftwareKeystore
 config :uniris, Uniris.Crypto.SoftwareKeystore,
   seed: System.get_env("UNIRIS_CRYPTO_SEED", "node1")
 
-config :uniris, Uniris.DB, impl: Uniris.DB.KeyValueImpl
+# config :uniris, Uniris.DB, impl: Uniris.DB.KeyValueImpl
```
samuel-uniris, line 70: Forgot to clean

lib/uniris/application.ex
```diff
@@ -24,7 +24,7 @@ defmodule Uniris.Application do
 
   def start(_type, _args) do
     children = [
-      Uniris.Telemetry,
+      # Uniris.Telemetry, # REVERT
```
samuel-uniris, line 27: Forgot to clean

lib/uniris/db/cassandra_impl.ex
```diff
@@ -26,7 +26,9 @@ defmodule Uniris.DB.CassandraImpl do
   """
   @spec start_link(Keyword.t()) :: {:ok, pid()}
   def start_link(opts \\ []) do
-    nodes = Keyword.get(opts, :nodes, ["127.0.0.1:9042"])
+    # nodes = Keyword.get(opts, :nodes, ["127.0.0.1:9042"])
+    # REVERT
+    nodes = Keyword.get(opts, :nodes, ["cassandra.gambitstream.com:31300"])
```
samuel-uniris, line 31: Forgot to clean

lib/uniris/oracles/oracle_server.ex
```diff
@@ -0,0 +1,84 @@
+defmodule Uniris.Oracles.OracleCronServer do
```
samuel-uniris, line 1: If you can change the name to `Scheduler`, to stay consistent with the other cron based modules

lib/uniris/oracles/oracle_server.ex
```diff
@@ -0,0 +1,84 @@
...
+  use GenServer
+
+  alias Uniris.TransactionChain.{
+    Transaction,
+    TransactionData
+  }
+
+  # Public
+
+  @spec start_link(mfa: mfa(), interval: pos_integer()) ::
```
samuel-uniris, line 13: In other schedulers, CronTab intervals are used(i.e 0 * * * * *), so you can leverage it also to help the customization and avoid conflicts in the scheduling

lib/uniris/oracles/supervisor.ex
```diff
@@ -0,0 +1,38 @@
...
+
+  alias Uniris.Utils
+
+  def start_link(args) do
+    Supervisor.start_link(__MODULE__, args, name: __MODULE__)
+  end
+
+  def init(_args) do
+    optional_children = [
+      {OracleCronServer, [mfa: {Coingecko, :fetch, [DateTime.utc_now()]}, interval: 5_000],
```
samuel-uniris, line 19: I would prefer to have a UCO Price Feed Oracle which will leverage Coingecko in its implementation , instead to put directly the implementation detail here

lib/uniris/oracles/supervisor.ex
```diff
@@ -0,0 +1,38 @@
...
+       id: :a},
+      {OracleCronServer,
```
samuel-uniris, line 21: I don't get the use of this second oracle scheduler and moreover with this particular date

samuel-uniris (changes requested):
I think you should keep the oracle supervisor in the application.

lib/uniris/oracles/supervisor.ex
dmytronasyrov: nevermind

lib/uniris/transaction_chain/transaction.ex
```diff
@@ -171,6 +173,13 @@ defmodule Uniris.TransactionChain.Transaction do
     {previous_public_key, next_public_key}
   end
 
+  defp get_transaction_public_keys(:oracle) do
```
samuel-uniris, line 176: I think we should leverage a new derivation algorithm inspired from the beacon chain, to derive key based on the oracle and the time of the transaction. So we will have a cryptographic chain for the specific oracle.

lib/uniris/mining/proof_of_existence.ex
```diff
@@ -0,0 +1,35 @@
...
+    Transaction,
+    TransactionData
+  }
+
+  # Public
+
+  @spec do_proof_of_existence(Transaction.t()) :: boolean()
+  def do_proof_of_existence(tx) do
+    %TransactionData{content: content} = tx.data
+    %TransactionContent{mfa: {m, f, a}, payload: payload} = :erlang.binary_to_term(content)
```
samuel-uniris, line 16: I think we should avoid `binary_to_term` as it's not safe and some malicious code can be injected. We can easily provide a serialization mechanism using binaries and bitstrings adapted for the given oracle.

samuel-uniris (changes requested):

lib/uniris/oracles/oracles/coingecko.ex
```diff
@@ -0,0 +1,29 @@
+defmodule Uniris.Oracles.Coingecko do
```
samuel-uniris, line 1: I think a name as `Uniris.Oracles.UCOPrice.Coingecko` will make more sense as the CoinGecko is not an Oracle by itself

lib/uniris/oracles/oracles/uco_price.ex
```diff
@@ -0,0 +1,25 @@
+defmodule Uniris.Oracles.UcoPrice do
+  @moduledoc false
+
+  alias Uniris.Oracles.Coingecko
+
+  @oracles [Coingecko]
```
samuel-uniris, line 6: Maybe rename it as  `endpoints` or `sources` as Coingecko is more an external source instead of an Oracle

lib/uniris/oracles/oracles/uco_price.ex
```diff
@@ -0,0 +1,25 @@
...
+
+  # Public
+
+  @spec start() :: [Keyword.t()]
+  def start do
+    @oracles
+    |> Task.async_stream(fn m -> m.start() end)
```
samuel-uniris, line 13: You are calling `m.start` so `Coingecko` but the function doesn't exist

lib/uniris/oracles/scheduler.ex
```diff
@@ -0,0 +1,108 @@
...
+
+  @impl GenServer
+  def handle_info(:fetch, state = %{mfa: {m, f, a}}) do
+    args_with_date = [date: DateTime.utc_now()] ++ a
+
+    payload =
+      apply(m, f, args_with_date)
+      |> :erlang.list_to_binary()
+
+    Process.send_after(self(), :send_tx, 0)
```
samuel-uniris, line 68: Why are you sending a message to yourself ? You can leverage directly `Task` for a non blocking private function without blocking the GenServer in the mailbox.

lib/uniris/oracles/oracles/uco_price.ex
```diff
@@ -0,0 +1,25 @@
+defmodule Uniris.Oracles.UcoPrice do
```
samuel-uniris, line 1: I think you should you put this file into `oracles/uco_price.ex` instead of `oracles/oracles/uco_price.ex` because the double `oracles` folder brings confusion.

lib/uniris/oracles/transaction_content.ex
```diff
@@ -0,0 +1,48 @@
...
+  @moduledoc false
+
+  defstruct [
+    :mfa,
+    :payload
+  ]
+
+  def serialize(%__MODULE__{mfa: {m, f, a}, payload: payload}) do
+    m_binary = :erlang.atom_to_binary(m)
+    f_binary = :erlang.atom_to_binary(f)
```
samuel-uniris, line 11: Why do you want to serialize the function name ? As we should have the knowledge based on the Oracle type the function to call (like behavior)

lib/uniris/oracles/oracles/uco_price.ex
```diff
@@ -0,0 +1,25 @@
...
+  @moduledoc false
+
+  alias Uniris.Oracles.Coingecko
+
```
samuel-uniris, line 5: Could be great to create a behavior for Oracle

lib/uniris/transaction_chain/transaction/validation_stamp.ex
```diff
@@ -11,6 +11,7 @@ defmodule Uniris.TransactionChain.Transaction.ValidationStamp do
     :signature,
     :proof_of_work,
     :proof_of_integrity,
+    :proof_of_existence,
```
samuel-uniris, line 14: Regarding this name, I have some concerns
- I don't think we understand clearly what it is 
- For other type of transaction this data will be unrelevant. 
- Proof of XXX => Usually are cryptograhic proofs , and here it's just a boolean

I was thinking to adapt this part for a more generic way, as some transaction may be distributed error detection: smart contract creation, smart contract calls, oracles, ...

Of cource we could use a boolean for `oracle_validation` as `contract_validation` but again it may be irrelevant for other transaction validation

So we could imagine a field in which we will find some custom errors, related to the pending transaction itself but aggregated and validated across the miners. It may be more scalable for further implementations.

lib/uniris/application.ex
```diff
@@ -54,7 +54,8 @@ defmodule Uniris.Application do
       SelfRepairSupervisor,
       WebSupervisor,
       {Bootstrap, port: port, transport: transport},
-      {Task.Supervisor, name: Uniris.TaskSupervisor}
+      {Task.Supervisor, name: Uniris.TaskSupervisor},
+      Uniris.Oracles.Supervisor
```
samuel-uniris, line 58: I think Oracle should start after bootstrap, as we may want to load from DB some previously stored Oracle transaction to fill an internal state

lib/uniris/oracles/scheduler.ex
```diff
@@ -0,0 +1,108 @@
...
+  @impl GenServer
+  def handle_continue(:prepare, state = %{mfa: {m, _, _}}) do
+    m.start()
+
+    Process.send_after(self(), :fetch, 5_000)
+    {:noreply, state}
+  end
+
+  @impl GenServer
+  def handle_call(:get_payload, subscriber, state = %{payload: nil, subscribers: subscribers}) do
```
samuel-uniris, line 51: I think we can leverage `Registry` to support subscribers to avoid timeouts and GenServer bottlenecks using a asynchronous communication mechanism, like in `Uniris.PubSub`.

So contracts will register directly to a topic for example `{:oracle, :uco_price}` and will receive the good message after the payload retrieval.

samuel-uniris (changes requested):

PR #26, opened 2021-02-15
Mutable storage folder

pvmart:
Move all mutable files away from application's `priv` folder.


samuel-uniris:
Everything is ok, except a warning during test.

`warning: unused alias Utils
  test/uniris/crypto_test.exs:7`

If you can fix it, it will be great !
Thanks

PR #39, opened 2021-07-06
Create $INSTALL_DIR if not exists

roychowdhuryrohit-dev:


scripts/install.sh
```diff
@@ -157,7 +157,8 @@ sudo apt install tpm2-tools
 cd $SCRIPT_DIR/..
 make
 
-FILENAME=`cat /sys/class/net/eno1/address`
+MAC=$(ip addr show | awk '/inet.*brd/{print $NF; exit}')
```
samuel-uniris, line 160: Maybe rename the variable, because the command is no the MAC, but the network interface.

samuel-uniris (changes requested):

samuel-uniris (approved):

PR #40, opened 2021-07-07
Use TPM2-TSS Release 3.1 and Minor Fixes

drvarun:
Removed Docker Double Installation (line 75 to 85 were repeated)
Use TPM2-TSS Release 3.1

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -131,8 +131,8 @@ sudo apt -y install \
 
 cd $INSTALL_DIR
 wget https://github.com/tpm2-software/tpm2-tss/releases/download/3.1.0/tpm2-tss-3.1.0.tar.gz
-tar -xf tpm2-tss-3.1.0.tar.gz && rm tpm2-tss-3.1.0.tar.gz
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -142,9 +129,10 @@ sudo apt -y install \
   libcurl4-openssl-dev \
   acl
 
-git clone https://github.com/tpm2-software/tpm2-tss.git $INSTALL_DIR/tpm2-tss
-cd $INSTALL_DIR/tpm2-tss
-./bootstrap
+cd $INSTALL_DIR
+wget https://github.com/tpm2-software/tpm2-tss/releases/download/3.1.0/tpm2-tss-3.1.0.tar.gz
+tar -xf tpm2-tss-3.1.0.tar.gz && rm tpm2-tss-3.1.0.tar.gz
```
samuel-uniris, line 134: I think you can directly extract the archive into a given directory.
This will avoid `mv tpm2-tss-3.1.0 tpm2-tss`

samuel-uniris (changes requested):

samuel-uniris (approved):

PR #42, opened 2021-07-16
Changed host ip to 0.0.0.0 for dev environment

roychowdhuryrohit-dev:


samuel-uniris:
I think just updating the env var should be sufficient. As it may be specific.

roychowdhuryrohit-dev:
Okay thanks. Actually the default behaviour in popular http servers is to host at the 0.0.0.0 undefined address so that it’s accessible from all interfaces.

samuel-uniris:
Yes I know just this static IP is used for the p2p discovery inside node transaction.
For accepting incoming request on the web part , maybe see the Phoenix.Endpoint configuration. https://hexdocs.pm/phoenix/Phoenix.Endpoint.html

roychowdhuryrohit-dev:
Okay I create a new PR with phoenix config.

Issue #49, opened 2021-07-27
TransactionChain.list_transactions_by_type - Expose API 


redDwarf03:
Could you expose please API TransactionChain.list_transactions_by_type
Thx

samuel-uniris:
Hello.
What will be the purpose ? Because this function list the local stored transaction by type but no globally .

redDwarf03:
I would like to display in the mobile wallet the oracle chains like https://mainnet.archethic.net/explorer/chain/oracle

samuel-uniris:
In the case of network transaction : Oracle, Shared Secret, Nodes ... Those transaction are stored in each node , so we could create endpoint to list them based on the type . 

For example:
```
networkTransactions(type: "oracle_summary") {}
```


Any PR is welcome ;)

redDwarf03:
I propose PR #51

PR #51, opened 2021-08-07
node list GraphQL Api : add fields + Add GraphQL Api to fetch transactions list on type

redDwarf03:
See Issue #49


lib/archethic_web/graphql_schema.ex
```diff
@@ -97,6 +97,20 @@ defmodule ArchEthicWeb.GraphQLSchema do
...
+    Query the network to list the transaction on the type
+    """
+    field :network_transactions, list_of(:transaction) do
+      arg(:type, :string)
+      arg(:page, :integer)
+
+      resolve(fn args, _ ->
+        type = Map.get(args, :type)
+        page = Map.get(args, :page, 1)
+        {:ok, Resolver.network_transactions(String.to_atom(type), page)}
```
samuel-uniris, line 111: You should avoid to use `String.to_atom` as it may create a overflow of the atom list in the Erlang VM. It's a security issue, and a way to bring down nodes for attackers.

But you can either create a function to parse stringified type to atomized type, or use `String.to_existing_atom`

samuel-uniris, line 111: You should avoid to use `String.to_atom` as it may create a overflow of the atom table in the Erlang VM. It's a security issue, and a way to bring down nodes for attackers.

But you can create a function to parse stringified type to atomized type. This approach will enable a validation for the existing type and will sanitize the input.

Those checks can be done in the resolver itself.

samuel-uniris (changes requested):

redDwarf03:
is it possible to fix this PR with someone who have better skills in elixir than me please :) that's a small evolution ;)
thank you for that

samuel-uniris:
Moved to #198

PR #63, opened 2021-09-06
Added Beacon Explorer

imnik11:


lib/archethic_web/live/chains/beacon_live.ex
```diff
@@ -0,0 +1,133 @@
...
+  def render(assigns) do
+    View.render(ExplorerView, "beacon_chain_index.html", assigns)
+  end
+
+  def handle_params(%{"page" => page}, _uri, socket = %{assigns: %{dates: dates}}) do
+    case Integer.parse(page) do
+      {number, ""} when number > 0 ->
+        transactions =
+          list_transactions_by_date(
+            Enum.at(dates, number + 1),
```
samuel-uniris, line 82: You should avoid duplicated calls to a Enum, as it will iterate twice over the same list

lib/archethic_web/live/chains/beacon_live.ex
```diff
@@ -0,0 +1,133 @@
...
+          |> assign(:transactions, transactions)
+
+        {:noreply, new_assign}
+
+      _ ->
+        {:noreply, socket}
+    end
+  end
+
+  def handle_params(%{}, _, socket) do
```
samuel-uniris, line 98: I think you can avoid that, try to not be too much in the defensive programming

lib/archethic_web/live/chains/beacon_live.ex
```diff
@@ -0,0 +1,133 @@
...
+    {:noreply, socket}
+  end
+
+  def handle_event("goto", %{"page" => page}, socket) do
+    {:noreply, push_patch(socket, to: Routes.live_path(socket, __MODULE__, %{"page" => page}))}
+  end
+
+  def get_beacon_dates do
```
samuel-uniris, line 106: This one should be private I think

lib/archethic_web/live/chains/beacon_live.ex
```diff
@@ -0,0 +1,133 @@
...
+    %Node{enrollment_date: enrollment_date} =
+      P2P.list_nodes() |> Enum.sort_by(& &1.enrollment_date, {:asc, DateTime}) |> Enum.at(0)
+
+    enrollment_date
+    |> SummaryTimer.previous_summaries()
+    |> Enum.sort({:asc, DateTime})
+  end
+
+  def list_transactions_by_date(date = %DateTime{}, all_tx) do
```
samuel-uniris, line 115: I should be private as well

lib/archethic_web/live/chains/beacon_live.ex
```diff
@@ -0,0 +1,133 @@
...
+    all_tx
+    |> Enum.flat_map(& &1)
+    |> Enum.filter(fn %ArchEthic.BeaconChain.Slot.TransactionSummary{
```
samuel-uniris, line 118: I don't get this block. Why do you need do that ?

lib/archethic/beacon_chain.ex
```diff
@@ -84,6 +84,32 @@ defmodule ArchEthic.BeaconChain do
     end)
   end
 
+  @doc """
+  Retrieve the beacon summaries storage nodes from two dates
+  """
+  @spec get_summary_pool(DateTime.t(), DateTime.t()) :: pools()
+  def get_summary_pool(
```
samuel-uniris, line 91: Why do you need to have two date to get the storage nodes for the beacon summary ?

samuel-uniris (changes requested):

lib/archethic_web/live/chains/beacon_live.ex
imnik11: so, instead of calling Enum 3 times I should use two var which will store them and pass them to fetch. Right this will  reduce Enum calls from 3 to 2

samuel-uniris: If you use fetch_txs_summaries with a single date it should be better I think

imnik11: I am passing two dates to BeaconChain.get_summary_pool/2 as BeaconChain.get_summary_pools/1 takes one date and get data from the last sync date. So, I am trying a get that data in particular time interval say date1 and date2.

samuel-uniris: Summary transaction address are based on a single date. So I don't see yet the need to use two dates to get the transaction.
Moreover it should improve the performance I think.

lib/archethic/beacon_chain.ex
imnik11: when I get that from the one date it is taking a lot of time,

    summary_times1 = SummaryTimer.previous_summaries(last_sync_date)
    summary_times2 = SummaryTimer.previous_summaries(date)
    summary_times = summary_times1 -- summary_times2

so, this what I did to overcome that.

samuel-uniris: Once you have a date, you should not iterate over the dates anymore.
Just perform election based on the subsets and the date, and you should have all the nodes.
So no need to leverage `SummaryTimer.previous_summaries` anymore.

PR #64, opened 2021-09-06
Add faucet and beacon chain explorer

roychowdhuryrohit-dev:
Added Faucet feature for testnet and its fixes.
Added Beacon chain explorer and its fixes.
Fixed formatting issues.
Fixed failing test case.


roychowdhuryrohit-dev:
This PR also contains the changes in https://github.com/archethic-foundation/archethic-node/pull/63

assets/css/app.scss
```diff
@@ -4,3 +4,53 @@
 @import "~bulma/bulma";
 
 @import "explorer";
+
+.alert {
```
samuel-uniris, line 8: You could leverage bulma component like this one: https://bulma.io/documentation/elements/notification/

lib/archethic_web/controllers/faucet_controller.ex
```diff
@@ -0,0 +1,137 @@
...
+        |> redirect(to: "/faucet")
+
+      {:error, _} ->
+        conn
+        |> put_flash(:error, "Unable to transfer")
+        |> render("index.html", address: address)
+    end
+  end
+
+  def transfer(
```
samuel-uniris, line 37: I think you should put it in private as it's only used by the controller function.

lib/archethic_web/controllers/faucet_controller.ex
```diff
@@ -0,0 +1,137 @@
...
+        address,
+        curve \\ Crypto.default_curve()
+      )
+      when is_bitstring(address) do
+    gen_seed =
```
samuel-uniris, line 42: I think for this one you can leverage module attributes generated at the compilation time, unless you want to customize it at runtime

lib/archethic_web/controllers/faucet_controller.ex
```diff
@@ -0,0 +1,137 @@
...
+      },
+      gen_seed,
+      last_index,
+      curve
+    )
+    |> ArchEthic.send_new_transaction()
+  end
+end
+
+#   def transfer(
```
samuel-uniris, line 86: Clean this one ;)

lib/archethic_web/router.ex
```diff
@@ -25,6 +25,8 @@ defmodule ArchEthicWeb.Router do
     get("/", RootController, :index)
     get("/up", UpController, :up)
     get("/metrics", MetricsController, :index)
+    get("/faucet", FaucetController, :index)
```
samuel-uniris, line 28: For those endpoints, you should add a condition when to display them.
We do not want to have this one production.
For example, you could add a configuration for this at compile time

lib/archethic_web/templates/faucet/index.html.eex
```diff
@@ -0,0 +1,25 @@
...
+  <div class="container">
+
+    <h1 class="title main_title">ArchEthic Faucet</h1>
+    <h2 class="subtitle main_subtitle">
+      <div class="columns">
+        <div class="column">
+          Discover the most decentralized and unlimited P2P network
+        </div>
+      </div>
+      <form class="columns" action="/faucet" method="post">
```
samuel-uniris, line 12: For the action, you should leverage Phoenix helpers to generate the route

lib/archethic_web/templates/layout/root.html.eex
```diff
@@ -25,7 +25,11 @@
 
         <div id="navbar" class="navbar-menu">
             <div class="navbar-end">
-    <!-- 
+
+            <a class="navbar-item" href="<%= Routes.faucet_path(@conn, :index) %>">
```
samuel-uniris, line 29: Same, it should be displayed only when there is a configuration enabling this (not in production)

lib/archethic_web/views/faucet_view.ex
```diff
@@ -0,0 +1,3 @@
+defmodule ArchEthicWeb.FaucetView do
```
samuel-uniris, line 1: I don't think you need a specific view for this, you can leverage the explorer view.

`Phoenix.View.render(view, template, assigns)`

scripts/install.sh
```diff
@@ -101,12 +101,13 @@ sudo apt-get update
 sudo apt-get install -y openjdk-8-jre-headless
 sudo update-java-alternatives --jre-headless -s java-1.8.0-openjdk-amd64
 
-sudo scylla_setup --no-raid-setup --no-ec2-check --no-kernel-check --no-verify-package --no-sysconfig-setup --io-setup=1 --no-version-check --no-cpuscaling-setup --no-fstrim-setup --no-memory-setup --no-swap-setup
+sudo scylla_setup --no-raid-setup --no-ec2-check --no-kernel-check --no-verify-package --no-sysconfig-setup --io-setup=1 --no-version-check --no-cpuscaling-setup --no-fstrim-setup --no-memory-setup --no-swap-setup 2>/dev/null
```
samuel-uniris, line 104: Can you extract this into another PR ?

test/archethic/governance/code_test.exs
```diff
@@ -12,7 +12,7 @@ defmodule ArchEthic.Governance.CodeTest do
 
   test "applicable_proposal? should succeed" do
     changes = File.read!(Path.join(__DIR__, "0001-Propose-to-ignore-file.patch"))
-    assert Code.applicable_proposal?(%Proposal{changes: changes, address: "123"})
+    # assert Code.applicable_proposal?(%Proposal{changes: changes, address: "123"})
```
samuel-uniris, line 15: You should uncomment this one

priv/static/docs/development/testnet/local.md
```diff
@@ -36,12 +36,13 @@ iex -S mix
...
 - Ensure a fresh start
-```
+```bash
 make clean
 ```
 
 - Restart the node with some configuration
-```
+```bash
+rm -rf ./_build
```
samuel-uniris, line 45: Why you need this one ?

priv/static/docs/index.html
```diff
@@ -24,6 +24,6 @@
   <script src="//cdn.jsdelivr.net/npm/docsify@4"></script>
   <script src="//cdn.jsdelivr.net/npm/docsify/lib/plugins/search.min.js"></script>
   <script src="//cdn.jsdelivr.net/npm/docsify-copy-code"></script>
-
+  <script src="//cdn.jsdelivr.net/npm/prismjs@1/components/prism-bash.min.js"></script>
```
samuel-uniris, line 27: What is it for ?

samuel-uniris (changes requested):

priv/static/docs/index.html
roychowdhuryrohit-dev: Bash syntax highlighting in docsify.

priv/static/docs/development/testnet/local.md
roychowdhuryrohit-dev: Bash syntax highlighting in docsify.

samuel-uniris: I mean the `rm-rf ./build`

test/archethic/governance/code_test.exs
roychowdhuryrohit-dev: Okay.

priv/static/docs/index.html
samuel-uniris: ok

scripts/install.sh
roychowdhuryrohit-dev: I will remove this change since it doesn’t make any difference during Scylladb installation.

lib/archethic_web/controllers/faucet_controller.ex
roychowdhuryrohit-dev: Ofcourse.

scripts/install.sh
samuel-uniris: ok

priv/static/docs/development/testnet/local.md
roychowdhuryrohit-dev: Yes it is there to ensure a fresh start to be able to fund some addresses you can specify in the configuration which will be the addresses and the amount as genesis pool during the network initialization. Few weeks ago I tried this and `make clean` was not enough I had to remove the build to solve the errors.

lib/archethic_web/router.ex
roychowdhuryrohit-dev: Which approach will be a wiser decision? To make a separate testnet git branch or use environment variable to detect the testnet config?

priv/static/docs/development/testnet/local.md
samuel-uniris: If you face this issue, it's due to a configuration based on compilation instead of runtime; you can check this part to avoid this

scripts/install.sh
roychowdhuryrohit-dev: I have used `sleep 5` since the `nodetool status` shows error as scylla-server server takes time to start.

lib/archethic_web/router.ex
samuel-uniris: I think the best can be to use environment variable so in dev you will have access to the faucet.

priv/static/docs/development/testnet/local.md
roychowdhuryrohit-dev: Yes it is in fact due to the configuration during compilation differs from configuration during runtime. This is why in order to do a fresh start you have to explicitly remove the previous compiled build and rebuild the project again with new configuration. This is for dev environment.

roychowdhuryrohit-dev: Should I keep this command for posterity?

samuel-uniris: You can move this to use runtime configuration call instead

lib/archethic_web/templates/layout/root.html.eex
roychowdhuryrohit-dev: Okay I will use the environment variable. But I also need to deploy the faucet in testnet production release.

samuel-uniris: configuration from env variables should make it

lib/archethic_web/controllers/faucet_controller.ex
roychowdhuryrohit-dev: For the curve or pool_gen_seed?

samuel-uniris: For the genesis seed for testnet

priv/static/docs/development/testnet/local.md
roychowdhuryrohit-dev: Fix https://github.com/archethic-foundation/archethic-node/pull/74 to be reviewed

lib/archethic_web/controllers/faucet_controller.ex
roychowdhuryrohit-dev: I want to know what is ultimately the default curve for our whole project? Is it **secp256r1** or ### ed25519? Because the currently the default value of **ARCHETHIC_TESTNET_GENESIS_ADDRESS** is generated using **secp256r1** whereas our default curve in the config is **ed25519**.

samuel-uniris: You have the capability to determine which curve to use because each public key is prepended by two bytes: curve id and origin id.
Using Crypto module you can determine the curve and you can use this one in the derivation key algorithm .

roychowdhuryrohit-dev: Oh yes I can do that as well thanks.

roychowdhuryrohit-dev: To derive the public key from seed, we need to know the curve first right?

roychowdhuryrohit-dev: Suppose I have the seed and address (of pool) both as environment variables. Is it possible to derive the curve from these two informations?

roychowdhuryrohit-dev: Is it possible to determine the curve just from seed/address? Because the private key is not available in environment.

If I can get the curve I can use this :
```elixir
new(type, data, seed, index, curve \\ Crypto.default_curve())
```

If I can get the private key I can use this instead : 
```elixir
new_with_keys(type, data, previous_private_key, previous_public_key, next_public_key)
```

But without either of them I can't generate a new transaction.

lib/archethic_web/router.ex
roychowdhuryrohit-dev: Is it possible to get the public key from the seed/address? Without the public key I can't determine the curve.

If I can get the curve I can use this :
```elixir
new(type, data, seed, index, curve \\ Crypto.default_curve())
```

If I can get the private key I can use this instead : 
```elixir
new_with_keys(type, data, previous_private_key, previous_public_key, next_public_key)
```

But without either of them I can't generate a new transaction.

lib/archethic_web/controllers/faucet_controller.ex
roychowdhuryrohit-dev: Is it possible to get the public key from the seed/address? Without the public key I can't determine the curve.

If I can get the curve I can use this :
```elixir
new(type, data, seed, index, curve \\ Crypto.default_curve())
```

If I can get the private key I can use this instead : 
```elixir
new_with_keys(type, data, previous_private_key, previous_public_key, next_public_key)
```

But without either of them I can't generate a new transaction.

samuel-uniris: I don't think it's possible.
We should rethink the way to add the testnet seed into the conf.
We could imagine something like:

config/config.exs
```elixir
 config :archethic, ArchEthicWeb.FaucetController,
      seed: "3A7B579DBFB7CEBE26293850058F180A65D6A3D2F6964543F5EDE07BEB2EFDA4" |> Base.decode16!
```

config/dev.exs

```elixir
config  :archethic, ArchEthic.Bootstrap.NetworkInit, 
   genesis_pools: [
      # Address generated from the faucet seed
      %{ address: "00EC64107CA604A6B954037CFA91ED18315A77A94FBAFD91275CEE07FA45EAF893", amount: 10000000}
   ]
  
```

config/prod.exs

```elixir
config :archethic, ArchEthic.Bootstrap.NetworkInit, 
  genesis_pools: [
    ...
    ...,
   (if System.get_env("ARCHETHIC_TESTNET", "false") == "true", do:       
       "00EC64107CA604A6B954037CFA91ED18315A77A94FBAFD91275CEE07FA45EAF893"
    )
  ]
  |> Enum.filter(& &1)
```

roychowdhuryrohit-dev: Suppose I have both the seed and address now. But I don't think I can do a new transaction without knowing the curve which I cannot derive without knowing the public key.

samuel-uniris: You can because the address uses the default curve, as we are not customizing the faucet address

roychowdhuryrohit-dev: But in docsify it is mentioned we can customise the pool address. Should we remove that from doc?

samuel-uniris: Yes we will if it works more

assets/css/app.scss
roychowdhuryrohit-dev: I was wondering if the JavaScript DOM manipulation works in the context of Server Side Rendering nature of Phoenix. Do I need to re-render the root layout every time any child elements get removed? If that's the case then why does the initial loading of root layout look like this? 

Screenshot 2021-09-14 at 01 30 18

Screenshot 2021-09-14 at 01 25 58

roychowdhuryrohit-dev: Using the default Bootstrap alert css for now.

lib/archethic_web/controllers/faucet_controller.ex
roychowdhuryrohit-dev: > You can because the address uses the default curve, as we are not customizing the faucet address

Right now the address is `0073bdaf847037115914ff5ca15e52d162db57b5089d5e4bf2005d825592c9c945` which has been derived using secp256r1. I am changing the existing pool address to the one using ed25519 default curve.

roychowdhuryrohit-dev:
_TODO : Remove https://github.com/archethic-foundation/archethic-node/pull/63 from git history using rebase _

lib/archethic_web/router.ex
roychowdhuryrohit-dev: In release `Mix.env()` doesn't work as Mix is not available. I want to inject these two lines at only compile time according to Mix.env() condition (:dev or :prod with testnet).
```elixir
get("/faucet", FaucetController, :index)
post("/faucet", FaucetController, :create_transfer)
```
Do I need to leverage Macros to inject something at compile time according to conditions?

roychowdhuryrohit-dev: Using config and not Macros code injection to resolve this issue.

roychowdhuryrohit-dev: Okay `Mix.env() == :dev` was enough for this since in release we can check `System.get_env("ARCHETHIC_NETWORK_TYPE") == "testnet"`.

roychowdhuryrohit-dev:
Closing this PR since https://github.com/archethic-foundation/archethic-node/tree/36cb7d25033be65b79d37e1e84966592fa1e6dc8 is outdated.

PR #74, opened 2021-09-08
Fix the testnet genesis overloading on runtime

samuel-uniris:
Use runtime configuration instead of compile configuration
to be able to overload the testnet genesis address and amount.


roychowdhuryrohit-dev:
Yes it is working fine.

PR #76, opened 2021-09-09
Added Beacon Explorer

### LINKED ISSUE #66: Add BeaconChain Explorer
Provide an interface like the OracleChain to get the list of transaction in the network using the BeaconChain

imnik11:
BeaconChain Explorer
Interface to fetch the transactions from the BeaconChain (chain where all the transaction are listed) and to split by days using pages

Fixes #66


samuel-uniris (changes requested):

lib/archethic_web/live/chains/beacon_live.ex
```diff
@@ -0,0 +1,102 @@
...
+  alias ArchEthic.BeaconChain.SummaryTimer
+  alias ArchEthic.Crypto
+  alias ArchEthic.Election
+  alias ArchEthic.P2P
+  alias ArchEthic.P2P.Node
+  alias ArchEthic.SelfRepair.Sync.BeaconSummaryHandler
+  alias ArchEthicWeb.ExplorerView
+  alias Phoenix.View
+
+  defp f2(date = %DateTime{}) do
```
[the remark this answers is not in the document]
imnik11: so that to list_transaction_by_date

lib/archethic_web/live/chains/beacon_live.ex
```diff
@@ -0,0 +1,102 @@
...
+      end)
+      |> :lists.flatten()
+      |> Task.async_stream(
+        fn {address, nodes, patch} ->
+          BeaconSummaryHandler.download_summary(address, nodes, patch)
+        end,
+        on_timeout: :kill_task,
+        max_concurrency: 256
+      )
+      |> Enum.to_list()
```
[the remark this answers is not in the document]
imnik11: No, I just want it be a list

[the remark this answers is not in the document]
samuel-uniris: The result of async_stream it's a list but streamed, so you can make operations on the list, only the calls will trigger the processing.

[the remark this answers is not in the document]
imnik11: cool. Then we don't need it I assumed Stream.filter takes Enum list for filtering. 

Any other changes ?

lib/archethic_web/live/chains/beacon_live.ex
```diff
@@ -12,28 +13,27 @@ defmodule ArchEthicWeb.BeaconChainLive do
...
+    end)
+    |> Task.async_stream(
+      fn {address, nodes, patch} ->
+        BeaconSummaryHandler.download_summary(address, nodes, patch)
+      end,
+      on_timeout: :kill_task,
+      max_concurrency: 256
+    )
+    |> Stream.filter(&match?({:ok, {:ok, %BeaconSummary{}}}, &1))
+    |> Enum.map(fn {:ok, {:ok, %BeaconSummary{transaction_summaries: transaction_summaries}}} ->
```
samuel-uniris, line 33: You can also use Stream.map or even use Stream.flat_map, to  map and flatten at once

lib/archethic_web/templates/explorer/beacon_chain_index.html.leex
```diff
@@ -0,0 +1,73 @@
...
+        </ul>
+      </nav>
+    </div>
+  </div>
+
+
+  <div class="columns">
+    <div class="column">
+      <div class="box">
+        <p class="heading is-size-6">Transaction chain for <%= format_date(Enum.at(@dates, @current_date_page)) %></p>
```
samuel-uniris, line 50: I think you should display the date from the `@current_date_page - 1`.
For instance on the page 1, if there is not more than one date, `Enum.at(@dates, 1)` will return nil

lib/archethic_web/live/chains/beacon_live.ex
```diff
@@ -0,0 +1,102 @@
...
+      |> Enum.to_list()
+      |>:lists.flatten()
+      |> Stream.filter(&match?({:ok, {:ok, %BeaconSummary{}}}, &1))
+      |> Enum.map(fn {:ok, {:ok, %BeaconSummary{transaction_summaries: transaction_summaries}}} -> transaction_summaries end)
+      |>:lists.flatten()
+  end
+
+  def mount(_params, _session, socket) do
+    # Todo handle live transactions
+    # if connected?(socket) do
```
samuel-uniris, line 42: It will be better to active realtime for beacon_summary transaction and update the pagination and the transaction list

lib/archethic_web/live/chains/beacon_live.ex
```diff
@@ -0,0 +1,104 @@
...
+    {:noreply, push_patch(socket, to: Routes.live_path(socket, __MODULE__, %{"page" => page}))}
+  end
+
+  defp get_beacon_dates do
+    %Node{enrollment_date: enrollment_date} =
+      P2P.list_nodes() |> Enum.sort_by(& &1.enrollment_date, {:asc, DateTime}) |> Enum.at(0)
+
+    enrollment_date
+    |> SummaryTimer.previous_summaries()
+    |> Enum.sort({:asc, DateTime})
```
samuel-uniris, line 102: I think it should be descendant because we want to display the last one first

lib/archethic_web/live/chains/beacon_live.ex
```diff
@@ -0,0 +1,104 @@
...
+  def mount(_params, _session, socket) do
+    # Todo handle live transactions
+    # if connected?(socket) do
+    #   PubSub.register_to_new_transaction_by_type(:beacon)
+    #   PubSub.register_to_new_transaction_by_type(:beacon_summary)
+    # end
+
+    dates = get_beacon_dates()
+
+    beacon_dates =  if Enum.empty?(dates) do
```
samuel-uniris, line 49: You should run `mix format` to format the indentation

lib/archethic_web/live/chains/beacon_live.ex
imnik11: Ok, changed to :desc

lib/archethic/pub_sub.ex
```diff
@@ -67,6 +68,14 @@ defmodule ArchEthic.PubSub do
     dispatch(:new_oracle_data, {:new_oracle_data, data})
   end
 
+  @doc """
+  Notify next summary time beacon chain to the subscribers
+  """
+  def notify_next_summary_time(date = %DateTime{}) do
+    next_summary_time = BeaconChain.next_summary_date(date)
```
samuel-uniris, line 75: Again several times the same thing, because from your function above you pass the next summary time to this function, so you will get again the same or the next one depending on the seconds/microseconds.
This function should just straightforward: given a time, it will notify all subscribers.

samuel-uniris (changes requested):

lib/archethic/pub_sub.ex
imnik11: so, changed notify fn to this

  def notify_next_summary_time(date = %DateTime{}) do
    dispatch(:next_summary_time, {:next_summary_time, date})
  end

lib/archethic/beacon_chain/summary_timer.ex
```diff
@@ -98,6 +99,24 @@ defmodule ArchEthic.BeaconChain.SummaryTimer do
...
 
+  def handle_info(
+        :next_summary_time,
+        state = %{
+          interval: interval
+        }
+      ) do
+    timer = schedule_next_summary_time(interval)
+
+    slot_time = DateTime.utc_now() |> Utils.truncate_datetime()
```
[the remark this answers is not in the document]
imnik11: renamed slot_time with current_time

lib/archethic/beacon_chain/summary_timer.ex
```diff
@@ -98,6 +99,24 @@ defmodule ArchEthic.BeaconChain.SummaryTimer do
...
+
+    PubSub.notify_next_summary_time(next_summary(slot_time))
```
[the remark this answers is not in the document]
imnik11: but I wanted to know will it make any difference?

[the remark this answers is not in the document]
samuel-uniris: Yes in terms of performance, as it will require `ets` lookup + CronTab next scheduling time. Also,  the computation is not necessary, as you are on summary time.

samuel-uniris (approved):

PR #78, opened 2021-09-09
Imple alpine

ambareesha7:
I'm using alpinejs on front end for little functionalities like switching between tab's, intention of this is to minimise the calls to backend for trivial functions, as suggested by Sam 
but its still  work in progress


lib/archethic_web/live/transaction_details_live.ex
```diff
@@ -19,9 +19,7 @@ defmodule ArchEthicWeb.TransactionDetailsLive do
        exists: false,
        previous_address: nil,
        transaction: nil,
-       hide_content: true,
        tab_panel: "tx",
```
samuel-uniris, line 22: I think you can clean the `tab_panel` and the `operation_section`

lib/archethic_web/templates/explorer/transaction_details.html.leex
```diff
@@ -30,22 +28,30 @@
                         <% end %>
                     </div>
                 </div>
-
             <% end %>
-
-            <div class="is-2">
+            <div class="is-2" x-data="{
+                                tab_panel: '<%= @tab_panel %>',
```
samuel-uniris, line 33: I think you can default as `tx` panel

transaction_HTTP_response.txt
```diff
@@ -0,0 +1,130 @@
+START
```
samuel-uniris, line 1: This file should be removed also

samuel-uniris (changes requested):

lib/archethic_web/live/transaction_details_live.ex
```diff
@@ -72,7 +70,7 @@ defmodule ArchEthicWeb.TransactionDetailsLive do
 
   defp handle_transaction(
          socket,
-         tx = %Transaction{address: address, data: %TransactionData{content: content}}
+         tx = %Transaction{address: address, data: %TransactionData{content: _content}}
```
samuel-uniris, line 73: I think you can also clean this one

lib/archethic_web/live/transaction_details_live.ex
samuel-uniris: Why do you need to keep this from the server side ?

lib/archethic_web/live/transaction_details_live.ex
ambareesha7: transaction.data.content we are using it in leex file
Screenshot 2021-09-20 at 2 37 02 PM

samuel-uniris: We are already passing the transaction to the view, but here the pattern matching was used to determine if you show the content depending on its size, for performance issue.
So if you want to use this var, you should set this variable into the alpine.js data also.

ambareesha7: ok, i have removed "data: %TransactionData{content: _content"

samuel-uniris (approved):

samuel-uniris:
Look good to me

PR #91, opened 2021-09-17
Add testnet faucet

roychowdhuryrohit-dev:
Added faucet for dev and prod (testnet).
Fixed #64 PR suggestions.
Added custom logger level for dev config.

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -1,5 +1,6 @@
 import Config
 
+config :logger, level: System.get_env("ARCHETHIC_LOGGER_LEVEL", "debug") |> String.to_atom()
 config :archethic,
        :mut_dir,
        System.get_env(
@@ -36,13 +37,9 @@ config :archethic, ArchEthic.Bootstrap.NetworkInit,
   genesis_pools: [
     %{
       address:
-        System.get_env(
-          "ARCHETHIC_TESTNET_GENESIS_ADDRESS",
-          "0073bdaf847037115914ff5ca15e52d162db57b5089d5e4bf2005d825592c9c945"
-        )
+        "00EC64107CA604A6B954037CFA91ED18315A77A94FBAFD91275CEE07FA45EAF893"
         |> Base.decode16!(case: :mixed),
-      amount:
-        System.get_env("ARCHETHIC_TESTNET_GENESIS_AMOUNT", "10000000") |> Float.parse() |> elem(0)
+      amount: "10000000" |> Float.parse() |> elem(0)
     }
   ]
 
--- a/lib/archethic_web/controllers/faucet_controller.ex
+++ b/lib/archethic_web/controllers/faucet_controller.ex
@@ -1,0 +1,80 @@
+defmodule ArchEthicWeb.FaucetController do
+  @moduledoc false
+
+  alias ArchEthic.TransactionChain.{
+    Transaction,
+    TransactionData,
+    TransactionData.Ledger,
+    TransactionData.UCOLedger
+  }
+
+  alias ArchEthic.Crypto
+
+  use ArchEthicWeb, :controller
+
+  def index(conn, __params) do
+    conn
+    |> put_resp_header("cache-control", "no-cache, no-store, must-revalidate")
+    |> put_resp_header("pragma", "no-cache")
+    |> put_resp_header("expires", "0")
+    |> render("index.html", address: "")
+  end
+
+  def create_transfer(conn, %{"address" => address}) do
+    case transfer(address, Crypto.default_curve()) do
+      :ok ->
+        conn
+        |> put_flash(:info, "Transferred successfully")
+        |> redirect(to: "/faucet")
+
+      {:error, _} ->
+        conn
+        |> put_flash(:error, "Unable to transfer")
+        |> render("index.html", address: address)
+    end
+  end
+
+  defp transfer(
+         address,
+         curve
+       )
+       when is_bitstring(address) do
+    gen_seed = Application.get_env(:archethic, __MODULE__)[:seed]
+
+    {gen_pub, _} = Crypto.derive_keypair(gen_seed, 0, curve)
+
+    pool_gen_address = Crypto.hash(gen_pub)
+
+    last_address =
+      case ArchEthic.get_last_transaction(pool_gen_address) do
+        {:ok, transaction} ->
+          %ArchEthic.TransactionChain.Transaction{address: last_address} = transaction
+          last_address
+
+        {:error, :transaction_not_exists} ->
+          pool_gen_address
+      end
+
+    last_index = ArchEthic.get_transaction_chain_length(last_address)
+
+    Transaction.new(
+      :transfer,
+      %TransactionData{
+        ledger: %Ledger{
+          uco: %UCOLedger{
+            transfers: [
+              %UCOLedger.Transfer{
+                to: Base.decode16!(address, case: :mixed),
+                amount: 100.0
+              }
+            ]
+          }
+        }
+      },
+      gen_seed,
+      last_index,
+      curve
+    )
+    |> ArchEthic.send_new_transaction()
+  end
+end
--- a/lib/archethic_web/router.ex
+++ b/lib/archethic_web/router.ex
@@ -26,6 +26,11 @@ defmodule ArchEthicWeb.Router do
     get("/up", UpController, :up)
     get("/metrics", MetricsController, :index)
     live_dashboard("/dashboard", metrics: ArchEthic.Telemetry)
+
+    if Mix.env() in [:dev, :test] or System.get_env("ARCHETHIC_NETWORK_TYPE") == "testnet" do
+      get("/faucet", FaucetController, :index)
+      post("/faucet", FaucetController, :create_transfer)
+    end
   end
 
   scope "/explorer", ArchEthicWeb do
--- a/lib/archethic_web/templates/layout/root.html.eex
+++ b/lib/archethic_web/templates/layout/root.html.eex
@@ -13,25 +13,29 @@
     <div class="container is-fluid">
       <nav class="navbar is-transparent" role="navigation" aria-label="main navigation">
         <div class="navbar-brand">
-            <a class="navbar-item is-size-5 is-uppercase" href="<%= Routes.live_path(@conn, ArchEthicWeb.ExplorerIndexLive) %>">
-              ArchEthic
-            </a>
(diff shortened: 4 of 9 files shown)


config/dev.exs
```diff
@@ -36,13 +37,9 @@ config :archethic, ArchEthic.Bootstrap.NetworkInit,
...
       address:
-        System.get_env(
-          "ARCHETHIC_TESTNET_GENESIS_ADDRESS",
-          "0073bdaf847037115914ff5ca15e52d162db57b5089d5e4bf2005d825592c9c945"
-        )
+        "00EC64107CA604A6B954037CFA91ED18315A77A94FBAFD91275CEE07FA45EAF893"
         |> Base.decode16!(case: :mixed),
-      amount:
-        System.get_env("ARCHETHIC_TESTNET_GENESIS_AMOUNT", "10000000") |> Float.parse() |> elem(0)
+      amount: "10000000" |> Float.parse() |> elem(0)
```
samuel-uniris, line 42: Why don't you use a float directly instead of parsing it ?

lib/archethic_web/controllers/faucet_controller.ex
```diff
@@ -0,0 +1,80 @@
...
+        |> render("index.html", address: address)
+    end
+  end
+
+  defp transfer(
+         address,
+         curve
+       )
+       when is_bitstring(address) do
+    gen_seed = Application.get_env(:archethic, __MODULE__)[:seed]
```
samuel-uniris, line 42: Because it's not subject to change, I think you can use compilation environment using module attributes

lib/archethic_web/controllers/faucet_controller.ex
```diff
@@ -0,0 +1,80 @@
...
+      {:error, _} ->
+        conn
+        |> put_flash(:error, "Unable to transfer")
+        |> render("index.html", address: address)
+    end
+  end
+
+  defp transfer(
+         address,
+         curve
```
samuel-uniris, line 39: Why do you want to use a curve because faucet transactions will be the same everytime.

lib/archethic_web/controllers/faucet_controller.ex
```diff
@@ -0,0 +1,80 @@
...
+  def index(conn, __params) do
+    conn
+    |> put_resp_header("cache-control", "no-cache, no-store, must-revalidate")
+    |> put_resp_header("pragma", "no-cache")
+    |> put_resp_header("expires", "0")
+    |> render("index.html", address: "")
+  end
+
+  def create_transfer(conn, %{"address" => address}) do
+    case transfer(address, Crypto.default_curve()) do
```
samuel-uniris, line 24: I think you need to check if the address is well formatted

lib/archethic_web/router.ex
```diff
@@ -26,6 +26,11 @@ defmodule ArchEthicWeb.Router do
     get("/up", UpController, :up)
     get("/metrics", MetricsController, :index)
     live_dashboard("/dashboard", metrics: ArchEthic.Telemetry)
+
+    if Mix.env() in [:dev, :test] or System.get_env("ARCHETHIC_NETWORK_TYPE") == "testnet" do
```
samuel-uniris, line 30: Will this `Mix.env() in [:dev, :test]` work with release ?

lib/archethic_web/templates/layout/root.html.eex
```diff
@@ -13,25 +13,29 @@
...
-
         <div id="navbar" class="navbar-menu">
-            <div class="navbar-end">
-    <!-- 
-          TODO: implement the listing of transactions through BeaconChain
-
-              <a class="navbar-item" href="<%= Routes.live_path(@conn, ArchEthicWeb.TransactionListLive) %>">
-                 Transactions
+          <div class="navbar-end">
+            <%= if Mix.env() in [:dev, :test] or System.get_env("ARCHETHIC_NETWORK_TYPE") == "testnet" do %>
```
samuel-uniris, line 27: Will this `Mix.env() in [:dev, :test]` work with release ?

lib/archethic_web/router.ex
samuel-uniris: Otherwise, you can configuration

samuel-uniris (changes requested):

lib/archethic_web/templates/layout/root.html.eex
roychowdhuryrohit-dev: Yes since in release `System.get_env("ARCHETHIC_NETWORK_TYPE") == "testnet"`.

config/dev.exs
roychowdhuryrohit-dev: Okay sure.

lib/archethic_web/router.ex
samuel-uniris: Yes but Mix.env() will be called

roychowdhuryrohit-dev: Okay I get it.

lib/archethic_web/templates/layout/root.html.eex
samuel-uniris: I think you can use the layout view to create an helper to test the network type and use it to display or not the link

```elixir
defmodule ArchEthicWeb.LayoutView do
   def network_type do
      System.get_env("ARCHETHIC_NETWORK_TYPE") == "testnet"
   end
end
```

lib/archethic_web/router.ex
samuel-uniris: Maybe using Application.get_env should do it, the System.get_env could loaded from the conf

roychowdhuryrohit-dev: What if I use `cond`?

samuel-uniris: Cond is use for multiple condition like switch

roychowdhuryrohit-dev: For example :

```elixir
if System.get_env("ARCHETHIC_NETWORK_TYPE") == "testnet" do
  do_something()
else
  if Mix.env() in [:dev, :test] do
    do_something()
  end
end
```

Will this logic work?

roychowdhuryrohit-dev: For example :

```elixir
if System.get_env("ARCHETHIC_NETWORK_TYPE") == "testnet" do
  do_something()
else
  if System.get_env("ARCHETHIC_NETWORK_TYPE") != "mainnet" and Mix.env() in [:dev, :test] do
    do_something()
  end
end
```

Will this logic work?

samuel-uniris: You can avoid all of this complexity by just using conf .

roychowdhuryrohit-dev: But is conf executed in release?

samuel-uniris: Yes sure . Compilation or runtime configuration

samuel-uniris: It's the cleanest way to deal with env variable

lib/archethic_web/controllers/faucet_controller.ex
roychowdhuryrohit-dev: I was using this as a utility function for debugging.

samuel-uniris: So you this one as default parameter, ie: ```def transfer(address, curve \\ Crypto.default_curve())```

roychowdhuryrohit-dev: Yes I did that in my earlier version when it was not a private function. Should I make it public now?

samuel-uniris: No need to make it public

lib/archethic_web/controllers/faucet_controller.ex
roychowdhuryrohit-dev: Is there any utility API function to do that or do I have to pattern match the binary explicitly?

lib/archethic_web/controllers/faucet_controller.ex
roychowdhuryrohit-dev: Using the config as module attribute like this?

```elixir
defmodule ArchEthicWeb.FaucetController do
  @external_resource "config/config.exs"
  @seed Application.get_env(:archethic, __MODULE__)[:seed]

  defp transfer do
    @seed
  end
end
```

lib/archethic_web/router.ex
roychowdhuryrohit-dev: So to keep testnet and mainnet separate in prod, one needs to change the config?

lib/archethic_web/templates/layout/root.html.eex
roychowdhuryrohit-dev: Should I use ARCHETHIC_NETWORK_TYPE env variable or do I manage it from config?

lib/archethic_web/controllers/faucet_controller.ex
samuel-uniris: You can indeed use `@seed` attribute to load from the configuration.

But the no need for the `@external_resource` attribute

lib/archethic_web/controllers/faucet_controller.ex
samuel-uniris: You can take inspiration from https://github.com/archethic-foundation/archethic-node/blob/39e466dab1feb26c6446ba248247cbd58354f93b/lib/archethic_web/controllers/explorer_controller.ex#L13

lib/archethic_web/router.ex
samuel-uniris: I think it's the idea so we could spawn Mainnet and testnet node just by specifying this conf and of course the bootstrapping seeds (but this is not on this PR subject)

lib/archethic_web/templates/layout/root.html.eex
samuel-uniris: The best would be to use configuration to avoid direct call to System.

lib/archethic_web/controllers/faucet_controller.ex
roychowdhuryrohit-dev: I shouldn't do this check right?

`{:ok, tx} <- ArchEthic.search_transaction(address)`

Because it won't exist for first time users.

samuel-uniris: Yes indeed for the first time address it will not exist. So no need to make this check

PR #95, opened 2021-09-21
Add testnet Faucet

roychowdhuryrohit-dev:
Added faucet for dev and prod (testnet).
Fixed #64 PR suggestions.
Added custom logger level for dev config.
Fix #91 PR suggestions


config/dev.exs
```diff
@@ -115,6 +113,8 @@ config :archethic, ArchEthic.SharedSecrets.NodeRenewalScheduler,
 config :archethic, ArchEthic.P2P.Endpoint,
   port: System.get_env("ARCHETHIC_P2P_PORT", "3002") |> String.to_integer()
 
+config :archethic, faucet: true
```
samuel-uniris, line 116: I think you can use 

```elixir
config :archethic, ArchEthicWeb.FaucetController, 
  enabled: true
``` 
instead of `config :archethic, faucet: true`

to more consistent with the rest and also it will only be related to the controller

lib/archethic_web/controllers/faucet_controller.ex
```diff
@@ -0,0 +1,85 @@
...
+    |> render("index.html", address: "")
+  end
+
+  def create_transfer(conn, %{"address" => address}) do
+    with {:ok, address} <- Base.decode16(address, case: :mixed),
+         true <- Crypto.valid_hash?(address),
+         :ok <- transfer(address) do
+      conn
+      |> put_flash(:info, "Transferred successfully")
+      |> redirect(to: "/faucet")
```
samuel-uniris, line 30: I didn't catch this one before, but if you can use routes helper to make the redirection dynamically generated by phoenix pointing to the `index/2` function, it will be great.

samuel-uniris (changes requested):
Seems good.
Just a few catches to make it more consistent with the actual code base.

config/dev.exs
roychowdhuryrohit-dev: I will be using faucet in both router and LayoutView. Can I mention multiple modules one config call?

lib/archethic_web/controllers/faucet_controller.ex
```diff
@@ -0,0 +1,85 @@
+defmodule ArchEthicWeb.FaucetController do
+  @moduledoc false
+  @pool_seed Application.get_env(:archethic, __MODULE__)[:seed]
```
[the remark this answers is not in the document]
roychowdhuryrohit-dev: It should have been automatically ordered by ElixirLS but alright I will do it.

config/dev.exs
samuel-uniris: For the LayoutView, because your route is related to the FaucetController, I think you can keep it inside the FaucetController configuration

roychowdhuryrohit-dev:
I have resolved the suggestions.

PR #100, opened 2021-09-24
Add faucet controller plug

roychowdhuryrohit-dev:
Using plug to enable faucet route to fix compilation issues.
Added cron.service to run archethic node service after cron.


lib/archethic_web/controllers/faucet_controller.ex
```diff
@@ -14,6 +14,19 @@ defmodule ArchEthicWeb.FaucetController do
 
   @pool_seed Application.compile_env(:archethic, __MODULE__)[:seed]
 
+  plug(:enabled)
+
+  defp enabled(conn, _) do
+    if Application.get_env(:archethic, ArchEthicWeb.FaucetController)
```
samuel-uniris, line 20: I think you can use `__MODULE__` instead

lib/archethic_web/controllers/faucet_controller.ex
```diff
@@ -14,6 +14,19 @@ defmodule ArchEthicWeb.FaucetController do
 
   @pool_seed Application.compile_env(:archethic, __MODULE__)[:seed]
```
samuel-uniris, line 15: I think it may raise an error, if you change on runtime the `enabled` flag from env variable, because the `@pool_seed` is from compilation env.

lib/archethic_web/controllers/faucet_controller.ex
```diff
@@ -14,6 +14,19 @@ defmodule ArchEthicWeb.FaucetController do
...
 
+  plug(:enabled)
+
+  defp enabled(conn, _) do
+    if Application.get_env(:archethic, ArchEthicWeb.FaucetController)
+       |> Keyword.get(:enabled, false) do
+      conn
+    else
+      conn
```
samuel-uniris, line 24: I think you can use built-in error handling by Phoenix by using something like: `raise Phoenix.Router.NoRouteError, conn: conn, router: router`

samuel-uniris (changes requested):

lib/archethic_web/controllers/faucet_controller.ex
roychowdhuryrohit-dev: I have already looked this up but it's only available during debugging (the NoRouteError page).

roychowdhuryrohit-dev: You can test this using any random route in mainnet.archethic.com.

lib/archethic_web/controllers/faucet_controller.ex
roychowdhuryrohit-dev: If I use get_env instead, credo throws error.

lib/archethic_web/controllers/faucet_controller.ex
samuel-uniris: If you go to `https://mainnet.archethic.net/faucet` it will show `{"errors":{"detail":"Not Found"}}`.
So in dev, you will have an error page, but on the prod, you will have the JSON render

lib/archethic_web/controllers/faucet_controller.ex
samuel-uniris: you can use the specific key to load from compilation: `@pool_seed Application.compile_env(:archethic, [__MODULE__, :seed])`

lib/archethic_web/controllers/faucet_controller.ex
roychowdhuryrohit-dev: Yes. Since I was deploying this in prod I thought I would replicate the behaviour in mainnet. So do I detect mix.env() and then output different error pages based on that?

samuel-uniris: in the config.exs you have this:

```
config :archethic, ArchEthicWeb.Endpoint,
  secret_key_base: "5mFu4p5cPMY5Ii0HvjkLfhYZYtC0JAJofu70bzmi5x3xzFIJNlXFgIY5g8YdDPMf",
  render_errors: [view: ArchEthicWeb.ErrorView, accepts: ~w(json)],
  pubsub_server: ArchEthicWeb.PubSub,
  live_view: [
    signing_salt: "3D6jYvx3",
    layout: {ArchEthicWeb.LayoutView, "live.html"}
  ]
```

and in dev.exs you have a `debug` flag enabled, so in prod it will display json and on dev a specific UI

samuel-uniris: You can use something like this otherwise:
```elixir
conn                                                                                         
|> put_status(:not_found)                                                                        
|> render(ArchEthicWeb.ErrorView, "404.html")
```

roychowdhuryrohit-dev: I used the following.
```elixir
conn
|> put_status(:not_found)
|> put_view(ArchEthicWeb.ErrorView)
|> render("404.html")
|> halt()
```

This is the output during dev mode.
Screenshot 2021-09-27 at 21 16 51

roychowdhuryrohit-dev: @samuel-uniris Is this the desired behaviour? This is showing in both dev and prod.

samuel-uniris (approved):

PR #116, opened 2021-10-11
Live Stream Transactions Beacon explorer

### LINKED ISSUE #86: [Feature]: Support live transaction streaming of beacon chain summary
The BeaconChain's explorer should support a feature to stream the transactions submitted to the beacon summary in near real time.

The idea would be to create processes responsible for creating asynchronous communication with the pool of storage node for the next summary.  Those nodes will notify clients about new transaction arrived (note: this approach will be less intensive than a polling mechanism).

Each Beacon's subset can have a list of subscribers for summary updates

The feature should support the switch of pool when there is a new summary timer event. (pools may change for every interval).

When those processes received a notification of new transaction, they will submit a message through PubSub to update the UI elements.

imnik11:
Fixes #86 

Live Feed transaction directly from Beacon Pool

Enhances Beacon Chain Explorers ability like Oracle Chain Explorer to display real time transactions


config/dev.exs
```diff
@@ -1,6 +1,5 @@
 import Config
-
-config :logger, level: System.get_env("ARCHETHIC_LOGGER_LEVEL", "debug") |> String.to_atom()
```
samuel-uniris, line 3: I think this is a forgetting ;)

lib/archethic/beacon_chain/subset.ex
```diff
@@ -330,4 +349,9 @@ defmodule ArchEthic.BeaconChain.Subset do
     }
     |> ValidationStamp.sign()
   end
+
+  def subscribe_for_beacon_updates(subset, node_public_key) do
```
samuel-uniris, line 353: Can you add some typespec for public functions and maybe some documentation?

lib/archethic_web/live/chains/beacon_live.ex
```diff
@@ -135,4 +192,18 @@ defmodule ArchEthicWeb.BeaconChainLive do
     |> SummaryTimer.previous_summaries()
     |> Enum.sort({:desc, DateTime})
   end
+
+  defp register_to_beacon_pool_updates do
+    date = BeaconChain.next_summary_date(DateTime.utc_now())
+
+    Enum.map(BeaconChain.list_subsets(), fn subset ->
```
samuel-uniris, line 199: I don't think you really need Enum.map , maybe Enum.each

samuel-uniris (changes requested):

lib/archethic/beacon_chain/subset.ex
```diff
@@ -187,10 +201,17 @@ defmodule ArchEthic.BeaconChain.Subset do
...
+        state,
+        :current_slot,
+        %Slot{subset: subset, slot_time: next_time}
+      )
+
     Map.put(
-      state,
-      :current_slot,
-      %Slot{subset: subset, slot_time: next_time}
+      new_state,
```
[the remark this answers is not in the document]
imnik11: I think I changed this in https://github.com/archethic-foundation/archethic-node/pull/116/files/5f6d1e3521fd1ab2ccbeba0eae4f42173ceda898

lib/archethic/beacon_chain/subset.ex
imnik11: definitely docs is must and also spec

lib/archethic_web/live/chains/beacon_live.ex
imnik11: Ok. got it

PR #123, opened 2021-10-28
Add node IP validation in prod

roychowdhuryrohit-dev:
Added filter for private IP ranges
Added node duplication check
Added testcases

### DIFF
--- a/config/prod.exs
+++ b/config/prod.exs
@@ -189,6 +189,8 @@ config :archethic, ArchEthic.P2P.BootstrappingSeeds,
   # TODO: define the default list of P2P seeds once the network will be more open to new miners
   genesis_seeds: System.get_env("ARCHETHIC_P2P_BOOTSTRAPPING_SEEDS")
 
+config :archethic, ArchEthic.Mining.PendingTransactionValidation, validate_connection: true
+
 config :archethic, ArchEthicWeb.FaucetController,
   enabled: System.get_env("ARCHETHIC_NETWORK_TYPE") == "testnet"
 
--- a/lib/archethic/mining/pending_transaction_validation.ex
+++ b/lib/archethic/mining/pending_transaction_validation.ex
@@ -8,7 +8,7 @@ defmodule ArchEthic.Mining.PendingTransactionValidation do
 
   alias ArchEthic.Governance
   alias ArchEthic.Governance.Code.Proposal, as: CodeProposal
-
+  alias ArchEthic.Networking
   alias ArchEthic.OracleChain
 
   alias ArchEthic.P2P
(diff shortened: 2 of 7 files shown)


lib/archethic/mining/pending_transaction_validation.ex
```diff
@@ -114,15 +116,17 @@ defmodule ArchEthic.Mining.PendingTransactionValidation do
          },
          previous_public_key: previous_public_key
        }) do
-    with {:ok, _, _, _, _, key_certificate} <- Node.decode_transaction_content(content),
+    with {:ok, ip, port, _, _, key_certificate} <- Node.decode_transaction_content(content),
          root_ca_public_key <- Crypto.get_root_ca_public_key(previous_public_key),
+         true <- validate_node?(@validate_node, ip, port, previous_public_key),
```
samuel-uniris, line 121: maybe renaming it to `valid_ip?`, so this will be more meaningful

lib/archethic/mining/pending_transaction_validation.ex
```diff
@@ -114,15 +116,17 @@ defmodule ArchEthic.Mining.PendingTransactionValidation do
...
          true <-
            Crypto.verify_key_certificate?(
              previous_public_key,
              key_certificate,
              root_ca_public_key
            ) do
       :ok
+      content
```
samuel-uniris, line 129: What is this addition?

lib/archethic/p2p.ex
```diff
@@ -538,4 +538,21 @@ defmodule ArchEthic.P2P do
         do_reply_atomic(rest, message, compare_fun)
     end
   end
+
```
samuel-uniris, line 541: If you can add some documentation here, it should be great!

samuel-uniris (changes requested):

lib/archethic/mining/pending_transaction_validation.ex
roychowdhuryrohit-dev: We are already using `Networking.validate_ip?(ip)`

config/prod.exs
```diff
@@ -189,6 +189,8 @@ config :archethic, ArchEthic.P2P.BootstrappingSeeds,
   # TODO: define the default list of P2P seeds once the network will be more open to new miners
   genesis_seeds: System.get_env("ARCHETHIC_P2P_BOOTSTRAPPING_SEEDS")
 
+config :archethic, ArchEthic.Mining.PendingTransactionValidation, validate_node: true
```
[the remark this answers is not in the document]
roychowdhuryrohit-dev: How about `validate_node_connection`?

samuel-uniris (approved):

PR #199, opened 2022-01-20
Metric dashboard

apoorv-2204:
please review my code . my code  is not good


lib/archethic/metrics/generalise_metric_structure.ex
```diff
@@ -0,0 +1,156 @@
+defmodule ArchEthic.Metrics.GeneraliseMetricStructure do
```
samuel-uniris, line 1: I think you can rename it to: `ArchEthic.Metrics.Parser`

lib/archethic/metrics/generalise_metric_structure.ex
```diff
@@ -0,0 +1,156 @@
...
+  @moduledoc """
+  MetricsDataParser.run() parses metrics data and convert into structured data
+  """
+
+  def run(content) do
```
samuel-uniris, line 6: you can provide also a typespec to help dialyzer catches some errors.
As well you can include doctest to ensure the parsing works as expected

lib/archethic/metrics/metric_client.ex
```diff
@@ -0,0 +1,51 @@
+defmodule ArchEthic.Metrics.MetricClient do
+  @moduledoc """
+  gensever
```
samuel-uniris, line 3: If you can put a more meaning full description, it will be better, I think

lib/archethic/metrics/metric_client.ex
```diff
@@ -0,0 +1,51 @@
...
+  def get_this_node_points() do
+    ArchEthic.Metrics.MetricNodePoller.get_points()
+  end
+
+  def get_network_points() do
+    ArchEthic.Metrics.MetricNetworkPoller.get_points()
+  end
+
+  def monitor() do
+    GenServer.call(@process_name, :monitor)
```
samuel-uniris, line 30: You can use directly `__MODULE__`

lib/archethic/metrics/metric_client.ex
```diff
@@ -0,0 +1,51 @@
...
+  @process_name __MODULE__
+  @default_state %{
+    counter: 0
+  }
+  def start_link(_init_arg) do
+    GenServer.start_link(__MODULE__, @default_state, name: @process_name)
+  end
+
+  def init(_state) do
+    {:ok, @default_state}
```
samuel-uniris, line 18: also the default state can be put directly.
Module attributes are useful when you want something at compilation time and or reused in the module

lib/archethic_web/templates/layout/root.html.eex
```diff
@@ -73,6 +80,32 @@
...
+                    }
+                    100% {
+                      background-position: 0% 50%;
+                    }
+
+                  }
+
+    </style>
     <script type="text/javascript" src="<%= Routes.static_path(@conn, "/js/app.js") %>"></script>
+   <script src="https://cdn.jsdelivr.net/npm/echarts@5.2.2/dist/echarts.min.js"></script>
```
samuel-uniris, line 106: I think you can manage the dependencies through the app.js with `import`

lib/archethic_web/templates/layout/root.html.eex
```diff
@@ -22,8 +22,15 @@
...
           </a>
         </div>
-        <div id="navbar" class="navbar-menu">
+       <div id="navbar" class="navbar-menu">
+        <a class="navbar-item" href="<%= Routes.live_path(@conn, ArchEthicWeb.NodeStats) %>">
+             thisNode
+            </a>
+
           <div class="navbar-end">
+            <a class="navbar-item" href="<%= Routes.live_path(@conn, ArchEthicWeb.NetworkStats) %>">
```
samuel-uniris, line 31: I think you can rename it to `Metrics` and keep only the network based metrics

lib/archethic_web/router.ex
```diff
@@ -25,6 +25,8 @@ defmodule ArchEthicWeb.Router do
     get("/", RootController, :index)
     get("/up", UpController, :up)
     get("/metrics", MetricsController, :index)
+    live("/this-Node", NodeStats)
+    live("/Network", NetworkStats)
```
samuel-uniris, line 29: Maybe use a route like: `/metrics/dashboard`

lib/archethic_web/router.ex
```diff
@@ -25,6 +25,8 @@ defmodule ArchEthicWeb.Router do
     get("/", RootController, :index)
     get("/up", UpController, :up)
     get("/metrics", MetricsController, :index)
+    live("/this-Node", NodeStats)
```
samuel-uniris, line 28: this one may not be relevant I think

lib/archethic_web/live/network_stats.ex
```diff
@@ -0,0 +1,66 @@
+defmodule ArchEthicWeb.NetworkStats do
```
samuel-uniris, line 1: Maybe rename it, to: `ArchEthicWeb.NetworkMetricsLive`

lib/archethic_web/live/network_stats.ex
```diff
@@ -0,0 +1,66 @@
...
+  @moduledoc false
+
+  use Phoenix.LiveView
+  use Phoenix.HTML
+
+  def mount(_params, _session, socket) do
+    if connected?(socket) do
+      ArchEthic.Metrics.MetricClient.monitor()
+      :timer.send_interval(5_000, self(), :update)
```
samuel-uniris: I think it may be better to let the polling process notifying the liveview and pushing instead of pulling the points each 5s

lib/archethic_web/live/network_stats.ex
```diff
@@ -0,0 +1,66 @@
...
+    end
+
+    {:ok, socket}
+  end
+
+  def handle_info(:update, socket) do
```
samuel-uniris, line 16: Then by pushing, you should get the points in the message avoiding a GenServer.call

lib/archethic/metrics/network_metric.ex
```diff
@@ -0,0 +1,40 @@
+defmodule ArchEthic.Metrics.NetworkMetric do
```
samuel-uniris, line 1: Maybe rename it: `ArchEthicWeb.Metrics.NetworkCollector`

lib/archethic/metrics/metric_client.ex
```diff
@@ -0,0 +1,51 @@
+defmodule ArchEthic.Metrics.MetricClient do
```
samuel-uniris, line 1: I think you can merge this module and `ArchEthic.Metrics.NetworkPoller` into `ArchEthicWeb.Metrics.Poller`

lib/archethic/metrics/metric_client.ex
```diff
@@ -0,0 +1,51 @@
...
+
+  def get_this_node_points() do
+    ArchEthic.Metrics.MetricNodePoller.get_points()
+  end
+
+  def get_network_points() do
+    ArchEthic.Metrics.MetricNetworkPoller.get_points()
+  end
+
+  def monitor() do
```
samuel-uniris, line 29: maybe rename it to `subscribe_to_updates`

lib/archethic/metrics/metric_client.ex
```diff
@@ -0,0 +1,51 @@
...
+    GenServer.call(@process_name, :monitor)
+  end
+
+  def handle_call(:monitor, {from_pid, _ref}, state) do
+    ArchEthic.Metrics.MetricNodePoller.set_flag()
+    ArchEthic.Metrics.MetricNetworkPoller.set_flag()
+    IO.inspect "metric_client"
+    IO.inspect state
+    _mref = Process.monitor(from_pid)
+    {:reply, :ok, %{state | counter: state.counter + 1}}
```
samuel-uniris, line 42: I think it should be better to register the client pid.
Therefore, we will get a list of clients processes.
It's every time good to keep a track of the registered processes to monitor.
Then you want to stop the poller, just check if the processes list is empty.

lib/archethic/metrics/metric_network_poller.ex
```diff
@@ -0,0 +1,76 @@
...
+    {:noreply, new_state}
+  end
+
+  def handle_call(:get_points, _from, current_state) do
+    response = current_state.points
+    new_state = current_state
+    {:reply, response, new_state}
+  end
+
+  def handle_info(:periodic_calculation_of_points, current_state) do
```
samuel-uniris, line 54: I think you can leverage pattern matching from the process state instead of `case` statement.
It will create a better readability.

lib/archethic/metrics/metric_helper_functions.ex
```diff
@@ -0,0 +1,381 @@
+defmodule ArchEthic.Metrics.MetricHelperFunctions do
```
samuel-uniris, line 1: I think you should leverage doctests  to validate your functions, as they are business logic oriented for the metrics point of view.

Maybe rename it to: `ArchEthicWeb.Metrics.Helpers`

assets/js/app.js
```diff
@@ -72,7 +72,344 @@ Hooks.Logs = {
   }
 }
 
+function generateEchartObjects(heading , echartContainer ,  x_axis_data){
```
samuel-uniris, line 75: You should extract all of this logic into a dedicated file, so you can import it into the app.js

samuel-uniris (changes requested):
Thanks for the PR.
I propose several changes to have a better architecture and maintainability and readability.
Otherwise good bases and foundations for this first PR and complex one 👍

lib/archethic/metrics/metric_client.ex
apoorv-2204: I did but I removed due to urgency of newsletter

lib/archethic/metrics/metric_client.ex
apoorv-2204: noted

lib/archethic/metrics/metric_client.ex
apoorv-2204: noted

lib/archethic/metrics/metric_client.ex
apoorv-2204: But the data of grow  will grow on large number of user . ??

assets/js/app.js
apoorv-2204: noted

lib/archethic/metrics/metric_client.ex
apoorv-2204: I removed all the the IO.inspect  , this might be an older commit.

samuel-uniris: this is no big brainer, as the clients will be removed once they disconnect

PR #207, opened 2022-01-27
Change faucet success flash link to tx URL, fix #204

### LINKED ISSUE #204: [Feature]: Redirect faucet success alert to the transaction address
### Problem to solve

When a valid transaction is submitted through the faucet controller, the alert redirects to the recipient address but not to the new transaction address .

### Solution

We can easily redirect to the generated transaction address to have a better UX and responsiveness of feedback and audit of the transactions.

roychowdhuryrohit-dev:
Faucet success flash now points to transaction address instead of recipient address
Added testcases for faucet controller create_transfer/2
Fixed error pattern matching in faucet controller to better detect network errors

### DIFF
--- a/lib/archethic_web/controllers/faucet_controller.ex
+++ b/lib/archethic_web/controllers/faucet_controller.ex
@@ -40,13 +40,13 @@ defmodule ArchEthicWeb.FaucetController do
   def create_transfer(conn, %{"address" => address}) do
     with {:ok, recipient_address} <- Base.decode16(address, case: :mixed),
          true <- Crypto.valid_hash?(recipient_address),
-         :ok <- transfer(recipient_address) do
+         {:ok, tx_address} <- transfer(recipient_address) do
       conn
       |> put_resp_header("cache-control", "no-cache, no-store, must-revalidate")
       |> put_resp_header("pragma", "no-cache")
       |> put_resp_header("expires", "0")
       |> put_flash(:info, "Transferred successfully (click to view)")
-      |> render("index.html", address: "", link_address: address)
+      |> render("index.html", address: "", link_address: Base.encode16(tx_address))
     else
       {:error, _} ->
         conn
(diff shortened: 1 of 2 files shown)


lib/archethic_web/controllers/faucet_controller.ex
```diff
@@ -40,22 +40,22 @@ defmodule ArchEthicWeb.FaucetController do
   def create_transfer(conn, %{"address" => address}) do
     with {:ok, recipient_address} <- Base.decode16(address, case: :mixed),
          true <- Crypto.valid_hash?(recipient_address),
-         :ok <- transfer(recipient_address) do
+         {tx_address, :ok} <- transfer(recipient_address) do
```
samuel-uniris, line 43: The convention in Elixir/Erlang is to use `{:ok, result} ` where `result` is the data returned by the function. But the `ok` atom is specified in the beginning.

lib/archethic_web/controllers/faucet_controller.ex
```diff
@@ -40,22 +40,22 @@ defmodule ArchEthicWeb.FaucetController do
...
       conn
       |> put_resp_header("cache-control", "no-cache, no-store, must-revalidate")
       |> put_resp_header("pragma", "no-cache")
       |> put_resp_header("expires", "0")
       |> put_flash(:info, "Transferred successfully (click to view)")
-      |> render("index.html", address: "", link_address: address)
+      |> render("index.html", address: "", link_address: tx_address)
     else
-      {:error, _} ->
+      e when e in [:error, false] ->
```
samuel-uniris, line 51: I think you should keep the previous checking, with `{:error, _}` as before and use `_` for the malformed address, as it will be more explicit and easy to understand

lib/archethic_web/controllers/faucet_controller.ex
```diff
@@ -73,30 +73,34 @@ defmodule ArchEthicWeb.FaucetController do
...
-      curve
-    )
-    |> ArchEthic.send_new_transaction()
+        },
+        @pool_seed,
+        transaction_index,
+        curve
+      )
+
+    {Base.encode16(tx.address),
```
samuel-uniris, line 102: You should encode before to render the view, here it's more a business function, so keep in binary. 
Also you should return `{:ok, tx.address}` only when the transaction is submitted

samuel-uniris (changes requested):

lib/archethic_web/controllers/faucet_controller.ex
roychowdhuryrohit-dev: I discovered that in a "network_error" event, I was getting `Malformed address` error using the earlier approach.

roychowdhuryrohit-dev: During that error the e value was like this 
> {"0098fe10e8633bce19c59a40a089731c1f72b097c5a8f7dc71a37eb26913aa4f80",{:error,....}}

PR #219, opened 2022-02-15
Use a FSM in the OracleChain Scheduler

samuel-uniris:
By using a Finite State Machine we are gaining more control of the
events triggered and the transitions between states.

This way we can fix the oracle chain duplication as we know how to
trigger polling events and therefore dates for the key derivation.

### DIFF
--- a/lib/archethic/oracle_chain/scheduler.ex
+++ b/lib/archethic/oracle_chain/scheduler.ex
@@ -15,22 +15,19 @@ defmodule ArchEthic.OracleChain.Scheduler do
   alias ArchEthic.OracleChain.Services
   alias ArchEthic.OracleChain.Summary
 
-  alias ArchEthic.TaskSupervisor
-
   alias ArchEthic.TransactionChain
   alias ArchEthic.TransactionChain.Transaction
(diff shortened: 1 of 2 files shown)


lib/archethic/oracle_chain/scheduler.ex
```diff
@@ -284,13 +308,16 @@ defmodule ArchEthic.OracleChain.Scheduler do
...
 
-      Logger.debug("New data pushed to the oracle")
-    end
+    Task.start(fn -> ArchEthic.send_new_transaction(tx) end)
+
+    Logger.debug("New data pushed to the oracle",
+      transaction_address: Base.encode16(tx.address),
+      transaction_type: :oracle
+    )
   end
```
imnik11, line 319: maybe we can add spec & some doc here and summary

imnik11: works fine

samuel-uniris: > maybe we can add spec & some doc here and summary

Forgot the put in private . Thanks

imnik11 (approved):

PR #227, opened 2022-02-17
165 transaction content max size

### LINKED ISSUE #165: [Feature]: Set the maximum size of the transaction's content
### Problem to solve

To avoid the exhaustion of the network in terms of latency and disk usage, we need to limit the transaction's content section size.

### Solution

When a welcome node will receive a transaction, it will assert the transaction's content does not exceed for example`3MB` (but with could specify the limit size from the configuration)
Therefore, validation nodes can ensure the verification of this size during the transaction validation.

blackode:
# Description
In this change, we are keeping the **threshold** limit for the size of the `content` in the `data` field of the `transaction`. This means we are not allowing the content size to be greater than the configured threshold limit.

Add validation functions for the changesest `TransactionPayload` and `TransactionData` structs.

A new config line was added as follows

```elixir
config :archethic, :transaction_data_content_max_size, 3_145_728
```

Fixes #165 

## Type of change

- [x] New feature (setting up the max-content size )

# How Has This Been Tested?

Creating the content size greater than content_max_size using the LOC 
```elixir
Base.encode16(:crypto.strong_rand_bytes(4 * 1024 * 1024))
```` 
and trying to create a changeset of the transaction payload.

- [x] Test  **API** should return an error if the content size is greater than the `content_max_size`
- [x] Test  **Mining**  should return an error if the content size is greater than the `content_max_size`

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -46,7 +46,7 @@ config :archethic, :mut_dir, "data"
 config :archethic, :marker, "-=%=-=%=-=%=-"
 
 # size represents in bytes binary
-config :archethic, :transaction_data_content_max_size, 3145728
+config :archethic, :transaction_data_content_max_size, 3_145_728
 
 config :archethic, ArchEthic.Crypto,
   supported_curves: [
--- a/config/config.exs
+++ b/config/config.exs
@@ -45,8 +45,8 @@ config :archethic, :mut_dir, "data"
 
 config :archethic, :marker, "-=%=-=%=-=%=-"
 
-# size represents in mb i.e 1024 * 1024  * 3 bytes
-config :archethic, :transaction_data_content_max_size, 3
+# size represents in bytes binary
+config :archethic, :transaction_data_content_max_size, 3145728
 
 config :archethic, ArchEthic.Crypto,
   supported_curves: [
--- a/config/config.exs
+++ b/config/config.exs
@@ -45,6 +45,9 @@ config :archethic, :mut_dir, "data"
 
 config :archethic, :marker, "-=%=-=%=-=%=-"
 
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: config/config.exs [hunk 1]
```diff
@@ -45,6 +45,9 @@ config :archethic, :mut_dir, "data"
 
 config :archethic, :marker, "-=%=-=%=-=%=-"
 
+# size represents in mb i.e 1024 * 1024  * 3 bytes
+config :archethic, :transaction_data_content_max_size, 3
```
samuel-uniris, line 49: What do you think of putting the number in bytes instead of megabytes ? 
This will avoid the division in the changeset to convert bytes to mb .

lib/archethic_web/controllers/api/schema/transaction_data.ex
```diff
@@ -1,5 +1,6 @@
 defmodule ArchEthicWeb.API.Schema.TransactionData do
   @moduledoc false
+  @content_max_size Application.get_env(:archethic, :transaction_data_content_max_size)
```
samuel-uniris, line 3: If it is a compile configuration, you should use `Application.compile_env`

samuel-uniris (approved):

PR #236, opened 2022-02-21
WIP : Metric Dashboard : unit testing 

apoorv-2204:
Why this pr?
 - to remove the test data from metric Dashboard ui
 - add doc tests


apoorv-2204:
Request for urgent Review and Merge , before pushing updates to nodes. @samuel-uniris  , Critical involving brand image of archethic as test data  was no removed from ui,

assets/js/app.js
```diff
@@ -74,17 +74,15 @@ Hooks.Logs = {
...
+//metric dashboard hook /metrics/dashboard
 Hooks.network_charts = {
   mounted() {
 
     var network_metric_obj = metric_config_obj.create_network_live_visuals();
     this.handleEvent("network_points", ({
       points
     }) => {
-      console.log("---------------")
       console.log(points);
```
samuel-uniris, line 85: I think you can remove also the console.log here

assets/js/app.js
```diff
@@ -104,11 +102,7 @@ Hooks.explorer_charts = {
         points
       }) => {
         console.log(points);
```
samuel-uniris, line 104: The same goes here

apoorv-2204:
thanks for immediate review @samuel-uniris

apoorv-2204:
Hi ,
@samuel-uniris 
I tried mox , dropped it . Case test don't fit I think.
In doc test the input data is too big . 
and output too big  to be mentioned in that module doc 
,or even to be formatted to be place in metric module.
please suggest some measure to counter it.
please share your views on  property based testing.
thanks

samuel-uniris:
For doctest, you shoud leverage function which are pure, where the output is only based on the input.
For example: 
- https://github.com/archethic-foundation/archethic-node/blob/3d3e5dae5c55e67d2c6e127104eecfea7df2c49f/lib/archethic/metrics/parser.ex#L6
- https://github.com/archethic-foundation/archethic-node/blob/3d3e5dae5c55e67d2c6e127104eecfea7df2c49f/lib/archethic/metrics/helpers.ex#L425
- https://github.com/archethic-foundation/archethic-node/blob/3d3e5dae5c55e67d2c6e127104eecfea7df2c49f/lib/archethic/metrics/helpers.ex#L440

But for those which are based on external data or mutated data, we can use simple test with Mock for instance:
- https://github.com/archethic-foundation/archethic-node/blob/3d3e5dae5c55e67d2c6e127104eecfea7df2c49f/lib/archethic/metrics/poller.ex#L1
- https://github.com/archethic-foundation/archethic-node/blob/3d3e5dae5c55e67d2c6e127104eecfea7df2c49f/lib/archethic/metrics/helpers.ex#L89

However, you have plenty of public functions in the `/lib/archethic/metrics/helpers.ex` which are used a private functions. I suggest you either to make them private or externalize them into other modules if you still want to use doctests.

But for doctests length, you are free to use the input data, so you make the test example shorter.
Nevertheless, the test should be meaningful regarding the implementation, to ensure the test stay consistent with the code.
In the code base, we used some doctest which are not short, sometime a bit long, so you can leverage that example.

apoorv-2204:
Limited data was taken as input, still escape characters are creating issue in creating doc test.

lib/archethic/metrics/parser.ex
```diff
@@ -5,6 +5,79 @@ defmodule ArchEthic.Metrics.Parser do
...
+          ...>     archethic_election_storage_nodes_duration_bucket{nb_nodes='2',le='0.001'} 120
+          ...>     archethic_election_storage_nodes_duration_sum{nb_nodes='2'} 0.011
+          ...>     archethic_election_storage_nodes_duration_count{nb_nodes='2'} 122
+          ...>     archethic_election_storage_nodes_duration_sum{nb_nodes='3'} 3.11
+          ...>     archethic_election_storage_nodes_duration_count{nb_nodes='3'} 45719
+          ...>     # HELP vm_memory_atom
+          ...>     # TYPE vm_memory_atom gauge
+          ...>     vm_memory_atom 1589609
+          ...>     """
+          iex> length(ArchEthic.Metrics.Parser.run(data))==Enum.count(ArchEthic.Metrics.Parser.run(data),fn x ->
```
samuel-uniris, line 63: I think you can simplify, but just expecting the real result, without the Enum.count fn, etc...

samuel-uniris: Also If you want you can provide different test cases, maybe one with histogram, one with gauge,  one with labels, ..., 
this will reduce the size of the test

apoorv-2204:
Hi 
, how do I test for periodic function in genserver?
Creating a start_supervised!  is reporting error  already started.

apoorv-2204:
Hello sir how to spawn a dummy process that dont die , until we pass a message

describe "ArchEthic.Metrics.Poller" do

    test "logice and flow" do
      # spawn Poller with different name
      start_options = [
        options: [name: ArchEthic.Metrics.Poller_dummy],
        default_state: ArchEthic.Metrics.Helpers.poller_default_state()
      ]
      dummy_poller_pid = start_supervised!({ArchEthic.Metrics.Poller, start_options})
      # assert initial state be default state
      assert :sys.get_state(dummy_poller_pid) == ArchEthic.Metrics.Helpers.poller_default_state()

      # create a dummy client
      Process.flag(:trap_exit, true)
      dummy_client_pid  =  spawn(fn ->
            ArchEthic.Metrics.Poller.monitor()
                loop()
            end)

            IO.inspect(:sys.get_state(dummy_poller_pid))
      # assert Map.fetch( :sys.get_state(dummy_poller_pid).pid_refs , dummy_client_pid) == nil

      # Process.kill(dummy_client_pid)
      # assert Map.fetch( :sys.get_state(dummy_poller_pid).pid_refs , pid) == :error

    end

  end

  def loop() do
    receive do
     _ -> nil
     end
      loop()
    end

samuel-uniris:
> Hi 
> , how do I test for periodic function in genserver?
> Creating a start_supervised!  is reporting error  already started. 

This because the process is named by default. If you want to spawn multiple you should not name it by default. Look at the start_link function

test/archethic_test.exs
```diff
@@ -412,27 +412,24 @@ defmodule ArchEthicTest do
       assert :sys.get_state(dummy_poller_pid) == ArchEthic.Metrics.Helpers.poller_default_state()
 
       # create a dummy client
-      Process.flag(:trap_exit, true)
-      dummy_client_pid  =  spawn(fn ->
-            ArchEthic.Metrics.Poller.monitor()
-                loop()
-            end)
+      dummy_client_pid  =  Task.start(fn ->  ArchEthic.Metrics.Poller_dummy.monitor
+       loop()    end)
```
apoorv-2204: I  did add here. , but I am confused , that it requires to start Poller with different name , but how to contact that same dummy poller to invoke its method monitor

lib/archethic/metrics/poller.ex
```diff
@@ -2,16 +2,14 @@ defmodule ArchEthic.Metrics.Poller do
...
-    pid_refs: %{},
-    data: ArchEthic.Metrics.Helpers.get_client_metric_default_value()
-  }
-
-  def start_link(_state) do
-    GenServer.start_link(__MODULE__, @default_state, name: __MODULE__)
+  def start_link(opts) do
+    options = Keyword.fetch!(opts, :options)
+    default_state = Keyword.fetch!(opts, :default_state)
+    GenServer.start_link(__MODULE__, default_state, options)
```
apoorv-2204, line 12: Hi , I did change the parameters for start_link

lib/archethic/metrics/poller.ex
```diff
@@ -23,45 +21,21 @@ defmodule ArchEthic.Metrics.Poller do
...
-      case Enum.empty?(current_state.pid_refs) do
-        false ->
-          send_updates(current_state)
-          %{data: get_new_data(), pid_refs: current_state.pid_refs}
-
-        true ->
-          current_state
-      end
-
+    new_state = Helpers.process_new_state(current_state)
```
apoorv-2204, line 37: Hi ,
@samuel-uniris 
How can I test for external dependency which is in another module and being called by genserver Poller

apoorv-2204: def process_new_state(current_state) do
    case Enum.empty?(current_state.pid_refs) do
      false ->
        dipatch_updates(current_state)
        %{data: network_collector(), pid_refs: current_state.pid_refs}

      true ->
        current_state
    end
  end

samuel-uniris (approved):

Issue #225, opened 2022-02-16
Add worldmap for the node distribution


samuel-uniris:
The explorer should include a view/map to distribute the node around the world using the `geo_patch` provided by the network.
- [ ] Create an algorithm to transform a geographical patch into a `x/y` coordinates
- [ ] Create the JavaScript component to display it on the node explorer

PR #238, opened 2022-02-21
Add node worldmap in explorer

Neylix:
### Description
Working on issue #225

Implement first task :

- [x] Create an algorithm to transform a geographical patch into a x/y coordinates
- [x] Create the JavaScript component to display it on the node explorer

I noticed that some coordinates cannot compute a geo patch. For exemple when longitude is equal to -45, sdc and tdc are equal to 1 or -1 and then are not in range for the index_patch function

Test are just on 4 geo patch, but I tested all and it works fine (cannot test all in geo_patch_test cause compute_patch is private)

I'll work soon on second task to add map on the explorer, I think to use [Leaflet](https://leafletjs.com/) to display the map


lib/archethic/p2p/geo_patch.ex
```diff
@@ -93,4 +93,127 @@ defmodule ArchEthic.P2P.GeoPatch do
       x2 / 2
     end
   end
+
+  @doc """
+  Get range of longitude / latitude coordinates from geo patch
+  """
+  @spec from_geo_patch(binary()) :: list(list(float()))
```
samuel-uniris, line 100: Firstly, thanks for the contribution
But I would suggest renaming it `to_coordinates`, as it will bring more meaning to the function without reading it.
Also, the returning list can be a tuple I think, as it will be a fixed size.

samuel-uniris:
For the UI, what do you think of:  
- https://echarts.apache.org/examples/en/editor.html?c=effectScatter-bmap
- https://telemetry.polkadot.io/#map/0x91b171bb158e2d3848fa23a9f1c25182fb8e20313b2c1eb49219da7a70ce90c3
- https://solanabeach.io/

Something more graphical should be great.
But however, thanks for the contribution 👍

lib/archethic/p2p/geo_patch.ex
Neylix: Updated in https://github.com/archethic-foundation/archethic-node/pull/238/commits/5054594bb22ca132aa8dcbdab05aac81a2adcb85

Neylix:
Thanks for reply !

Using echarts is a good idea as it is already imported and used in telemetry.

However using points on the map is not interesting because the range of coordinates is to large. Here the worldmap with all ranges displayed : 

image

So we cannot have precise point on the map, unless if we get the exact coordinates from nodes instead of getting a range using geo patch.

In all case, I think that displaying the range is a good idea as it is an important variable for the ARCH consensus.

What do you think about getting node exact coordinates and grouping them with a range of +-1 and so have good precision to diplay them as a point on the map like in your exemple ?

samuel-uniris:
Hello.
Thanks for the changes and the proposals.
In fact GeoPatch are designed with sub levels of details.
So each patch have 3 digits.
Each digit corresponds to a subdivision of 16 squares.
Example: 012, it will be located on the first square, then inside it will go in the  second nested square and finally it will to the third nested  square of the latter.
This way we have some precision but without giving too much information.

Also if there are too many points we can aggregate or if for UI clarity it will not render nicely, we can keep two digits and show popup with the number of nodes inside.

Neylix:
Hi ! 

Correct me if I'm wrong but I think that each digit do not subdivise by 16 the possible range, the first one yes, but the 2 others only subdivise by 4 :
Depending on real coordinates, the first digit can be all possiblities (from 0 to F)
Then, due to function ``` resolve_with_sign() ``` the second digit must have ``` f_i ``` and ``` s_i ``` with the same sign as the original coordinates so it let open only 4 possiblities.
For exemple if first digit is "0" so ``` f_i ``` positive and ``` s_i ``` negative, second digit must have same sign so it remain only "0", "1", "4", "5".
It's same thing for the third digit.

So with that said, we can have only 256 geo patch (16 * 4 * 4).
I tested this with "brute force" method (calculate all coordinates possiblities incrementing by 0.5) and the result is 256 unique geo patch

If I'm right, the final subdivision looks like the map I presented in my last comment (16 by 16 subdivisions)

samuel-uniris:
> Hi !
> 
> Correct me if I'm wrong but I think that each digit do not subdivise by 16 the possible range, the first one yes, but the 2 others only subdivise by 4 : Depending on real coordinates, the first digit can be all possiblities (from 0 to F) Then, due to function `resolve_with_sign()` the second digit must have `f_i` and `s_i` with the same sign as the original coordinates so it let open only 4 possiblities. For exemple if first digit is "0" so `f_i` positive and `s_i` negative, second digit must have same sign so it remain only "0", "1", "4", "5". It's same thing for the third digit.
> 
> So with that said, we can have only 256 geo patch (16 * 4 * 4). I tested this with "brute force" method (calculate all coordinates possiblities incrementing by 0.5) and the result is 256 unique geo patch
> 
> If I'm right, the final subdivision looks like the map I presented in my last comment (16 by 16 subdivisions)

Hello.
If I'm remember well, normally it should be 16x16 squares at least for the two first digits, and probably also for the third.
@I-Archer-Zero Can you confirm ?
If so, we would have to change the coordinates to geo patch algorithm.

Neylix:
Hi @samuel-uniris @I-Archer-Zero 
Do you have an update on this subject ? 
Is there a need to rework on geo patch algorithm ?

I-Archer-Zero:
Hello @Neylix, @samuel-uniris 

Here is the full solution to implement the Geopatch ID for each node and then plot all the nodes on the world map

image
FYI: BASED ON THE NUMBER OF NODES IN A PATCH, THE DENSITY OF THE COLOUR OF THAT PATCH IS TUNED DEPENDING ON THE NUMBER OF NODES IN IT.

Geolocation patch for a node is represented by 3 Digits, for example “XYZ”  where 
- X is the latitude patch id
- Y is the longitude patch id
- Z is the node’s availability (to be obtained from the nodes list)

For example on the graph where a node is represented by + the patch id of the node is 96D, where 
- 9 is the latitude patch id
- 6 is the longitude patch id
- D is the node’s availability (obtained from the nodes list)

The ID values can vary from [0,1,2,3,4,5,6,7,8,9,A,B,C,D,E,F] 

Steps to calculate the patch id of a node
Step 1: Obtain the Geolocation represented in terms of Latitude (-90,90), longitude (-180,180) 
Step 2: Convert the Latitude (-90,90), Longitude (-180, 180) into (0,180), (0, 360) respectively 
Step 3: Assign Latitude Patch ID and Longitude Patch ID 
Step 4: Get the average node availability from the node list 

For example: 
<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01//EN" "http://www.w3.org/TR/html4/strict.dtd">
<html>
<head>

<meta http-equiv="Content-Style-Type" content="text/css">
<title></title>
<meta name="Generator" content="Cocoa HTML Writer">
<meta name="CocoaVersion" content="2113.2">
<style type="text/css">
p.p1 {margin: 0.0px 0.0px 0.0px 0.0px; font: 13.0px '.AppleSystemUIFont'}
table.t1 {border-collapse: collapse}
td.td1 {border-style: solid; border-width: 1.0px 1.0px 1.0px 1.0px; border-color: #9a9a9a #9a9a9a #9a9a9a #9a9a9a; padding: 1.0px 5.0px 1.0px 5.0px}
</style>
</head>
<body>


Node | Step 1:  Latitude (-90, 90)  Longitude (-180, 180) | Step 2: Latitude (0, 180)  Longitude (0, 360) | Step 3: Assign Lat, long ID | Step 4: Patch ID
-- | -- | -- | -- | --
Node 1 | (10, 45) | (190, 135) | 96 | 96A
Node 2 | (-10, 45) | (170, 135) | 86 | 865
Node 3 | (10, -45) | (190, 45) | 92 | 92E
Node 4 | (-10, -45) | (170, 45) | 82 | 823



</body>
</html>

samuel-uniris:
Related to #253

Neylix:
How to test : 
- Testing data sent from Phoenix Live is done in worldmap_live_test.ex
- Testing maps updates in real time : run ```docker-compose up node 1``` and then run other node containers
- Testing multiple nodes display without running hundred of container : update ```worldmap_live.ex``` file and replace ```P2P.available_nodes()``` by your own datas (can create a loop to randomly generates Nodes data)

samuel-uniris (approved):

PR #240, opened 2022-02-22
Prevent network split

samuel-uniris:


samuel-uniris:
Not really just if it's to ensure most of the nodes have the same information in term of P2P and sharding consistency.

PR #246, opened 2022-02-25
Fix address curve & serialization issues

samuel-uniris:
# Description

In this change, we are fixing the serialization of adresses with the curve type prepended.

Also the faucet address validation is fixed with the new format.

# Type of change

- [x] Bug fix

# How Has This Been Tested?

Go the http://localhost:4000/faucet try to input an address without curve type, it should mention a malformed address.

Also it should not have serialization issue in the logs during the runtime period. 

Should be checked also with docker-compose.


imnik11 (approved):
- [x] Read PR description
- [x] PR manually tested
- [x] Manually tested and working as expected i.e we are able to transfer UCO to the valid_address via http://localhost:4000/faucet
- [x] Both faucet and explorer search are returning malformed address and invalid address if it is not valid_address
- [x]  Working in docker-compose as expected

PR #248, opened 2022-02-28
Added Github CI actions 

imnik11:
## Description
Adding Github CI action on every push to master branch the workflow is triggered which spawns a container with scylla and dependencies like libsodium, tpm2-tss, etc and does the following
- [x] mix clean
- [x] mix format --check-formatted
- [x] mix compile --warnings-as-errors
- [x] mix credo
- [x] mix test --trace
- [x] mix dialyzer


samuel-uniris:
Should you put the test failing tagged as CI excluded test ?
I didn't figure out yet why this is failing

PR #252, opened 2022-03-09
Fix crypto identifiers bootstrap

samuel-uniris:


apoorv-2204 (approved):
Hi. Thanks for the quick fix. I approve on the b

PR #258, opened 2022-03-14
Added Rest API for origin device public key

imnik11:
#  Description

We are adding Rest API to post origin public key with certificate along with the validation of  certificate as per the public key and  validate and Implement the transaction generation with new authorized device for renewal

- [x] post req -> to get origin public key and cert
- [x] generate new shared origin keys, and encrypt with new AES key which will be encrypted by the origin public keys (refer Ownership module)
- [x] add the TransactionData containing content, ownerships and secrets to the transaction 
- [x] add unit test 

# Type of Change 
- [x] Feature

# How Has This Been Tested?

Post Req to API Endpoint http://localhost:4000/api/origin_public_key 
Request
`
{
	"publicKey" : "00005D6B602E754079AEA9A69675BDBE2468CFAC3EE25F38FF3EE5E8974A3F779E97",
	"certificate": ""
}
`
Response
`
{
	"certificate": "valid",
	"originPublicKey": "valid"
}
`

### DIFF
--- a/lib/archethic/shared_secrets/node_renewal.ex
+++ b/lib/archethic/shared_secrets/node_renewal.ex
@@ -138,8 +138,6 @@ defmodule ArchEthic.SharedSecrets.NodeRenewal do
 
     new_origin_shared_priv_keys = :erlang.list_to_binary(new_origin_shared_priv_keys_list)
 
-    origin_public_keys_bin = :erlang.list_to_binary(origin_public_keys)
-
     encrypted_origin_shared_priv_keys =
       Crypto.aes_encrypt(new_origin_shared_priv_keys, secret_key)
 
@@ -148,7 +146,7 @@ defmodule ArchEthic.SharedSecrets.NodeRenewal do
     Transaction.new(
       :origin_shared_secrets,
       %TransactionData{
-        content: <<origin_public_keys_bin <> new_origin_shared_pub_keys>>,
+        content: new_origin_shared_pub_keys,
         ownerships: [
           Ownership.new(secret, secret_key, origin_public_keys)
         ]
--- a/lib/archethic/shared_secrets/node_renewal.ex
+++ b/lib/archethic/shared_secrets/node_renewal.ex
@@ -134,13 +134,11 @@ defmodule ArchEthic.SharedSecrets.NodeRenewal do
       new_shared_origin_keys
       |> Enum.map(fn {_pub, priv} -> priv end)
 
-    new_origin_shared_pub_keys =
-      new_origin_shared_pub_keys_list
-      |> Enum.reduce(fn x, acc -> acc <> x end)
+    new_origin_shared_pub_keys = :erlang.list_to_binary(new_origin_shared_pub_keys_list)
 
-    new_origin_shared_priv_keys =
-      new_origin_shared_priv_keys_list
-      |> Enum.reduce(fn x, acc -> acc <> x end)
+    new_origin_shared_priv_keys = :erlang.list_to_binary(new_origin_shared_priv_keys_list)
+
+    origin_public_keys_bin = :erlang.list_to_binary(origin_public_keys)
 
     encrypted_origin_shared_priv_keys =
       Crypto.aes_encrypt(new_origin_shared_priv_keys, secret_key)
@@ -150,7 +148,7 @@ defmodule ArchEthic.SharedSecrets.NodeRenewal do
     Transaction.new(
       :origin_shared_secrets,
       %TransactionData{
-        content: new_origin_shared_pub_keys,
+        content: <<origin_public_keys_bin <> new_origin_shared_pub_keys>>,
         ownerships: [
           Ownership.new(secret, secret_key, origin_public_keys)
         ]
--- a/lib/archethic/shared_secrets/node_renewal.ex
+++ b/lib/archethic/shared_secrets/node_renewal.ex
@@ -114,6 +114,35 @@ defmodule ArchEthic.SharedSecrets.NodeRenewal do
     )
   end
 
+  @spec new_origin_shared_secrets_transaction(
+          origin_public_keys :: list(Crypto.key()),
+          secret_key :: binary()
+        ) :: Transaction.t()
+  def new_origin_shared_secrets_transaction(
+        origin_public_keys \\ [<<0::0, 0::0, :crypto.strong_rand_bytes(32)::binary>>],
+        secret_key \\ :crypto.strong_rand_bytes(32)
+      )
+      when is_binary(secret_key) and
+             is_list(origin_public_keys) do
+    {new_shared_origin_pub_key, new_shared_origin_priv_key} = Crypto.generate_random_keypair()
+    origin_public_key = Enum.at(origin_public_keys, 0)
+
+    encrypted_origin_shared_priv_key =
+      Crypto.ec_encrypt(new_shared_origin_priv_key, origin_public_key)
+
+    secret = encrypted_origin_shared_priv_key
+
+    Transaction.new(
+      :origin_shared_secrets,
+      %TransactionData{
+        content: new_shared_origin_pub_key,
+        ownerships: [
+          Ownership.new(secret, secret_key, origin_public_keys)
+        ]
+      }
+    )
+  end
+
   @doc """
   Decode the transaction content from the node renewal transaction
   """
--- a/lib/archethic/shared_secrets/node_renewal.ex
+++ b/lib/archethic/shared_secrets/node_renewal.ex
@@ -119,23 +119,38 @@ defmodule ArchEthic.SharedSecrets.NodeRenewal do
           secret_key :: binary()
         ) :: Transaction.t()
   def new_origin_shared_secrets_transaction(
-        origin_public_keys \\ [<<0::0, 0::0, :crypto.strong_rand_bytes(32)::binary>>],
-        secret_key \\ :crypto.strong_rand_bytes(32)
+        origin_public_keys,
+        secret_key
       )
       when is_binary(secret_key) and
              is_list(origin_public_keys) do
-    {new_shared_origin_pub_key, new_shared_origin_priv_key} = Crypto.generate_random_keypair()
-    origin_public_key = Enum.at(origin_public_keys, 0)
+    new_shared_origin_keys = 1..10 |> Enum.map(fn _ -> Crypto.generate_random_keypair() end)
 
-    encrypted_origin_shared_priv_key =
-      Crypto.ec_encrypt(new_shared_origin_priv_key, origin_public_key)
+    new_origin_shared_pub_keys_list =
+      new_shared_origin_keys
+      |> Enum.map(fn {pub, _priv} -> pub end)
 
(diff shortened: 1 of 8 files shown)


lib/archethic_web/controllers/api/origin_public_key_payload.ex
```diff
@@ -0,0 +1,49 @@
...
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  alias ArchEthic.Utils
+
+  alias ArchEthicWeb.API.Types.Hex
+  alias ArchEthicWeb.API.Types.PublicKey
+
+  embedded_schema do
+    field(:PublicKey, PublicKey)
```
samuel-uniris, line 13: Please use camelCase standard to be consistent

lib/archethic_web/controllers/api/origin_public_key_payload.ex
```diff
@@ -0,0 +1,49 @@
...
+    |> validate_required([
+      :PublicKey
+    ])
+  end
+
+  def to_map(changes, acc \\ %{})
+
+  def to_map(%{changes: changes}, acc) do
+    Enum.reduce(changes, acc, fn {key, value}, acc ->
+      key = Macro.underscore(Atom.to_string(key))
```
samuel-uniris, line 32: Why do you want to convert key as string ?

lib/archethic_web/controllers/api/transaction_controller.ex
```diff
@@ -119,4 +121,37 @@ defmodule ArchEthicWeb.API.TransactionController do
...
+  def origin_public_key_verify(conn, data) do
+    case OriginPublicKeyPayload.changeset(data) do
+      %{valid?: true} ->
+        cert =
+          data["PublicKey"]
+          |> Base.decode16()
+          |> elem(1)
+          |> Crypto.get_key_certificate()
+
+        if cert == data["Certificate"] do
```
samuel-uniris, line 134: I think you should use:
- `Crypto.get_root_ca_public_key` to get the root CA from the given public key,
- `Crypto.verify_key_certificate?` to check the given certificate for the given public key and retrieved root CA

samuel-uniris (changes requested):

lib/archethic_web/controllers/api/origin_public_key_payload.ex
imnik11: Done

lib/archethic_web/controllers/api/origin_public_key_payload.ex
imnik11: Thought to check the post request certificate and the correct certificate is matched so, would need string but didn't used.

Now Deleted as they is no usage

lib/archethic/shared_secrets/node_renewal.ex
```diff
@@ -114,6 +114,35 @@ defmodule ArchEthic.SharedSecrets.NodeRenewal do
     )
   end
 
+  @spec new_origin_shared_secrets_transaction(
+          origin_public_keys :: list(Crypto.key()),
+          secret_key :: binary()
+        ) :: Transaction.t()
+  def new_origin_shared_secrets_transaction(
+        origin_public_keys \\ [<<0::0, 0::0, :crypto.strong_rand_bytes(32)::binary>>],
```
samuel-uniris, line 122: Why do you want to define a default origin public keys  ?
 I think you can expect/enforce a list of public keys

lib/archethic/shared_secrets/node_renewal.ex
```diff
@@ -114,6 +114,35 @@ defmodule ArchEthic.SharedSecrets.NodeRenewal do
...
+        secret_key \\ :crypto.strong_rand_bytes(32)
+      )
+      when is_binary(secret_key) and
+             is_list(origin_public_keys) do
+    {new_shared_origin_pub_key, new_shared_origin_priv_key} = Crypto.generate_random_keypair()
+    origin_public_key = Enum.at(origin_public_keys, 0)
```
samuel-uniris, line 128: If you are taking a list of public keys, why are you online encrypting for the first ?

lib/archethic/shared_secrets/node_renewal.ex
```diff
@@ -114,6 +114,35 @@ defmodule ArchEthic.SharedSecrets.NodeRenewal do
...
+
+    encrypted_origin_shared_priv_key =
+      Crypto.ec_encrypt(new_shared_origin_priv_key, origin_public_key)
+
+    secret = encrypted_origin_shared_priv_key
```
samuel-uniris, line 133: The secret should be encrypted with the secret key and not with the origin public key.
The `Owernship.new/3` will encrypt the `secret_key` for all the origin public keys directly.

lib/archethic/shared_secrets/node_renewal.ex
```diff
@@ -114,6 +114,35 @@ defmodule ArchEthic.SharedSecrets.NodeRenewal do
...
+          origin_public_keys :: list(Crypto.key()),
+          secret_key :: binary()
+        ) :: Transaction.t()
+  def new_origin_shared_secrets_transaction(
+        origin_public_keys \\ [<<0::0, 0::0, :crypto.strong_rand_bytes(32)::binary>>],
+        secret_key \\ :crypto.strong_rand_bytes(32)
+      )
+      when is_binary(secret_key) and
+             is_list(origin_public_keys) do
+    {new_shared_origin_pub_key, new_shared_origin_priv_key} = Crypto.generate_random_keypair()
```
samuel-uniris, line 127: I think you can create multiple ones, for instance 10 keys

lib/archethic/shared_secrets/node_renewal.ex
```diff
@@ -114,6 +114,35 @@ defmodule ArchEthic.SharedSecrets.NodeRenewal do
...
+
+    encrypted_origin_shared_priv_key =
+      Crypto.ec_encrypt(new_shared_origin_priv_key, origin_public_key)
+
+    secret = encrypted_origin_shared_priv_key
+
+    Transaction.new(
+      :origin_shared_secrets,
+      %TransactionData{
+        content: new_shared_origin_pub_key,
```
samuel-uniris, line 138: I don't mind changing it, but we need to be consistent with `SharedSecrets.MemTablesLoader`

lib/archethic/shared_secrets/node_renewal_scheduler.ex
```diff
@@ -124,6 +124,9 @@ defmodule ArchEthic.SharedSecrets.NodeRenewalScheduler do
 
     ArchEthic.send_new_transaction(tx)
 
+    tx = NodeRenewal.new_origin_shared_secrets_transaction()
```
samuel-uniris, line 127: Why do you want to create new origin shared secrets after each renewal of node shared keys ?
The node and origin shared secrets should be distinct I think

samuel-uniris (changes requested):

lib/archethic/shared_secrets/node_renewal.ex
imnik11: As we are adding public key, content is want we expect to have i.e binary with pub_keys concatenated. Just once confirm this is exactly want we want then can add more pub_keys by concating

lib/archethic/shared_secrets/node_renewal.ex
imnik11: yes definitely we can extend this to n keys

lib/archethic/shared_secrets/node_renewal.ex
imnik11: ok, I will encrypt with `secret_key` instead of origin_public_key

lib/archethic/shared_secrets/node_renewal_scheduler.ex
imnik11: This is just to temporary trigger for tx

samuel-uniris: You can generate the transaction in the controller for now.

lib/archethic/shared_secrets/node_renewal.ex
samuel-uniris: From the YP perspective, it was mentioned a list of public keys but per chain of family.
Now the question will be to have a single chain or multiple chain, but otherwise I think you can concatenate generated origin public keys in the transaction's content

lib/archethic/shared_secrets/node_renewal_scheduler.ex
imnik11: OK

lib/archethic/shared_secrets/node_renewal.ex
```diff
@@ -119,23 +119,38 @@ defmodule ArchEthic.SharedSecrets.NodeRenewal do
...
+
+    new_origin_shared_pub_keys_list =
+      new_shared_origin_keys
+      |> Enum.map(fn {pub, _priv} -> pub end)
+
+    new_origin_shared_priv_keys_list =
+      new_shared_origin_keys
+      |> Enum.map(fn {_pub, priv} -> priv end)
+
+    new_origin_shared_pub_keys =
```
samuel-uniris, line 137: To simplify you can do: `:erlang.list_to_binary`

lib/archethic/shared_secrets/node_renewal.ex
```diff
@@ -119,23 +119,38 @@ defmodule ArchEthic.SharedSecrets.NodeRenewal do
...
+      new_origin_shared_pub_keys_list
+      |> Enum.reduce(fn x, acc -> acc <> x end)
+
+    new_origin_shared_priv_keys =
```
samuel-uniris, line 141: The same goes here

samuel-uniris (changes requested):

lib/archethic/shared_secrets/node_renewal.ex
samuel-uniris: But you would have to adapt the mem_table_loader then

lib/archethic_web/controllers/api/transaction_controller.ex
```diff
@@ -157,4 +161,16 @@ defmodule ArchEthicWeb.API.TransactionController do
         |> render("400.json", changeset: changeset)
     end
   end
+
+  def create_new_origin_shared_secret_tx(origin_public_key) do
+    aes_key =
```
samuel-uniris, line 166: I think this should be random

lib/archethic/shared_secrets/node_renewal.ex
```diff
@@ -150,7 +148,7 @@ defmodule ArchEthic.SharedSecrets.NodeRenewal do
     Transaction.new(
       :origin_shared_secrets,
       %TransactionData{
-        content: new_origin_shared_pub_keys,
+        content: <<origin_public_keys_bin <> new_origin_shared_pub_keys>>,
```
samuel-uniris, line 151: I don't think we need the origin public keys here, as they will be in the `ownerships` as authorized keys

samuel-uniris (changes requested):

PR #262, opened 2022-03-17
Add http port in node details

### LINKED ISSUE #243: Add HTTP port in the node details
### Problem to solve

Sometimes a node can be used different port for the explorer or endpoint for different reason like multiple nodes in the same host.
Even if clients can use the right port for the explorer or the API, internal services like metrics polling leverages public HTTP API. If the port configured is not the default one, the metrics polling will not work.

### Solution

We  should:
- [x] add an information in the node transaction's content to identify the HTTP port used
- [x] add this information in the mem tables and the relative structs
- [x] use this port in the metrics polling
- [x] Update doc tests with new node structure

blackode:
# Description

Before this change, there isn't the `http_port` information in the node  `content` section. So, added the `http_port` information in the node structure.

# Code Changes
## New Node struct
```elixir
   [
    :first_public_key,
    :last_public_key,
    :last_address,
    :reward_address,
    :ip,
    :port,
# --------new field-----------
     :http_port,
# ------------------------------
    :geo_patch,
    :network_patch,
    :enrollment_date,
    available?: false,
    average_availability: 1.0,
    availability_history: <<1::1>>,
    authorized?: false,
    authorization_date: nil,
    transport: :tcp
  ]
```

- Updated the `@discovery_table` related `MemTable`. Here, we used the index to update the fields in `ets` tables, as we introduced a new field the new indexes got updated to relevant fields.
- Fetching Metrics function now it is `[{ip_address, http_port()}]` before it is `[ip_address]`

Fixes #243 

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?
As this is related to node struct, I have updated the relevant test cases to use the new node struct as a result a lot of existing test cases were failed. So, update the relevant test units with the new node struct and see that all test cases are passed.

- [x] Node struct has new field `http_port`
- [x] Will have `http_port` part while serializing and de-serializing the node
- [x] Adding and connecting to node now has `http_port` information
- [x] In Metrics polling now we use the added `http_port`

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules


config/dev.exs
```diff
@@ -51,7 +51,7 @@ config :archethic, ArchEthic.P2P.BootstrappingSeeds,
   genesis_seeds:
     System.get_env(
       "ARCHETHIC_P2P_BOOTSTRAPPING_SEEDS",
-      "127.0.0.1:3002:00011D967D71B2E135C84206DDD108B5925A2CD99C8EBC5AB5D8FD2EC9400CE3C98A:tcp"
+      "127.0.0.1:3002:4000:00011D967D71B2E135C84206DDD108B5925A2CD99C8EBC5AB5D8FD2EC9400CE3C98A:tcp"
```
samuel-uniris, line 54: I don't think we need the HTTP port there, as this information is just for bootstrapping seed where only the P2P port is required

test/test_helper.exs
```diff
@@ -3,7 +3,7 @@ File.rm_rf!(ArchEthic.Utils.mut_dir())
 ExUnit.start(
   exclude: [:infrastructure, :CI, :CD, :oracle_provider],
   timeout: :infinity,
-  max_failures: 1
+  max_failures: 5
```
samuel-uniris, line 6: Why did you change the max failures to 5 ?

samuel-uniris (changes requested):

samuel-uniris:
Also it would be great to add,  in the metrics collector for polling,  the reference to the HTTP port.

https://github.com/archethic-foundation/archethic-node/blob/d3b338a98a79bcb3feaff927626e323b97ce4f9d/lib/archethic/metrics/collector.ex#L17

test/test_helper.exs
blackode: This for testing personal modification.... I should have changed back to 1, but forgot. Will do that

config/dev.exs
blackode: If we don't we will get some serialization and deserialization errors.
We used `String.split` for splitting based on the `:` so, we expect `http_port` over there

samuel-uniris: We may assume the default HTTP port from the configuration for the bootstrapping nodes.
Then when the node will join the network the information will be updated, when it will get other nodes information.
Don't you think ?

samuel-uniris:
> Also it would be great to add, in the metrics collector for polling, the reference to the HTTP port.
> 
> https://github.com/archethic-foundation/archethic-node/blob/d3b338a98a79bcb3feaff927626e323b97ce4f9d/lib/archethic/metrics/collector.ex#L17

Do you need more details on this one ?

lib/archethic/metrics/collector.ex
```diff
@@ -12,17 +12,17 @@ defmodule ArchEthic.Metrics.Collector do
...
   end
 
   @doc """
   Responsible for retrieving network metrics.
   """
-  @spec retrieve_network_metrics(list(:inet.ip_address())) :: map()
-  def retrieve_network_metrics(node_ip_addresses) do
-    Task.async_stream(node_ip_addresses, &service().fetch_metrics(&1))
+  @spec retrieve_network_metrics(list({:inet.ip_address(), port()})) :: map()
+  def retrieve_network_metrics(node_ip_addresses_with_http_port) do
```
samuel-uniris, line 24: What do you think of:  `endpoints` instead of the `node_ip_addresses_with_http_port` ?

lib/archethic/metrics/collector.ex
```diff
@@ -12,17 +12,17 @@ defmodule ArchEthic.Metrics.Collector do
   @doc """
   Get the list of Node IP addresses
   """
-  @spec retrieve_node_ip_addresses() :: list(:inet.ip_address())
-  def retrieve_node_ip_addresses do
-    Enum.map(P2P.authorized_nodes(), & &1.ip)
+  @spec retrieve_node_ip_addresses_and_http_ports() :: list({:inet.ip_address(), port()})
+  def retrieve_node_ip_addresses_and_http_ports do
```
samuel-uniris, line 16: What do you think of: `get_node_endpoints` ?

blackode: LGTM

config/dev.exs
blackode: but, when we split with `:` we miss the order right?

samuel-uniris: Why don't you let: `127.0.0.1:3002:00011D967D71B2E135C84206DDD108B5925A2CD99C8EBC5AB5D8FD2EC9400CE3C98A:tcp` and  in the Node struct, let the HTTP be the default one from the configuration ?

blackode: Yeah Updated ...

PR #264, opened 2022-03-24
Fix GeoPatch algorithm

### LINKED ISSUE #253: Fix GeoPatch algorithm
This is an issue in the current algorithm raised in the #238
There is a problem with the last digit of the geo patch as it's not match the geo coordinates.

Solution:
Rebuild the Geo-Patch ID Algo Entirely: (takes only 5-10 lines of code)

Geolocation patch for a node is represented by 3 Digits, for example “XYZ”  where 
First Digit (X) is the longitude patch id
Second Digit (Y)  is the latitude patch id
Third Digit (Z) is for better precision within the patch
First Digit and Second Digits give the patch ID, the third digit gives more precision

For example on the graph where a node is represented by C68, where 
C is the longitude patch id
6 is the latitude patch id
8 is the node’s precision within the C6 Patch
Digits can vary from [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, A, B, C, D, E, F]

- [x] Step 1: (longitude, latitude) 
Obtain the Geolocation represented in terms of longitude (-180,180), latitude (-90,90).
Eg 1: Node A = (82, 36) (82 is longitude and 36 is latitude)
Eg 2: Node B = (-82, 36) (-82 is longitude and 36 is latitude)
Eg 3: Node C = (82, -36) (82 is longitude and -36 is latitude)
Eg 4: Node D = (-82, -36) (-82 is longitude and -36 is latitude)

- [x] Step 2: (longitude_reference, latitude_reference)
(longitude_reference, latitude_reference) = (longitude / 22.5, latitude / 22.5)
Eg 1: (longitude_reference, latitude_reference) of Node A(82, 36) = (3.64, 1.6)
Eg 2: (longitude_reference, latitude_reference) of Node A(-82, 36) = (-3.64, 1.6)
Eg 3: (longitude_reference, latitude_reference) of Node A(82, -36) = (3.64, -1.6)
Eg 4: (longitude_reference, latitude_reference) of Node A(-82, -36) = (-3.64, -1.6)

- [x] Step 3: (First Digit, Second Digit)
First Digit & Second Digit =  [8,  9,  A,  B,  C, D,  E,  F,  0, 1, 2, 3, 4, 5, 6, 7, 8]
Graphical Value in Range = [-8,-7, -6, -5, -4, -3, -2, -1, 0, 1, 2, 3, 4, 5, 6, 7, 8]
Assign the first digit and second digit based on the graphical range

- [x] Step 4: (longitude_second_reference, latitude_second_reference) 
Calculate the Third Digit for better precision
(longitude_second_reference, latitude_second_reference) = decimal portion of (longitude_reference, latitude_reference)
If decimal portion of (longitude_reference, latitude_reference) is negative subtract 1 by that number
Eg 1: Node A = (82, 36) (82 is longitude and 36 is latitude)
(longitude_reference, latitude_reference) of Node A = (3.64, 1.6)
(longitude_second_reference, latitude_second_reference) of Node A = (0.64,0.6)
Eg 2: Node B = (-82, -36) (-82 is longitude and -36 is latitude)
(longitude_reference, latitude_reference) of Node B = (-3.64, -1.6)
(longitude_second_reference, latitude_second_reference) of Node A = (0.36,0.4)

Based on the decimal value range, add the third digit

Neylix:
# Description

Fixes bug on GeoPatch algorithm to match the following rules : 
- first digit represent latitude patch
- second digit represent longitude patch
- third digit is a precision within the previous two digits

Fixes #253 

## Type of change

# How Has This Been Tested?

- Runned a script with all coordinates possibilities (2 nested loop from -90 to 90 and -180 to 180), result is 2048 distinct geo patch (16x8x16)
- Geo patch match the grid coordinates shown in issue (pos 0,0 is patch 000, pos -1,-1 is patch FFF)

# Checklist:

### DIFF
--- a/lib/archethic/p2p/geo_patch.ex
+++ b/lib/archethic/p2p/geo_patch.ex
@@ -24,16 +24,21 @@ defmodule ArchEthic.P2P.GeoPatch do
   end
 
   defp compute_random_patch do
-    list_char = Enum.concat([?0..?9, ?A..?F])
-    Enum.take_random(list_char, 3) |> List.to_string()
+    list_char1 = Enum.concat([?0..?9, ?A..?F])
+    list_char2 = Enum.concat([?0..?3, ?C..?F])
+
+    Enum.take_random(list_char1, 2)
+    |> List.insert_at(1, Enum.take_random(list_char2, 1))
+    |> List.to_string()
   end
 
   defp compute_patch(lat, lon) do
-    lat_pos = (lat + 90) / 11.25
     lon_pos = (lon + 180) / 22.5
+    # Adding 4 to have second digit hex value from C to 3
+    lat_pos = (lat + 90) / 22.5 + 4
 
-    first_digit = main_index_patch(trunc(lat_pos))
-    second_digit = main_index_patch(trunc(lon_pos))
+    first_digit = main_index_patch(trunc(lon_pos))
+    second_digit = main_index_patch(trunc(lat_pos))
 
     lat_precision = ((lat_pos - trunc(lat_pos)) / 0.25) |> trunc()
     lon_precision = ((lon_pos - trunc(lon_pos)) / 0.25) |> trunc()
--- a/lib/archethic/p2p/geo_patch.ex
+++ b/lib/archethic/p2p/geo_patch.ex
@@ -33,7 +33,7 @@ defmodule ArchEthic.P2P.GeoPatch do
   end
 
   defp compute_patch(lat, lon) do
-    # convert 90 and 180 to -90 and -180 to not get an out of bound index for array
+    # convert 90 and 180 to -90 and -180 to not get an out of bound index
     lat = if(lat == 90, do: -90) || lat
     lon = if(lon == 180, do: -180) || lon
 
@@ -54,18 +54,18 @@ defmodule ArchEthic.P2P.GeoPatch do
   end
 
   defp main_index_patch(index) do
-    ['8', '9', 'A', 'B', 'C', 'D', 'E', 'F', '0', '1', '2', '3', '4', '5', '6', '7']
-    |> Enum.at(index)
+    {'8', '9', 'A', 'B', 'C', 'D', 'E', 'F', '0', '1', '2', '3', '4', '5', '6', '7'}
+    |> elem(index)
   end
 
   defp precision_index_patch(index1, index2) do
-    [
-      ['0', '1', '2', '3'],
-      ['4', '5', '6', '7'],
-      ['8', '9', 'A', 'B'],
-      ['C', 'D', 'E', 'F']
-    ]
-    |> Enum.at(index1)
-    |> Enum.at(index2)
+    {
+      {'0', '1', '2', '3'},
+      {'4', '5', '6', '7'},
+      {'8', '9', 'A', 'B'},
+      {'C', 'D', 'E', 'F'}
+    }
+    |> elem(index1)
+    |> elem(index2)
   end
 end
--- a/lib/archethic/p2p/geo_patch.ex
+++ b/lib/archethic/p2p/geo_patch.ex
@@ -33,6 +33,10 @@ defmodule ArchEthic.P2P.GeoPatch do
   end
 
   defp compute_patch(lat, lon) do
+    # convert 90 and 180 to -90 and -180 to not get an out of bound index for array
+    lat = if(lat == 90, do: -90) || lat
+    lon = if(lon == 180, do: -180) || lon
+
     lon_pos = (lon + 180) / 22.5
     # Adding 4 to have second digit hex value from C to 3
     lat_pos = (lat + 90) / 22.5 + 4
--- a/lib/archethic/p2p/geo_patch.ex
+++ b/lib/archethic/p2p/geo_patch.ex
@@ -29,68 +29,34 @@ defmodule ArchEthic.P2P.GeoPatch do
   end
 
   defp compute_patch(lat, lon) do
-    lat_sign = sign(lat)
-    lon_sign = sign(lon)
+    lat_pos = (lat + 90) / 11.25
+    lon_pos = (lon + 180) / 22.5
 
-    fdc = [lat / 90, lon / 180]
+    first_digit = main_index_patch(trunc(lat_pos))
+    second_digit = main_index_patch(trunc(lon_pos))
 
-    sd =
-      [(lat - lat_sign * 45) / 2, (lon - lon_sign * 90) / 2]
-      |> resolve_with_sign([lat, lon])
+    lat_precision = ((lat_pos - trunc(lat_pos)) / 0.25) |> trunc()
+    lon_precision = ((lon_pos - trunc(lon_pos)) / 0.25) |> trunc()
 
-    sdc = [List.first(sd) / 22.5, List.last(sd) / 45]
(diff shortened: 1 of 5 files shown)


Neylix:
Requested review form @samuel-uniris @I-Archer-Zero

imnik11 (commented):
Overall looks good and precision is improved

lib/archethic/p2p/geo_patch.ex
```diff
@@ -29,68 +29,34 @@ defmodule ArchEthic.P2P.GeoPatch do
   end
 
   defp compute_patch(lat, lon) do
-    lat_sign = sign(lat)
-    lon_sign = sign(lon)
+    lat_pos = (lat + 90) / 11.25
+    lon_pos = (lon + 180) / 22.5
```
imnik11, line 33: @I-Archer-Zero 
Looks like added 90,180 to avoid negative sign but look at 

`
lat_pos 
`

imnik11: It must be
`lat_pos = (lat + 90) / 22.5`

Neylix: Hi !
Dividing lat by 22.5 will have a result between 0 and 8. But we are looking for a range between 0 and 16 to get all hexadecimal possibilities (0 to F).
Si we have to divide by 11.25 to get it.

I think on issue's example there is a mismatch. On the graph, latitude only have 8 parts, but on explication bellow it says that first 2 digits are on range of 0 to F so 16 parts...

samuel-uniris: @I-Archer-Zero

I-Archer-Zero: Hi @Neylix 
Good observation 

lat should be divided by 22.5 and not 11.25 because we need square patches (not rectangle)
This also means that the lat/22.5 will have a result between 0 to 8

imnik11 (commented):
@Neylix @internet-zero

The Edge case problem is till in the new Algo but I avoided it by taking that also in the range 
so, earlier 
` defp index_patch([f_i, s_i]) when f_i > 0.5 and f_i <= 1 and s_i < -0.5 and s_i >= -1, do: '0'`

and when we input (-45, x) it crashed and it shall be avoided 
by using just `>=` while condition check

Now
`  defp precision_patch(lat, lon) when lat >= 0 and lat <= 0.25 and lon >= 0 and lon <= 0.25, do: '0'`

lib/archethic/p2p/geo_patch.ex
Neylix: Hi ! Thanks for your return.
I didn't know it had to be a square, I'll update it as you requested. Also need to add 4 to result to get index from C to 3 as in your graph.
I also noticed that in issue you say that first digit is for longitude index, but I did latitude first, should I update it too ?
I'm actually out of my PC until sunday, I'll update it at this time.

Neylix: Hello ! 
I fixed calculation in lastest commits

Neylix:
@imnik11 
Hello ! I did some tests with the values -45,x but it seems to work, I added these in test files and it going well.
Can you show the error you got ?

lib/archethic/p2p/geo_patch.ex
```diff
@@ -24,73 +24,44 @@ defmodule ArchEthic.P2P.GeoPatch do
...
-    lon_sign = sign(lon)
-
-    fdc = [lat / 90, lon / 180]
-
-    sd =
-      [(lat - lat_sign * 45) / 2, (lon - lon_sign * 90) / 2]
-      |> resolve_with_sign([lat, lon])
+    lon_pos = (lon + 180) / 22.5
+    # Adding 4 to have second digit hex value from C to 3
+    lat_pos = (lat + 90) / 22.5 + 4
```
imnik11, line 38: avoid adding 4 and adding 90, 180 was in earlier implementation so, avoid it

Neylix: Hello ! 
I don't really understand your suggestion. Here is why I choose this implementation instead of the one did before : 
The first calculations with adding 90 / 180 / 4 result in an index between 0 and 15 for ```lon_pos``` one and between 4 and 11 for ```lat_pos```.
With these index, we can easily find the good hexadecimal value in the array of the function ```main_index_patch```.

The advantage of this implementation is that it takes less ressources and less execution times than the previous implementation, i.e guetting a value from an array by his index is faster than doing multiple if condition to find the good hexadecimal value.
An other advantage is that it takes less lines of code.

To be easier to read, instead of adding 4 to ```lat_pos``` we can create a new array [C,D,E,F,0,1,2,3] and get hexadecimal value from it, but I think that using same array as for ```lon_pos``` is better to reduce memory usage.

@samuel-uniris What is your advice about it ?

samuel-uniris: So in term of performance pattern matching (multiple function clause) is like if/else block under the good bu is quite fast and optimized during compilation.

Accessing a random element in a list is not the same in Erlang. List are not arrays but linked list , which means to access nth element, need to scan nth before . When the list is big is not efficient. 
However if the list is bounded, know and static, you can leverage tuple where access a specific element is fast. 

There is different strategy for this problem. 

In term of readability the pattern matching is better when you have a several if/else/case blocks.

samuel-uniris: So in term of performance pattern matching (multiple function clause) is like if/else block under the good bu is quite fast and optimized during compilation.

Accessing a random element in a list is not the same in Erlang. List are not arrays but linked list , which means to access nth element, need to scan nth before . When the list is big is not efficient. 
However if the list is bounded, know and static, you can leverage tuple where access a specific element is fast. 

There is different strategy for this problem. 

In term of readability the pattern matching is better when you have a several if/else/case blocks.

Neylix: Hi ! Updated in last commit to use tuple. Using pattern matching function was not as easy to read because it needs to much function to catch all conditions.

samuel-uniris:
Seems more clear now.
Could you add more tests to be sure close locations, would be closed about the geo-patch, for instance in the 1st and 2nd digit but maybe not the 3rd

Neylix:
Added some tests with near location cities

test/archethic/p2p/geo_patch_test.exs
```diff
@@ -40,5 +56,9 @@ defmodule ArchEthic.P2P.GeoPatchTest do
     assert "A1A" == GeoPatch.from_ip({15, 62, 246, 57})
     assert "021" == GeoPatch.from_ip({109, 164, 214, 168})
     assert "8C0" == GeoPatch.from_ip({1, 2, 3, 4})
+    assert "F1F" == GeoPatch.from_ip({1, 1, 1, 1})
```
samuel-uniris, line 59: Is this distance good ? I mean it's almost close

Neylix: Yes, first digit F means that longitude is >= -22.5 and < 0, when going in positive we return to 0. i.e 0 is >= 0 and < 22.5.
So just after FFF we have 000.

samuel-uniris: Ok .
@imnik11 can you have a last check ?

PR #268, opened 2022-03-30
Add fn in SC lib to get genesis address in smart contract

imnik11:
# Description

Add get genesis address fn in library for smart contracts

Fixes # (187)

## Type of change

# How Has This Been Tested?
Need to be decided

# Checklist:

### DIFF
--- a/lib/archethic/contracts/interpreter/library.ex
+++ b/lib/archethic/contracts/interpreter/library.ex
@@ -161,12 +161,6 @@ defmodule ArchEthic.Contracts.Interpreter.Library do
   @spec get_genesis_address(binary()) ::
           binary()
   def get_genesis_address(address) do
-    address =
-      case Base.decode16(address, case: :mixed) do
-        {:ok, address} -> address
-        _ -> address
-      end
-
     nodes = Election.chain_storage_nodes(address, P2P.available_nodes())
     {:ok, address} = download_first_address(nodes, address)
     address


lib/archethic/contracts/interpreter.ex
```diff
@@ -587,6 +587,14 @@ defmodule ArchEthic.Contracts.Interpreter do
        when scope != :root,
        do: {node, acc}
 
+  # Whitelist the get_genesis_address/1 function in condition
+  defp prewalk(
```
samuel-uniris, line 591: I think it should also be available in the actions

lib/archethic/contracts/interpreter/library.ex
```diff
@@ -149,4 +153,27 @@ defmodule ArchEthic.Contracts.Interpreter.Library do
...
+    with [node | _rest] <- P2P.available_nodes(),
+         public_key_request <- %GetFirstPublicKey{address: address},
+         {:ok, %FirstPublicKey{public_key: key}} <- P2P.send_message(node, public_key_request) do
+      {pub, _priv} = Crypto.derive_keypair(key, 0)
+      Crypto.derive_address(pub)
+    else
+      [] ->
+        {:error, :network_issue}
+
+      {:ok, %NotFound{}} ->
```
samuel-uniris, line 175: As mentioned before, we have to change the implementation of the P2P message to not return NotFound, but the public key passed, if the previous public key does not exist.
Because this function should return a string, we have to design it with this in mind.
So I think you can return address, in case of not found, until the fix of the P2P message is done.

lib/archethic/contracts/interpreter/library.ex
```diff
@@ -149,4 +153,27 @@ defmodule ArchEthic.Contracts.Interpreter.Library do
...
+  """
+  @spec get_genesis_address(binary()) ::
+          binary() | {:error, :network_issue} | {:error, :enoaddress}
+  def get_genesis_address(address) do
+    address = Base.decode16!(address)
+
+    with [node | _rest] <- P2P.available_nodes(),
+         public_key_request <- %GetFirstPublicKey{address: address},
+         {:ok, %FirstPublicKey{public_key: key}} <- P2P.send_message(node, public_key_request) do
+      {pub, _priv} = Crypto.derive_keypair(key, 0)
```
samuel-uniris, line 169: I think you can directly do, `Crypto.derive_address` on the key retrieved

samuel-uniris (changes requested):

lib/archethic/contracts/interpreter/library.ex
```diff
@@ -149,4 +153,26 @@ defmodule ArchEthic.Contracts.Interpreter.Library do
...
   def size(map) when is_map(map), do: map_size(map)
+
+  @doc """
+  Get the genesis address of the chain
+
+  """
+  @spec get_genesis_address(binary()) ::
+          binary()
+  def get_genesis_address(address) do
+    address = Base.decode16!(address)
```
samuel-uniris, line 164: You should use `Base.decode16!(address, case: :mixed)` because you don't know if the hex passed only uppercase

imnik11: ok

samuel-uniris (approved):

Issue #276, opened 2022-04-08
Improve `GetFirstPublicKey` message


samuel-uniris:
`GetFirstPublicKey` implementation is not clear about the behavior expected and how it's used.
 Actually it should load the first public key, but finally its implementation:

https://github.com/archethic-foundation/archethic-node/blob/819b8af47e63411e5e7fdd9e68cb65921ab58463/lib/archethic/p2p/message.ex#L1229-L1237

We are returning the first transaction public key or not found, but I think you should leverage more: 
https://github.com/archethic-foundation/archethic-node/blob/819b8af47e63411e5e7fdd9e68cb65921ab58463/lib/archethic/transaction_chain.ex#L70-L74

So we would not have not found return, but the just public key passed in case there are not previous public key.

As we are messing with first transaction and first public key concept.
Moreover, we are fetching the transaction data and fields, which will cause more I/O than just fetching the first public key from the indexing.

---

Related to #268

internet-zero:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Product-and-Design-Workspace-61efe71b9083a90018250d89/issues/archethic-foundation/archethic-node/276?planning-poker) with ZenHub @blackode @imnik11 @roychowdhuryrohit-dev @samuel-uniris

internet-zero:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Product-and-Design-Workspace-61efe71b9083a90018250d89/issues/archethic-foundation/archethic-node/276?planning-poker) with ZenHub @apoorv-2204

prix-uniris:
@samuel-uniris So, should i use only the Indexer to fetch details as per new db impl. 
https://github.com/archethic-foundation/archethic-node/blob/b4e7d2cb8dd7f116f6135660113bc7157bf431fa/lib/archethic/db/embedded_impl/chain_index.ex#L486-L513
Using this function

samuel-uniris:
> @samuel-uniris So, should i use only the Indexer to fetch details as per new db impl. 
> https://github.com/archethic-foundation/archethic-node/blob/b4e7d2cb8dd7f116f6135660113bc7157bf431fa/lib/archethic/db/embedded_impl/chain_index.ex#L486-L513
> Using this function

I guess you can use `EmbeddedImpl.get_first_public_key​(​public_key​)​`

prix-uniris:
[...]

PR #280, opened 2022-04-18
Made use of get_first_public_key to always return a public key even i…

prix-uniris:
Made use of TransactionChain.get_first_public_key/1 to always return a public key even in case of nonexistent transactions and lower I/O operations which are not required if only want to get First_Public_Key for any given address.

# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- Test A
- Test B

# Checklist:

### DIFF
--- a/lib/archethic/p2p/message.ex
+++ b/lib/archethic/p2p/message.ex
@@ -307,8 +307,8 @@ defmodule ArchEthic.P2P.Message do
     <<19::8, length(node_public_keys)::16, :erlang.list_to_binary(node_public_keys)::binary>>
   end
 
-  def encode(%GetFirstPublicKey{address: address}) do
-    <<20::8, address::binary>>
+  def encode(%GetFirstPublicKey{public_key: public_key}) do
+    <<20::8, public_key::binary>>
   end
 
   def encode(%GetLastTransactionAddress{address: address, timestamp: timestamp}) do
@@ -706,10 +706,10 @@ defmodule ArchEthic.P2P.Message do
   end
 
   def decode(<<20::8, rest::bitstring>>) do
-    {address, rest} = Utils.deserialize_address(rest)
+    {public_key, rest} = Utils.deserialize_address(rest)
 
     {%GetFirstPublicKey{
-       address: address
+       public_key: public_key
      }, rest}
   end
 
@@ -1227,7 +1227,7 @@ defmodule ArchEthic.P2P.Message do
   end
 
   # Returns the first public_key for a given public_key and if the public_key is used for the first time, return the same public_key.
-  def process(%GetFirstPublicKey{address: public_key}) do
+  def process(%GetFirstPublicKey{public_key: public_key}) do
     case TransactionChain.get_first_public_key(public_key) do
       pub_key ->
         %FirstPublicKey{public_key: pub_key}
--- a/lib/archethic/p2p/message.ex
+++ b/lib/archethic/p2p/message.ex
@@ -1227,12 +1227,9 @@ defmodule ArchEthic.P2P.Message do
   end
 
   def process(%GetFirstPublicKey{address: address}) do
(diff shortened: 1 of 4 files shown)


prix-uniris:
This Fixes Issue #276

lib/archethic/p2p/message.ex
```diff
@@ -1227,12 +1227,9 @@ defmodule ArchEthic.P2P.Message do
   end
 
   def process(%GetFirstPublicKey{address: address}) do
```
samuel-uniris, line 1229: We should also rename `address` to `public_key`, don't you think ?

prix-uniris: Sure sam, that makes more sense.

prix-uniris: but we are supplying address to get a public key for a given address right

prix-uniris: so address also is a better argument name, also %GetFirstPublicKey{} accepts a address: , so variable name of address make more sense.

apoorv-2204: Hi, please keep it as public key, and in comments put address explaining it .

samuel-uniris: The name is misleading, the purpose of the function
So in GetFirstPublicKey, the public key should have more meaning.
Moreover, the implementation underneath is not using the address at all

apoorv-2204 (approved):
https://misc.archethic.net/s/gbdQ_8L_a

lib/archethic/p2p/message.ex
prix-uniris: Okay thanks, I think it's more clear now

prix-uniris: I have made the changes.

samuel-uniris:
I think you can rename the field in the struct get_first_public_key.ex

prix-uniris:
Thanks sam, changed that too and in relevant places.

lib/archethic/p2p/message.ex
```diff
@@ -706,10 +706,10 @@ defmodule ArchEthic.P2P.Message do
   end
 
   def decode(<<20::8, rest::bitstring>>) do
-    {address, rest} = Utils.deserialize_address(rest)
+    {public_key, rest} = Utils.deserialize_address(rest)
```
samuel-uniris, line 709: You should use `Utils.deserialize_public_key` here

lib/archethic/p2p/message.ex
```diff
@@ -1226,13 +1226,11 @@ defmodule ArchEthic.P2P.Message do
...
-  def process(%GetFirstPublicKey{address: address}) do
-    case TransactionChain.get_first_transaction(address, [:previous_public_key]) do
-      {:ok, %Transaction{previous_public_key: key}} ->
-        %FirstPublicKey{public_key: key}
-
-      {:error, :transaction_not_exists} ->
-        %NotFound{}
+  # Returns the first public_key for a given public_key and if the public_key is used for the first time, return the same public_key.
+  def process(%GetFirstPublicKey{public_key: public_key}) do
+    case TransactionChain.get_first_public_key(public_key) do
```
samuel-uniris, line 1231: I think the case statement is not required anymore, as you are expecting a single result

prix-uniris: so should we just directly return it as return value is always expected

samuel-uniris: I think

prix-uniris:
Made the above mentioned changes.

lib/archethic/p2p/message.ex
prix-uniris: Thanks, that makes more sense to directly return

samuel-uniris (approved):

Issue #158, opened 2021-11-16
Expose API to get the origin key


samuel-uniris:
An API should be exposed to get the origin key encrypted by the device's public key.
This API will be used by the client's application to get the list of origin private keys to use for the signature.

- [x] Implement a function to get encrypted origin private keys based on some authorized public key
- [x] Expose through an API
- [x] Add test files

I-Archer-Zero:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Product-and-Design-Workspace-61efe71b9083a90018250d89/issues/archethic-foundation/archethic-node/158?planning-poker) with ZenHub @apoorv-2204 @blackode @imnik11 @roychowdhuryrohit-dev @samuel-uniris

PR #288, opened 2022-04-26
Leverage sharded origin key

### LINKED ISSUE #275: Leverage sharded origin keys
Right now we are using an hardcoded origin shared key, but we should use this hardcoded key as an authorized public key key for the initial origin private keys. 

So the initial origin shared private keys will be encrypted for the hardcoded software public key. 

This requires:
- [x] create first origin shared secrets transaction during the network initialization
- [x] add the initial private keys encrypted for the hardcoded genesis software public key
- [x] add an origin shared chain by family
- [x] add test files

This will have an impact on libjs and wallet as #158 should be implemented to test it.

Neylix:
Fixes #275

# Description
First software origin key is no more hardcoded but randomly generated during network init and encrypted with the hardcoded public key in secret transaction.
Hardcoded key is now able to read and decrypt first software origin key

Fixes #275

## Type of change

# How Has This Been Tested?

- Implement file test to ensure that an origin key is generated during network init
- Used GraphiQL to get first ```origin_shared_secret``` and decypt it using libjs. Resulted private key is the same that the one generated during network init

# Checklist:

### DIFF
--- a/lib/archethic/bootstrap/network_init.ex
+++ b/lib/archethic/bootstrap/network_init.ex
@@ -6,6 +6,7 @@ defmodule ArchEthic.Bootstrap.NetworkInit do
   """
 
   alias ArchEthic.Bootstrap
+  alias ArchEthic.Bootstrap.NetworkInit
 
   alias ArchEthic.BeaconChain.ReplicationAttestation
 
@@ -31,12 +32,18 @@ defmodule ArchEthic.Bootstrap.NetworkInit do
   alias ArchEthic.TransactionChain.TransactionData.Ledger
   alias ArchEthic.TransactionChain.TransactionData.UCOLedger
   alias ArchEthic.TransactionChain.TransactionData.UCOLedger.Transfer
+  alias ArchEthic.TransactionChain.TransactionData.Ownership
 
   alias ArchEthic.TransactionChain.TransactionSummary
 
   require Logger
 
   @genesis_seed Application.compile_env(:archethic, [__MODULE__, :genesis_seed])
+
+  @genesis_origin_public_keys Application.compile_env!(
+                                :archethic,
+                                [NetworkInit, :genesis_origin_public_keys]
+                              )
 
   defp get_genesis_pools do
     Application.get_env(:archethic, __MODULE__) |> Keyword.get(:genesis_pools, [])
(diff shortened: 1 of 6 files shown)


Neylix:
As this PR change the way to interact with blockchain from client software, we have to provide them the way to get the origin private key before merging this PR.
We have to implement #158 and add this API in all librairies (i.e. libjs and libdart actually)

lib/archethic/bootstrap/network_init.ex
```diff
@@ -31,13 +32,19 @@ defmodule ArchEthic.Bootstrap.NetworkInit do
...
 
   alias ArchEthic.TransactionChain.TransactionSummary
 
   require Logger
 
   @genesis_seed Application.compile_env(:archethic, [__MODULE__, :genesis_seed])
 
+  @genesis_origin_public_keys Application.compile_env!(
+                                :archethic,
+                                [NetworkInit, :genesis_origin_public_keys]
```
samuel-uniris, line 45: I don't think the alias is necessary, you can use `__MODULE__` if you want to mention the current module.

Neylix: updated alias and added formatter for origin_shared_secrets content in explorer

lib/archethic_web/views/explorer_view.ex
```diff
@@ -196,5 +196,9 @@ defmodule ArchEthicWeb.ExplorerView do
     """
   end
 
+  def format_transaction_content(:origin_shared_secrets, content) do
+    Base.encode16(content)
```
samuel-uniris, line 200: I was thinking more like decoding the binary to list the public keys

Neylix: Did it !

samuel-uniris (approved):

lib/archethic/bootstrap/network_init.ex
```diff
@@ -99,6 +99,14 @@ defmodule ArchEthic.Bootstrap.NetworkInit do
     Transaction.new(
       :origin_shared_secrets,
       %TransactionData{
+        code: """
```
samuel-uniris, line 102: Good for the code contracts, just I don't think we need to constraints to the origin family here, this will mean even the node which are doing the transaction will have to use an origin private key generated from software.
So nodes might use hardware key as proof of enrollment to leverage the POW, so I think you can omit it.

lib/archethic/crypto.ex
```diff
@@ -1168,6 +1168,24 @@ defmodule ArchEthic.Crypto do
   @spec default_curve() :: supported_curve()
   def default_curve, do: Application.get_env(:archethic, __MODULE__)[:default_curve]
 
+  @doc """
+  Get the origin seed for a given origin family
+  """
+  @spec get_origin_family_seed(supported_origin()) :: binary()
+  def get_origin_family_seed(origin_id) do
+    storage_nonce() <>
+      case origin_id do
```
samuel-uniris, line 1177: I think you can leverage `ID.from_origin` which will return an integer for the given origin family

samuel-uniris: Exemple: 
```elixir
<<storage_nonce()::binary, ID.from_origin(origin_id)::8>>
```

samuel-uniris: Instead of putting this in the crypto.ex, you should move it to shared_secrets.ex, to avoid confusion with origin_id for public keys

PR #300, opened 2022-05-04
Rename ArchEthic into Archethic

samuel-uniris:
Applies the new renaming of Archethic from the brand guidelines


Neylix (changes requested):
Running `make clean` gives me an error
```** (Mix) The task "arch_ethic.clean_db" could not be found. Did you mean "archethic.clean_db"?```
You have to update `arch_ethic` to `archethic`
change have to be done in makefile and mix.exs (`mix dev.lbench`)

Neylix (approved):

PR #311, opened 2022-05-10
Hybrid root of trust

### LINKED ISSUE #241: Implement hybrid root of trust
To avoid any scalability issue and throughput improvements, we have to implement a new scheme for node cryptography where hardware keys are use for the node transaction chain but still using software keys to support fast signing and decryption.

So a node transaction will be build with:
```
previous_public_key: software key
previous_signature: software key
origin_signature: hardware key if possible
transaction's content: origin public key + certificate of the origin public key
```

Then we should:
- [x] Embed the  origin public key and its certificate inside the transaction's content
- [x] Sign the transaction (origin signature) with the origin private key
- [x] The proof of work algorithm should verify the origin signature with public key from the content
- [x] Add a new field in the node struct and mem table to inject the origin public key as attestation key
- [x] For any other network transaction, the origin signature should be check among the list of attestation key of the nodes
- [x] Complete use of software key for signing and decryption messages
- [x] Avoid as much as possible the bottleneck (maybe remove GenServer and keep ETS table) for signing and decryption
- [x] Test with TPM implementation
- [x] Stress test

samuel-uniris:
# Description

This PR aims to resolve the issue related from the usage of hardware keys for node chains and signature.
In this scheme we are introducing an hybrid approach where origin keys (aka hardware) certify the software public keys which will be used frequently by the node to sign.

Using this model, signature for transaction validation may be less subject to bottleneck as it's only based on software implementation and used from the memory.

The node transactions now are embedding their origin public key which should be certify with through the embedded root CA. 

So depending on the network configuration, only TPM/Yubikey or Software origin key can accept or reject node transaction chain.

Fixes #241 

## Type of change

# How Has This Been Tested?

- When a node is started, it should not have `Proof of Work` related issue
- On the explorer, on the node transaction chain, we can see the origin public key in the transaction's content
- The origin public keys is filled from the one on the node transaction chains
- After restart of node, the origin key should be backed and the node software keys as well by advancing on the chain

# Checklist:


config/dev.exs
```diff
@@ -54,20 +54,20 @@ config :archethic, Archethic.P2P.BootstrappingSeeds,
...
-       Archethic.Crypto.NodeKeystore,
+       Archethic.Crypto.NodeKeystore.Origin,
        (case System.get_env("ARCHETHIC_CRYPTO_NODE_KEYSTORE_IMPL", "SOFTWARE")
              |> String.upcase() do
           "SOFTWARE" ->
-            Archethic.Crypto.NodeKeystore.SoftwareImpl
+            Archethic.Crypto.NodeKeystore.Origin.SoftwareImpl
 
           "TPM" ->
             Archethic.Crypto.NodeKeystore.TPMImpl
```
Neylix, line 68: `Archethic.Crypto.NodeKeystore.TPMImpl` -> `Archethic.Crypto.NodeKeystore.Origin.TPMImpl` ?

PR #312, opened 2022-05-11
Correct CrossValidate message error

Neylix:
# Description

Correction of a bug occuring in CrossValidate message :
`MatchError) no match of right hand side value: <<192, 6::size(4)>>
    (archethic 0.14.0) lib/archethic/p2p/message.ex:653: Archethic.P2P.Message.decode/1`

Error is reproductible by this way : 
Start node1 and node2 -> everything is going well
Start node3 -> node3 goes in pending
When node3 goes in pending, node1 send a CrossValidate message to node2 and then here is the error on node2 side

Do not have Github issue for this error

## Type of change

# How Has This Been Tested?

Run 3 nodes as described before, and there is no more error

# Checklist:


samuel-uniris (approved):
Nice catch ! 👍

Issue #270, opened 2022-03-30
IPLookup is unable to detect and avoid forbidden IP ranges


roychowdhuryrohit-dev:
### What happened?

In many countries the broadband ISPs use internal routing among its users which is why the network configuration at user's end (ex: wifi routers) only contain the internal private ip address and not the absolute public address which is exposed outside the ISP's network. Since the module `ArchEthic.Networking.IPLookup` first uses NAT for deriving user's public address it often includes forbidden(private) IP ranges like 0.0.0.0.

<img width="354" alt="Screenshot 2022-03-30 at 20 10 35" src="https://user-images.githubusercontent.com/24897721/160867171-ee426a01-83d5-46c4-b6de-a59693bb54e7.png">

### What is expected ?

If the IP returned by `IPLookup.NAT` is invalid i.e belongs to the forbidden IP ranges then it will use IPFY to get the correct public IP.

IP ranges to avoid:

https://github.com/archethic-foundation/archethic-node/blob/819b8af47e63411e5e7fdd9e68cb65921ab58463/lib/archethic/networking.ex#L9

internet-zero:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Product-and-Design-Workspace-61efe71b9083a90018250d89/issues/archethic-foundation/archethic-node/270?planning-poker) with ZenHub @blackode @imnik11 @roychowdhuryrohit-dev @samuel-uniris

internet-zero:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Product-and-Design-Workspace-61efe71b9083a90018250d89/issues/archethic-foundation/archethic-node/270?planning-poker) with ZenHub @Neylix

PR #313, opened 2022-05-11
Resolves IP Lookups #270 

apoorv-2204:
#  Description

## We must check the IP retrieved for given node, (in production) belongs to Public IP. It  should not be static or NAT IP.
- We should validate IP  is_public? only in prod mode and not  in dev mode.
- If in dev mode , if we get Static/ NAT IP we should fall back to IPIFY to get Public IP
Fixes # (#270 )

## Type of change

- Bug fix (Fix static ip in archethic Snap)
- New feature (In dev single node || dev docker mode we should not fallback to public Ip)

# How Has This Been Tested?

- [x] Add unit Tests 
- [ ] Run node in mix prod , it must use nat and then fallback to Ipify
# Checklist:


roychowdhuryrohit-dev:
LGTM so far well done. Correct log message outputting as expected for private ip NAT.

```
2022-05-11 16:13:10.976 [info] Trying IPFY as fallback
2022-05-11 16:13:11.648 [info] Node IP discovered: HIDDEN
2022-05-11 16:13:11.649 [info] Not previous synchronization date
2022-05-11 16:13:11.649 [info] We are using the default one
2022-05-11 16:13:11.649 [info] Node bootstrapping...
```

samuel-uniris:
Function ending with `?` should return a boolean , it's the convention in Elixir codebases. 
So your function should check the validity of the public ip it's all. The fallback should be part of the rest of caller function

apoorv-2204:
> The point is not to rename it but simplify it a bit. You just have to check the validity of the ip otherwise call the existing fallback method with something like this:
> 
> ```elixir
> with {:ok, ip} <- apply(provider, :get_node_ip, []),
>         true <- Networking.valid_ip?(ip) do
>     Logger.info("Node IP discovered: #{:inet.ntoa(ip)} by #{provider}")
>     ip
> else
>   false ->
>     fallback(provider, "NAT: Private IP")
>   {:error, reason} ->
>     fallback(provider, reason)
> end
> ```

I agree , my bad

lib/archethic/networking/ip_lookup.ex
```diff
@@ -29,6 +29,20 @@ defmodule Archethic.Networking.IPLookup do
...
   end
 
+  defp is_public_ip(ip, provider) do
+    case Networking.valid_ip?(ip) do
+      true ->
+        Logger.info("Node IP discovered by #{provider}")
+        ip
+
+      false ->
+        case provider == Strict do
```
samuel-uniris, line 39: I would suggest to put it in the `else` of the `with` statement, it will be easier to read.

samuel-uniris:
Can you update the PR description from the template described ?

Also it would be great to add some unit tests

lib/archethic/networking/ip_lookup.ex
```diff
@@ -16,10 +18,13 @@ defmodule Archethic.Networking.IPLookup do
     provider = get_provider()
 
     ip =
-      case apply(provider, :get_node_ip, []) do
-        {:ok, ip} ->
-          Logger.info("Node IP discovered by #{provider}")
-          ip
+      with {:ok, ip} <- apply(provider, :get_node_ip, []),
+           true <- provider == Static || Networking.valid_ip?(ip) do
```
roychowdhuryrohit-dev, line 22: Even when provider is `Static` we shouldn't allow private IP at production (MIX_ENV=prod) since it will be blocked for mining in pending_transaction_validation.

https://github.com/archethic-foundation/archethic-node/blob/76cefaceccc1cf0291622be10e473cd12d148475/lib/archethic/mining/pending_transaction_validation.ex#L141

roychowdhuryrohit-dev (changes requested):

lib/archethic/networking/ip_lookup.ex
roychowdhuryrohit-dev: Please use something like this:

https://github.com/archethic-foundation/archethic-node/blob/76cefaceccc1cf0291622be10e473cd12d148475/config/prod.exs#L210-L219

samuel-uniris: Then you should integrate with :
```elixir
defp should_validate_node_ip? do
    :archethic
    |> Application.get_env(__MODULE__, [])
    |> Keyword.get(:validate_node_ip, false)
  end
```

So if the ip must be validated, then forbid private IP

roychowdhuryrohit-dev: If condition A = should_validate_node_ip? condition B = valid_ip?, then you should implement:

A` + AB

lib/archethic/networking/ip_lookup.ex
```diff
@@ -17,26 +17,29 @@ defmodule Archethic.Networking.IPLookup do
...
 
     ip =
-      with {:ok, ip} <- apply(provider, :get_node_ip, []),
-           true <- Networking.valid_ip?(ip) do
-        Logger.info("Node IP discovered by #{provider}")
-        ip
-      else
-        false ->
-          fallback(provider, "NAT : Private IP")
+      case {apply(provider, :get_node_ip, []), should_validate_node_ip?()} do
```
samuel-uniris, line 20: Sorry it's hard to read, one propose could be to do like this:
```elixir
def get_node_ip(provider) do
...
  with {:ok, ip} <- apply_provider(provider, :get_node_ip, []),
       :ok <- validate_ip(ip, should_validate_node_ip?()) do
    Logger.info("Node IP discovered by #{provider}")
    ip
  else
    {:error, reason} ->
      fallback(provider, reason)
  end
...
end


defp validate_ip(ip, true) do
   if Networking.valid_ip?(ip) do
      :ok 
   else
      {:error, :invalid_ip}
   end
end

defp validate_ip(ip, false), do: :ok
```

config/prod.exs
```diff
@@ -166,9 +177,6 @@ config :archethic, Archethic.Networking.PortForwarding,
          false
      end)
 
-config :archethic, Archethic.Networking.IPLookup.Static,
```
samuel-uniris, line 169: Same here 

> Why did you remove this one ?

samuel-uniris (changes requested):
There are still improvements to be done, and you should avoid removing stuff already present, as it can lead to regression

lib/archethic/networking/ip_lookup.ex
apoorv-2204: agreed

apoorv-2204:
> There are still improvements to be done, and you should avoid removing stuff already present, as it can lead to regression

Sure, I will do the required changes.
thx

config/dev.exs
```diff
@@ -87,10 +87,7 @@ config :archethic, Archethic.OracleChain.Scheduler,
   # Aggregate chain at the 50th second
   summary_interval: "0 * * * * *"
 
-config :archethic, Archethic.Networking.IPLookup, Archethic.Networking.IPLookup.Static
-
-config :archethic, Archethic.Networking.IPLookup.Static,
```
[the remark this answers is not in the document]
apoorv-2204: I though may be we required something like this, 
```
config :archethic, Archethic.Networking.IPLookup, provider:  Archethic.Networking.IPLookup.Static,

validate_node_ip:
    (case(System.get_env("ARCHETHIC_NODE_IP_VALIDATION", "true")) do
       "true" ->
         true
       _ ->
         false
     end)
```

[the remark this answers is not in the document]
samuel-uniris: It's not a bad idea, just we can change a bit the flow.
We can factorize this idea into a merged conf

```elixir
config :archethic, Archethic.Networking,
   validate_ip: (case(System.get_env("ARCHETHIC_NODE_IP_VALIDATION", "true")) do
       "true" ->
         true
       _ ->
         false
     end)
```

In `lib/archethic/networking.ex`
```elixir
def validate_ip(ip, ip_validation? \\ validate_ip?() )
def validate_ip,(ip, true) do
   if valid_ip?(ip) do
      :ok 
   else
      {:error, :invalid_ip}
   end
end

def valid_ip?(_ip, false) , do: :ok

defp valid_ip?(ip) do
   case :inet.ntoa(ip) do
      {:error, :einval} ->
        false

      ip_str ->
        !Regex.match?(
          @ip_validate_regex,
          to_string(ip_str)
        )
    end
end

defp validate_ip? do
    :archethic
    |> Application.get_env(__MODULE__, [])
    |> Keyword.get(:validate_ip, false)
end
```

and remove `config :archethic, Archethic.Mining.PendingTransactionValidation`
and change in `lib/archethic/mining/pending_transaction_validation.ex`
```elixir

  defp do_accept_transaction(%Transaction{
         type: :node,
         data: %TransactionData{
           content: content
         },
         previous_public_key: previous_public_key
       }) do
   ...

 {:conn, :ok} <- {:conn, validate_connection(ip, port, previous_public_key)} do
 ...
 else
  ...
  {:conn,  {:error, :invalid_ip}} ->
      {:error, "Invalid node's IP address"}
 {:conn,  {:error, :existing_node}} ->
      {:error, "Invalid node connection (IP/Port) for for the given public key - already existing"}
end

...

defp validate_connection(ip, port, previous_public_key) do
   with :ok <- Networking.validate_ip(ip),
           false <- P2P.duplicating_node?(ip, port, previous_public_key) do
        :ok
     else
        {:error, :invalid_ip} ->
          {:error, :invalid_ip}
        true ->
          {:error, :existing_node}
    end
end
```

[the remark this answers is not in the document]
samuel-uniris: What do you think ?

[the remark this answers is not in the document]
apoorv-2204: hello,Samuel, 
I think this is better. I  will put it in config.exs
About tests, What do think about 
mock  either
or create dummy modules 
for testing `Archethic.Networking.IPLookup.get_node_ip`.
What do you think ?

config/prod.exs
apoorv-2204: my bad

config/dev.exs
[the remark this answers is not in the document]
samuel-uniris: For external system and I/O mock is great

config/dev.exs
```diff
@@ -87,7 +87,9 @@ config :archethic, Archethic.OracleChain.Scheduler,
   # Aggregate chain at the 50th second
   summary_interval: "0 * * * * *"
 
-config :archethic, Archethic.Networking.IPLookup, Archethic.Networking.IPLookup.Static
+config :archethic, Archethic.Networking.IPLookup,
+  provider: Archethic.Networking.IPLookup.Static,
+  ipify_provider: Archethic.Networking.IPLookup.IPIFY
```
samuel-uniris, line 92: Why do you want to add this ipify provider conf ?

config/prod.exs
```diff
@@ -143,18 +143,34 @@ config :archethic, Archethic.Mining.PendingTransactionValidation,
...
+      "NAT" ->
+        Archethic.Networking.IPLookup.NAT
+
+      "STATIC" ->
+        Archethic.Networking.IPLookup.Static
+
+      "IPFY" ->
+        Archethic.Networking.IPLookup.IPIFY
+    end,
+  nat_provider: Archethic.Networking.IPLookup.NAT,
```
samuel-uniris, line 170: Same goes here

config/test.exs
```diff
@@ -88,8 +88,17 @@ config :archethic, Archethic.OracleChain.Scheduler,
 config :archethic, Archethic.OracleChain.Services.UCOPrice, provider: MockUCOPriceProvider
 
 config :archethic, Archethic.Networking.IPLookup, MockIPLookup
+config :archethic, Archethic.Networking.IPLookup.Static, MockStatic
+config :archethic, Archethic.Networking.IPLookup.NAT, MockNAT
+config :archethic, Archethic.Networking.IPLookup.IPIFY, MockIPIFY
+
+config :archethic, Archethic.Networking.IPLookup,
+  provider: MockStatic,
+  ipify_provider: MockIPIFY
```
samuel-uniris, line 97: as well

lib/archethic/mining/pending_transaction_validation.ex
```diff
@@ -137,8 +137,8 @@ defmodule Archethic.Mining.PendingTransactionValidation do
               key_certificate,
               root_ca_public_key
             )},
-         {:conn, true} <-
-           {:conn, valid_connection?(ip, port, previous_public_key, should_validate_node_ip?())} do
+         {:conn, :ok} <-
+           {:conn, valid_connection?(ip, port, previous_public_key)} do
```
samuel-uniris, line 141: if you return an ok tuple, better to no use `?` at the end of the function, but maybe rename it as`validate_connectioon`

config/dev.exs
samuel-uniris: If you want to leverage mock for testing, you might define IPIFY or NAT as a behavior, where a default implementation will be the NAT or IPIFY and the test impl the mock

samuel-uniris (changes requested):

lib/archethic/mining/pending_transaction_validation.ex
apoorv-2204: agreed

config/dev.exs
apoorv-2204: > If you want to leverage mock for testing, you might define Public IP gateway or Local Discovery as a behavior, where a default implementation of the public gateway could be IPIFY and during testing the Mock and the local discovery default impl being NAT

Hello.
Are you referring to  default_implmentation like this 

```
defmodule Archethic.Networking.IPLookup.PublicIPGatewayImpl do
  @moduledoc false

  defmacro __using__(_) do
    quote do
      def get_node_ip() do
        Logger.info("Trying IPFY as fallback")
        Archethic.Networking.IPLookup.IPIFY.get_node_ip()
        defoverridable get_node_ip: 0
      end
    end
  end

  @callback get_node_ip() :: {:ok, :inet.ip_address()} | {:error, any()}
end
```
either -- 
```
defmodule Archethic.Networking.IPLookup.PublicIPGatewayImpl do
  @moduledoc false
  @callback get_node_ip() :: {:ok, :inet.ip_address()} | {:error, any()}
end
```
```
defmodule Archethic.Networking.IPLookup.IPIFY do
  @moduledoc """
  Module provides external IP address of the node identified by IPIFY service.
  """

  alias Archethic.Networking.IPLookup.PublicIPGatewayImpl

  @behaviour PublicIPGatewayImpl

  @impl PublicIPGatewayImpl
  @spec get_node_ip() :: {:ok, :inet.ip_address()} | {:error, :not_recognizable_ip}
  def get_node_ip do
    with {:ok, {_, _, inet_addr}} <- :httpc.request('http://api.ipify.org'),
         {:ok, ip} <- :inet.parse_address(inet_addr) do
      {:ok, ip}
    else
      {:error, {:failed_connect, _reason}} ->
        {:error, :not_recognizable_ip}

      {:error, :einval} ->
        :inets.stop()
        {:error, :not_recognizable_ip}
    end
  end
end
```
either ---
Is it correct to have PublicIP Module, in which implements IPify to get public IP?

samuel-uniris: Nice you have tried with Macro 😉 

But you can do much more simplifier using Knigge

```elixir
defmodule Archethic.Networking.IPLookup.PublicGateway do
    use Knigge, otp_app: :archethic, default:  __MODULE__.IPIFYImpl

    @callback get_ip() :: {:ok, :inet.ip_address()} | {:error, :not_recognizable_ip}
end
```

This will create, at compile time, the function PublicGateway.get_ip() pointing to IPIFYImpl

But this can be override for test as:
```elixir
config :archethic, Archethic.Networking.IPLookup.PublicGateway, MockPublicGateway
```

apoorv-2204: Hello.
I missed you review, 
Will do the changes now.

lib/archethic/networking/ip_lookup.ex
```diff
@@ -33,7 +33,7 @@ defmodule Archethic.Networking.IPLookup do
     ip
   end
 
-  defp fallback(LocalDiscovery, reason) do
+  defp fallback_nat(reason) do
```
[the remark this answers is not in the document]
samuel-uniris: I think PublicGateway will use IPIFY, so the final fallback may not be required.
So if you're chain of fallback fails, we can return an error

lib/archethic/networking/ip_lookup.ex
```diff
@@ -1,25 +1,29 @@
...
-        {:ok, ip} ->
-          Logger.info("Node IP discovered by #{provider}")
-          ip
+      with {:ok, ip} <- apply(provider, :get_node_ip, []),
+           :ok <- Networking.validate_ip(ip) do
+        Logger.info("Node IP discovered by #{provider}")
+        ip
+      else
+        {:error, reason} when reason == :invalid_ip ->
+          fallback_nat(reason)
```
samuel-uniris, line 26: Why are you enforcing a fallback to NAT in that case ?
Each provider should have its way to provide fallback.
So:
- NAT fallback -> IPIFY
- IPIFY -> none
- Static -> none

lib/archethic/networking/ip_lookup/public_gateway.ex
```diff
@@ -0,0 +1,9 @@
+defmodule Archethic.Networking.IPLookup.PublicGateway do
+  @moduledoc false
+
+  use Knigge,
+    otp_app: :archethic,
+    default: Archethic.Networking.IPLookup.IPIFY
+
+  @callback get_node_ip() :: {:ok, :inet.ip_address()} | {:error, :not_recognizable_ip}
```
samuel-uniris, line 8: I think you can specify the `behavior` in Knigge, to avoid duplication of the callbacks.
Hence this behaviour will be chained to `IPLookup.Impl` behaviour , and Knigge will provide abstraction for it. (https://github.com/sascha-wolf/knigge#the-behaviour-key-is-optional)

samuel-uniris (changes requested):
Overall it's good 👍  but we can enhance it I think

lib/archethic/networking/ip_lookup.ex
apoorv-2204: when in dev mode , the provider = get_provider() is Static
```   
with {:ok, ip} <- apply(provider, :get_node_ip, []),
           :ok <- Networking.validate_ip(ip) do
                # With :ok from networking.validate_ip()
                # static is local host which will pass here
                # NAt in docker mode will also pass here
               # IPIFY public not being used in dev mode
        Logger.info("Node IP discovered by #{provider}")
        ip
      else
        {:error, reason} when reason == :invalid_ip ->
          fallback_nat(reason)
 {:error, reason} ->
          fallback(provider, reason)
```

when in prod_mode , WE can have any of the three STATIC,NAT, IPIFIY (in a case)
as per prod.exs
```   
with {:ok, ip} <- apply(provider, :get_node_ip, []),
           :ok <- Networking.validate_ip(ip) do
                # With :ok from networking.validate_ip()
                 # if no errors then IPIFY will pass
                 # nat and static wont pass
        Logger.info("Node IP discovered by #{provider}")
        ip
      else

        {:error, reason} when reason == :invalid_ip ->
            # when in prod Static and nat provider will give invalid ip
            # for both case in prod we, fallback to fallback_nat to go through
           # IPIFY , .
          fallback_nat(reason)
 {:error, reason} ->
    #If error in ipify it will go here
          fallback(provider, reason)
```

apoorv-2204: In this scenario, we wont be able to mock, As we could either have 
```
  {:error, reason} when reason == :invalid_ip ->
          fallback_nat(reason)
```
or provider in config
or use 
```
 defp fallback(LocalDiscovery, reason) do
```
In such case we will need to mock NAT , which will have two behaviour

apoorv-2204: - Which will give error with two behavior's clash.

- Since we cant have params in configs, dev, prod due Regression
We can drop unit tests  for IPlookup.exs

dev-guidelines.md
```diff
@@ -0,0 +1,25 @@
+#    Development Guidelines
```
apoorv-2204, line 1: Hello, Is it a good idea have this file? 
or Should I remove it.

lib/archethic/networking/ip_lookup.ex
apoorv-2204: marking it as resolved ,
as we have new implementation

lib/archethic/networking/ip_lookup/public_gateway.ex
apoorv-2204: marking it as resolved ,
as we have new implementation

dev-guidelines.md
apoorv-2204: @samuel-uniris ,
do you think its good?
Or should I drop it.

config/prod.exs
```diff
@@ -158,13 +158,13 @@ config :archethic, Archethic.Networking,
 config :archethic,
        Archethic.Networking.IPLookup,
        (case(System.get_env("ARCHETHIC_NETWORKING_IMPL", "NAT") |> String.upcase()) do
-          local_d when local_d in ["NAT", "LocalDiscovery"] ->
-            Archethic.Networking.IPLookup.LocalDiscovery
+          nat when nat in ["NAT"] ->
```
samuel-uniris, line 161: You can simplify by not using a list but directly the string

lib/archethic/networking/ip_lookup/local_discovery.ex
```diff
@@ -1,9 +1,9 @@
-defmodule Archethic.Networking.IPLookup.LocalDiscovery do
+defmodule Archethic.Networking.IPLookup.NATDiscovery do
```
samuel-uniris, line 1: I think you didn't rename the file as well

config/dev.exs
```diff
@@ -87,13 +87,26 @@ config :archethic, Archethic.OracleChain.Scheduler,
...
+# -----Start-of-Networking-dev-configs-----
+config :archethic, Archethic.Networking,
+  validate_node_ip:
+    (case(System.get_env("ARCHETHIC_NODE_IP_VALIDATION", "false")) do
+       "true" ->
+         true
+
+       _ ->
+         false
+     end)
```
blackode, line 99:
```suggestion
  validate_node_ip: System.get_env("ARCHETHIC_NODE_IP_VALIDATION", false) == "true"
```

lib/archethic/networking.ex
```diff
@@ -7,6 +7,21 @@ defmodule Archethic.Networking do
...
+  def validate_ip(ip, ip_validation? \\ should_validate_node_ip?())
+
+  def validate_ip(ip, true) do
+    case valid_ip?(ip) do
+      true ->
+        :ok
+
+      _ ->
+        {:error, :invalid_ip}
+    end
```
blackode, line 20:
```suggestion
    if !valid_ip?(ip), do: {:error, :invalid_ip}
```
Any how if will send you the `:ok` once it is fall for else case

blackode (changes requested):

lib/archethic/networking.ex
samuel-uniris: If you prefer multiline, we could do like this :
```elixir
if valid_ip?(ip) do
   :ok
else
   {:error, :invalid_ip}
end

**Comment by apoorv-2204** [comment]
image

File: lib/archethic/networking.ex (hunk 17 above)
Comment (apoorv-2204) [reply lib/archethic/networking.ex]: thanks

File: config/dev.exs (hunk 16 above)
Comment (apoorv-2204) [reply config/dev.exs]: thanks

File: config/prod.exs [hunk 18]
```diff
@@ -143,17 +143,22 @@ config :archethic, Archethic.Mining.PendingTransactionValidation,
...
           "NAT" ->
-            Archethic.Networking.IPLookup.NAT
+            Archethic.Networking.IPLookup.NATDiscovery
 
           "STATIC" ->
             Archethic.Networking.IPLookup.Static
 
-          "IPFY" ->
-            Archethic.Networking.IPLookup.IPIFY
+          "IPIFY" ->
```
samuel-uniris, line 160: I think you can change it to RemoteDiscovery, we will update the snap conf as well

lib/archethic/networking/ip_lookup/nat_discovery.ex
```diff
@@ -0,0 +1,55 @@
...
+
+  defp fallback(PMP, reason) do
+    {:error, reason}
+  end
+
+  defp fallback(_provider, reason) do
+    {:error, reason}
+  end
+
+  defp module_args() do
```
samuel-uniris, line 50: I think you should rename it `provider` because it's hard to determine the meaning of `module_args`

samuel-uniris:
> As per req: `node in :prod crashes when given ARCHETHIC_NETWORKING_IMPL=STATIC` image

For static, we should set env var ARCHETHIC_STATIC_IP

lib/archethic/networking/ip_lookup/nat_discovery.ex
apoorv-2204: done

config/prod.exs
apoorv-2204: done

apoorv-2204:
> > As per req: `node in :prod crashes when given ARCHETHIC_NETWORKING_IMPL=STATIC` image
> 
> For static, we should set env var ARCHETHIC_STATIC_IP

There is no fallback for static, when ran in  with NAT, gets public ip.
And it is accessible via 127.0.0.1
image
image

config/config.exs
```diff
@@ -124,6 +125,16 @@ config :archethic, ArchethicWeb.FaucetController,
     "3A7B579DBFB7CEBE26293850058F180A65D6A3D2F6964543F5EDE07BEB2EFDA4"
     |> Base.decode16!(case: :mixed)
 
+# -----Start-of-Networking-configs-----
```
samuel-uniris, line 128: i think you can omit this as it is mentioned in the default `provider` function

config/prod.exs
```diff
@@ -143,17 +143,22 @@ config :archethic, Archethic.Mining.PendingTransactionValidation,
...
           "NAT" ->
-            Archethic.Networking.IPLookup.NAT
+            Archethic.Networking.IPLookup.NATDiscovery
 
           "STATIC" ->
             Archethic.Networking.IPLookup.Static
 
-          "IPFY" ->
-            Archethic.Networking.IPLookup.IPIFY
+          "RemoteDiscovery" ->
```
samuel-uniris, line 160: "REMOTE" would better to stay consistent with the rest

lib/archethic/networking.ex
```diff
@@ -7,6 +7,19 @@ defmodule Archethic.Networking do
   alias __MODULE__.PortForwarding
 
   @ip_validate_regex ~r/(^0\.)|(^127\.)|(^10\.)|(^172\.1[6-9]\.)|(^172\.2[0-9]\.)|(^172\.3[0-1]\.)|(^192\.168\.)/
+
+  def validate_ip(ip, ip_validation? \\ should_validate_node_ip?())
```
samuel-uniris, line 11: could you add function doc & typespec ;)

samuel-uniris:
otherwise it seems good 👍

Issue #296, opened 2022-05-01
GraphQL request : TransactionChain pagination


redDwarf03:
### What happened?

I created 14 transactions in a transaction chain

when I use a request with ```page: 0```, I obtain the 4 last transactions
```query { transactionChain(address: "00003C85EB8CC60B73A049FE80513BB5DD2204B6A506DD8D83ACC34FEE40B21CAA94", page: 0) {  address} }

{
  "data": {
    "transactionChain": [
      {
        "address": "00000C37AA448DB2913EC62E232C27BFB096D09542457778FB2AD04B0B56F8C18092"
      },
      {
        "address": "00000D597B72BD3711588A48504698F389E40B144F16BD94B8449E160AB51A7B731C"
      },
      {
        "address": "00008CE5F15A7564AB73F1589BE1A74A6A5F993DC377CFDBDD9FFADD40E5D1EAAAC4"
      },
      {
        "address": "00003C85EB8CC60B73A049FE80513BB5DD2204B6A506DD8D83ACC34FEE40B21CAA94"
      }
    ]
  }
}
```

when I use a request with ```page: 1```, I obtain the 10 first transactions I created
```query { transactionChain(address: "00003C85EB8CC60B73A049FE80513BB5DD2204B6A506DD8D83ACC34FEE40B21CAA94", page: 0) {  address} }

{
  "data": {
    "transactionChain": [
      {
        "address": "00007C7FAC9A741BF522EAF69824945B0721155C307270AE843A55F42C63F4526EC9"
      },
      {
        "address": "00000A9CE7F2DE51CFA0D3130099B2C6787AEB866647FD84945A124B21B14BB09D0E"
      },
      {
        "address": "00009D1AEA08BD7AF11DBEA0CA8CEAC2DA7C28B6CCEE4C69576827DB1C497655E080"
      },
      {
        "address": "000019565DF24A9AA5F82612E93BDB6BE18BCBA1158E74DBA810543C4948B523B79B"
      },
      {
        "address": "0000E15BF01140218B6A6269A514B8DCB5EA395A1C3351976EBE4FA2FCBFFC061917"
      },
      {
        "address": "0000FBB80E7223372C82C2CADF7C5A979AF1F266D1F404EBF5D3B231BFB0A3CD8CB0"
      },
      {
        "address": "0000E552B131BEEEBFE89903C9A97BD72424BCE755DD1BDFCDD9F31F127C6520B050"
      },
      {
[...]

PR #314, opened 2022-05-12
297 Pagination based on paging_state as last page last transaction address GraphQl API

### LINKED ISSUE #297: Use pagination state in GraphQL API
### Problem to solve

GraphQL API `transactionChains` uses pagination mechanism using number.
However it may create confusion which number to use 0 or 1 for the first page causing different results due to implementation `Enum.at/2`.
Also we would like to avoid memory consumption by loading the entire chain.

Related to #296

### Solution

We can leverage the `pagination_state` returned from the DB and provided by the `GetTransactionChain` message.

The parameter will be then be the last address retrieved as the starting point on the chain to be retrieved.

However, for `networkTransaction` API we can still use the page as a number greater than 0, as the state will be different for several chains.

### Development Tasks

- [x] Breaking existing code in to definitions for re-usability for newer approach
- [x] Creating a New Pagination approach based on `paging_address` instead page number where `paging_address` is page's last transaction address
- [x] Add new function resolver to fetch transaction chain based on `paging_address` instead of loading all transactions and slicing down to pages
- [x] Add `networkTransactions` API `page` attribute  validation 
- [x] Add or Update Relevant Unit Test cases for Transaction Chain

blackode:
# Description

Before this change, to get the specific page level transaction_chain, we are loading whole transactions and slicing them by the page size and retrieving the relevant page transactions. 

In this change, the `page` attribute which was `number` now converted to `pagingAddress` of type `string` is now the address of the last transaction in the page which act as a `paginate_state` for next page.

Here, we are loading specific `paging_state` based page size transactions only. 

Fixes #297 #296 

## Type of change

# How Has This Been Tested?
## Steps
1. Select the address whose chain length is greater 10 for an example consider chain_length of 15
2. Now run query `transactionChain(address: "address_selected"){address}`, which gives 10 transactions
3. Make sure it is listed only 10 transactions
4. Pick the last transaction address from above 10 transactions
5. Run the query `transactionChain(address: "address_selected", pagingAddress: "last_transaction_address"){address}`
6. Now above query should give only 5 transactions
7. Now take out the last transaction address from above 5 transactions and repeat the step 5
8. It should give empty list of transactions.

- Test with out paging state based transactions
- Test paging_state based transactions

# Checklist:

### DIFF
--- a/lib/archethic.ex
+++ b/lib/archethic.ex
@@ -240,11 +240,34 @@ defmodule Archethic do
   """
   @spec get_transaction_chain(binary()) :: {:ok, list(Transaction.t())} | {:error, :network_issue}
   def get_transaction_chain(address) when is_binary(address) do
+    local_available_nodes = locally_available_nodes(address)
+    get_transaction_chain(local_available_nodes, address)
+  end
+
+  def get_transaction_chain_by_paging_address(address, paging_address) when is_binary(address) do
+    options = [paging_state: paging_address]
+    local_available_nodes = locally_available_nodes(address)
+    transaction_chain_by_paging_address(local_available_nodes, address, options)
+  end
+
+  defp transaction_chain_by_paging_address([node | rest], address, options) do
+    case P2P.send_message(node, %GetTransactionChain{
+           address: address,
+           paging_state: Keyword.get(options, :paging_state)
+         }) do
+      {:ok, %TransactionList{transactions: transactions}} ->
+        {:ok, transactions}
+
+      {:error, _} ->
+        transaction_chain_by_paging_address(rest, address, options)
+    end
+  end
+
+  defp locally_available_nodes(address) do
     address
     |> Election.chain_storage_nodes(P2P.available_nodes())
     |> P2P.nearest_nodes()
     |> Enum.filter(&Node.locally_available?/1)
-    |> get_transaction_chain(address)
   end
 
   defp get_transaction_chain(nodes, address, opts \\ [], acc \\ [])
--- a/lib/archethic.ex
+++ b/lib/archethic.ex
@@ -244,32 +244,6 @@ defmodule Archethic do
     get_transaction_chain(local_available_nodes, address)
   end
 
-  def get_transaction_chain_by_paging_address(address, paging_address) when is_binary(address) do
-    options = [paging_state: paging_address]
-    local_available_nodes = locally_available_nodes(address)
-    transaction_chain_by_paging_address(local_available_nodes, address, options)
-  end
-
-  defp transaction_chain_by_paging_address([node | rest], address, options) do
-    case P2P.send_message(node, %GetTransactionChain{
-           address: address,
-           paging_state: Keyword.get(options, :paging_state)
-         }) do
-      {:ok, %TransactionList{transactions: transactions}} ->
-        {:ok, transactions}
-
-      {:error, _} ->
-        transaction_chain_by_paging_address(rest, address, options)
-    end
-  end
-
-  defp locally_available_nodes(address) do
-    address
-    |> Election.chain_storage_nodes(P2P.available_nodes())
-    |> P2P.nearest_nodes()
-    |> Enum.filter(&Node.locally_available?/1)
-  end
-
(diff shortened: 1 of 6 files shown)


lib/archethic_web/graphql_schema.ex
```diff
@@ -58,11 +58,19 @@ defmodule ArchethicWeb.GraphQLSchema do
     """
     field :transaction_chain, list_of(:transaction) do
       arg(:address, non_null(:address))
-      arg(:page, :integer)
+      arg(:paging_address, :string)
```
samuel-uniris, line 61: We need to valid as well if it's a valid address, so the type should be `address`

lib/archethic_web/graphql_schema.ex
```diff
@@ -58,11 +58,19 @@ defmodule ArchethicWeb.GraphQLSchema do
...
 
       resolve(fn args = %{address: address}, _ ->
-        page = Map.get(args, :page, 1)
-        Resolver.paginate_chain(address, page)
+        paging_address = Map.get(args, :paging_address)
+
+        paging_address =
+          if paging_address do
+            Base.decode16!(paging_address)
```
samuel-uniris, line 68: By validating the paging_address by the type `address` this cast will be not necessary

lib/archethic.ex
```diff
@@ -240,11 +240,34 @@ defmodule Archethic do
   """
   @spec get_transaction_chain(binary()) :: {:ok, list(Transaction.t())} | {:error, :network_issue}
   def get_transaction_chain(address) when is_binary(address) do
+    local_available_nodes = locally_available_nodes(address)
+    get_transaction_chain(local_available_nodes, address)
+  end
+
+  def get_transaction_chain_by_paging_address(address, paging_address) when is_binary(address) do
```
samuel-uniris, line 247: Could you add typespec and doc as well ?

lib/archethic_web/graphql_schema/resolver.ex
```diff
@@ -112,7 +112,7 @@ defmodule ArchethicWeb.GraphQLSchema.Resolver do
     )
   end
 
-  def network_transactions(type, page) do
+  def network_transactions(type, page) when page >= 1 do
```
samuel-uniris, line 115: I think you should also return an error when the page is < 1

samuel-uniris (changes requested):
Thanks for the PR.
Overall is good, just some changes to perform

lib/archethic.ex
blackode: Updated the function with doc and typespe

lib/archethic_web/graphql_schema/resolver.ex
blackode: Introduced new `page_type`

lib/archethic.ex
```diff
@@ -295,6 +269,38 @@ defmodule Archethic do
...
+           paging_state: Keyword.get(options, :paging_state)
+         }) do
+      {:ok, %TransactionList{transactions: transactions}} ->
+        {:ok, transactions}
+
+      {:error, _} ->
+        transaction_chain_by_paging_address(rest, address, options)
+    end
+  end
+
```
samuel-uniris, line 303: I think you are missing the case when there is not nodes available.
This is causing the dialyzer issue I think

blackode: Yeah, I missed the case when nodes list is empty `[]` 
Thanks :)

samuel-uniris:
I think it could be great also to apply the `page` type in `transactions` queries (locally stored)

lib/archethic_web/graphql_schema.ex
```diff
@@ -105,7 +107,7 @@ defmodule ArchethicWeb.GraphQLSchema do
     """
     field :network_transactions, list_of(:transaction) do
       arg(:type, :transaction_type)
```
samuel-uniris, line 109: I could be great also to add field requirement on the type

samuel-uniris (approved):
It's working thanks 👍

samuel-uniris (approved):

PR #317, opened 2022-05-12
Fix authorized nodes unavailability

### LINKED ISSUE #318: Temporary unavailability of miners cause proof of election failure
### What happened?

An error is raised when a previous authorized node become unavailable and the node shared secrets's renewal is performed, excluding this node from the list.

Then, when the node comes back, it's considered as authorized nodes but without the knowledge of previous daily nonce, causing an error of proof of election verification.

### What is expected ?

An authorized node should be able to rejoin the network without causing issue in the transaction validation

samuel-uniris:
# Description

This PR aims to resolve issue regarding nodes which become unavailable and try to rejoin later while the node shared secrets have been performed since.

But once the node is coming back it should be able to load missed transactions without issue regarding election or shared secrets

Fixes #318 

## Type of change

# How Has This Been Tested?

- Start few nodes and wait they become authorized
- Stop one
- Wait this node becomes unavailable
- Wait other nodes perform shared secrets transactions
- Start the paused node
- It should be able to decode the missed transactions and start validate transaction again

# Checklist:


lib/archethic/shared_secrets/node_renewal.ex
```diff
@@ -50,10 +50,19 @@ defmodule Archethic.SharedSecrets.NodeRenewal do
   @spec next_authorized_node_public_keys() :: list(Crypto.key())
   def next_authorized_node_public_keys do
     DB.get_latest_tps()
-    |> Election.next_authorized_nodes(P2P.available_nodes())
+    |> Election.next_authorized_nodes(candidates(), P2P.authorized_nodes())
     |> Enum.map(& &1.last_public_key)
```
Neylix, line 54: It should be `first_public_key` because if node disconnect and reconnect, it will have a new last public key but during disconnection time, other nodes don't know the new public key of the node.
And so when the node reconnect it will not be able to descrypt the secret as it's last public key is not the same than the one added by others nodes in the transaction.
By updating this, node shared secrets transaction have to be decrypted with node first public key

Neylix (changes requested):

Neylix (approved):

apoorv-2204:
@Neylix   I think We can test this on  3 node network.

apoorv-2204:
Soo many changes and Soo many commits.

lib/archethic_web/templates/node/list.html.leex
```diff
@@ -4,30 +4,56 @@
...
+      </div>
+      <%= for node <- @authorized_nodes do %>
+        <div class="box">
+          <div class="columns is-desktop">
+            <div class="column">
+                <div>
+                    <p class="heading">First public key</p>
+                    <%= link to: Routes.live_path(@socket, ArchethicWeb.NodeDetailsLive, Base.encode16(node.last_public_key)) do%>
+                      <span class="is-hidden-mobile"><%= Base.encode16(:binary.part(node.first_public_key, 0, 30)) %>...</span>
+                      <span class="is-hidden-touch"><%= Base.encode16(:binary.part(node.first_public_key, 0, 13)) %>...</span>
```
Neylix, line 24: To be more responsive you should do something like this to avoid double line :
```html
<span class="is-hidden-tablet"><%= Base.encode16(:binary.part(node.first_public_key, 0, 10)) %>...</span>
```

Neylix (changes requested):

Neylix (approved):

lib/archethic_web/templates/node/list.html.leex
samuel-uniris: done

apoorv-2204:
Wait its done???  so fast?

Issue #341, opened 2022-05-30
Create Mix task to test the Nat discovery and port forwarding


samuel-uniris:
### Problem to solve

Some ISP provide some issues regarding NAT discovery and Port forwarding using UPnP, and we have to identify which ISP is not supporting this.

### Solution

I propose to create a mix task to leverage the Networking module trying to get the public ip and trying to open a port.

Then we can compare it with miniupnpc to check the working state of the both tools.
If not we would have a way to identify which protocol is missing in our implementation.

PR #342, opened 2022-05-30
Mix task nat discovery port forwarding#341

apoorv-2204:
# Description

Include Mix task Support for Nettools

Fixes # (#341  )

## Type of change

# How Has This Been Tested?

- manually by cli

# Checklist:


samuel-uniris:
Nice !
I think you can reduce to two flags: `ip` and `punch` and let the fallbacks do the rest . In case of error you will know which provider failed

apoorv-2204:
> Nice ! I think you can reduce to two flags: `ip` and `punch` and let the fallbacks do the rest . In case of error you will know which provider failed

Fair enough, agreed.

apoorv-2204:
> Nice ! I think you can reduce to two flags: `ip` and `punch` and let the fallbacks do the rest . In case of error you will know which provider failed
I double checking, as mentioned in Issue public ip,
we Rely  on Remote Discovery
So : this 
`Networking.IPLookup.RemoteDiscovery.get_node_ip()`
we Rely  on System_env
or `Networking.IPLookup.get_node_ip()`

lib/mix/tasks/nettools.ex
```diff
@@ -0,0 +1,66 @@
...
+    args_to_internal_representation({[punch: true], more_ports, errors})
+  end
+
+  def args_to_internal_representation({[punch: true], _port_list = [], _errors}) do
+    Networking.PortForwarding.try_open_port(40_000, true)
+    Networking.PortForwarding.try_open_port(30_002, true)
+  end
+
+  def args_to_internal_representation({[ip: true], _, _}) do
+    {_, ip} = Networking.IPLookup.RemoteDiscovery.get_node_ip()
```
samuel-uniris, line 59: I suggest you to use Networking.IPLookup.get_node_ip to leverage fallback for NAT discovery as well, 

you will also need to setup the default  IPLookup as in dev, it's Static

```elixir
Application.put_env(:archethic, Archethic.Networking.IPLookup,
Archethic.Networking.IPLookup.NATDiscovery
```

lib/mix/tasks/nettools.ex
```diff
@@ -0,0 +1,66 @@
...
+    |> args_to_internal_representation()
+  end
+
+  def args_to_internal_representation({[help: true], [], []}) do
+    Mix.shell().cmd("mix help #{Mix.Task.task_name(__MODULE__)}")
+  end
+
+  def args_to_internal_representation({[punch: true], _port_list = [port | more_ports], errors}) do
+    Networking.PortForwarding.try_open_port(abs(Integer.parse(port) |> elem(0)), true)
+    args_to_internal_representation({[punch: true], more_ports, errors})
```
samuel-uniris, line 50: You might also need to set env variable as well otherwise this will be skipped.

Also `PortForwarding` must be updated at https://github.com/archethic-foundation/archethic-node/blob/165fc8c2664037b25d0d6e7d0b3b91055d2b439d/lib/archethic/networking/port_forwarding.ex#L35 to set `NATDiscovery`

lib/mix/tasks/nettools.ex
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  def args_to_internal_representation({[punch: true], _port_list = [], _errors}) do
+    Networking.PortForwarding.try_open_port(40_000, true)
+    Networking.PortForwarding.try_open_port(30_002, true)
+  end
+
+  def args_to_internal_representation({[ip: true], _, _}) do
+    {_, ip} = Networking.IPLookup.RemoteDiscovery.get_node_ip()
+    Logger.info("Public ip: #{:inet.ntoa(ip)}")
```
samuel-uniris, line 60: I also suggest to remove this line, as this will redundant print from the `IPLookup.get_node_ip` function

samuel-uniris (changes requested):
I pointed it some comments to make it more realistic
I tested at home, with those changes it's working

lib/mix/tasks/nettools.ex
apoorv-2204: Sincere apologies. my bad,

samuel-uniris: No harm 😉

lib/mix/tasks/nettools.ex
```diff
@@ -41,25 +43,38 @@ defmodule Mix.Tasks.Archethic.Nettools do
...
+  # For switch --help or -h
   def args_to_internal_representation({[help: true], [], []}) do
     Mix.shell().cmd("mix help #{Mix.Task.task_name(__MODULE__)}")
   end
 
-  def args_to_internal_representation({[punch: true], _port_list = [port | more_ports], errors}) do
-    Networking.PortForwarding.try_open_port(abs(Integer.parse(port) |> elem(0)), true)
-    args_to_internal_representation({[punch: true], more_ports, errors})
+  # For switch --punch or -p Opens default http and p2p ports
+  def args_to_internal_representation({[punch: true], _port_list = [], _errors}) do
```
samuel-uniris, line 52: For the punch you also need to set the env, as it's checking IP lookup or enforced by conf

lib/mix/tasks/nettools.ex
```diff
@@ -41,25 +43,38 @@ defmodule Mix.Tasks.Archethic.Nettools do
...
   def args_to_internal_representation({[ip: true], _, _}) do
-    {_, ip} = Networking.IPLookup.RemoteDiscovery.get_node_ip()
-    Logger.info("Public ip: #{:inet.ntoa(ip)}")
+    default = Application.get_env(:archethic, Networking.IPLookup)
+    # The default IPLookup value must be restored
+    try do
+      Application.put_env(:archethic, Networking.IPLookup, Networking.IPLookup.NATDiscovery)
+      Networking.IPLookup.get_node_ip()
+    after
+      Application.put_env(:archethic, Networking.IPLookup, default)
```
samuel-uniris, line 73: If it crashes I don't think you need to restore as the program will stop .

samuel-uniris (approved):

PR #343, opened 2022-05-31
Simplification of Origin key management

### LINKED ISSUE #326: Simplification of origin key management
### Problem to solve

To secure transaction we are using `origin signature` which is some kind of security to ensure a transaction have been forged by an authorized device or origin (app)

But to enlist an origin into the system the complex was difficult and the setup no scalable being interactive (according to the patents)

### Solution

I propose a new solution which is kind of proposal to support easy enlisting of origin devices but removing the concept of origin shared secrets.

Then, we can list all the origin keys for a given family (by certificate) with its own chain.

Each device would have to send a request to the network by passing its public key and its certificate, to be validated according to the CA available (TPM/Yubikey, Ledger, ...)

Once validated the public key will be listed in the origin public keys

Then the device will use its direct private key to sign the transaction and then the  nodes will certify with the list of public keys available in the network.

This scheme makes the system more scalable, easier to implement and to understand

This issue would require:
### Knowledge
- [x] Understanding the `api/origin_controller`
- [x] Finding the modules that gets affected on making changes
- [x] Understanding the OriginSignature part in transaction
### Coding Changes
- [x] Adapt API to add origin public key
- [x] Remove the API to get origin encrypted shared key
 - [x] Add certificate inside the origin tx
- [x] Changing the Existing function to fetch origin keys
- [x] Updated the explorer view with new origin tx content
- [x] Adapt Existing Test Cases with Changes
- [x] Dependency on libjs to adapt the behaviour

blackode:
# Description

Simplification of Origin key management with public_key and certificate

Fixes #326 

## Type of change

# Checklist:

### DIFF
--- a/lib/archethic/bootstrap/network_init.ex
+++ b/lib/archethic/bootstrap/network_init.ex
@@ -31,7 +31,6 @@ defmodule Archethic.Bootstrap.NetworkInit do
   alias Archethic.TransactionChain.TransactionData.Ledger
   alias Archethic.TransactionChain.TransactionData.UCOLedger
   alias Archethic.TransactionChain.TransactionData.UCOLedger.Transfer
-  alias Archethic.TransactionChain.TransactionData.Ownership
 
   alias Archethic.TransactionChain.TransactionSummary
 
@@ -83,33 +82,25 @@ defmodule Archethic.Bootstrap.NetworkInit do
   @doc """
   Create the first origin shared secret transaction
   """
-  @spec init_software_origin_shared_secrets_chain() :: :ok
-  def init_software_origin_shared_secrets_chain do
+  @spec init_software_origin_chain() :: :ok
+  def init_software_origin_chain do
     Logger.info("Create first software origin shared secret transaction")
 
-    origin_seed = :crypto.strong_rand_bytes(32)
-    secret_key = :crypto.strong_rand_bytes(32)
     signing_seed = SharedSecrets.get_origin_family_seed(:software)
 
-    # Default keypair generation creates software public key
-    {origin_public_key, origin_private_key} = Crypto.generate_deterministic_keypair(origin_seed)
-
-    encrypted_origin_private_key = Crypto.aes_encrypt(origin_private_key, secret_key)
+    [genesis_origin_public_key | _rest] = @genesis_origin_public_keys
 
     Transaction.new(
-      :origin_shared_secrets,
+      :origin,
       %TransactionData{
         code: """
           condition inherit: [
             # We need to ensure the type stays consistent
             # So we can apply specific rules during the transaction validation
-            type: origin_shared_secrets
+            type: origin
           ]
         """,
-        content: <<origin_public_key::binary>>,
-        ownerships: [
-          Ownership.new(encrypted_origin_private_key, secret_key, @genesis_origin_public_keys)
-        ]
+        content: <<genesis_origin_public_key::binary>>
       },
       signing_seed,
       0
--- a/lib/archethic/bootstrap/network_init.ex
+++ b/lib/archethic/bootstrap/network_init.ex
@@ -100,7 +100,7 @@ defmodule Archethic.Bootstrap.NetworkInit do
             type: origin
           ]
         """,
-        content: <<genesis_origin_public_key::binary>>
+        content: <<genesis_origin_public_key::binary, 0::16>>
       },
       signing_seed,
       0
--- a/lib/archethic/bootstrap/network_init.ex
+++ b/lib/archethic/bootstrap/network_init.ex
@@ -97,7 +97,8 @@ defmodule Archethic.Bootstrap.NetworkInit do
           condition inherit: [
             # We need to ensure the type stays consistent
             # So we can apply specific rules during the transaction validation
-            type: origin
+            type: origin,
+            content: true
           ]
         """,
         content: <<genesis_origin_public_key::binary, 0::16>>
--- a/lib/archethic/mining/pending_transaction_validation.ex
+++ b/lib/archethic/mining/pending_transaction_validation.ex
@@ -160,6 +160,28 @@ defmodule Archethic.Mining.PendingTransactionValidation do
   end
 
   defp do_accept_transaction(%Transaction{
+         type: :origin,
+         data: %TransactionData{
+           content: content
+         }
+       }) do
+    with {<<_curve_id::8, _origin_id::8, origin_public_key::binary>>, key_certificate} <-
+           Utils.deserialize_public_key(content),
+         root_ca_public_key <-
+           Crypto.get_root_ca_public_key(origin_public_key),
+         true <-
+           Crypto.verify_key_certificate?(origin_public_key, key_certificate, root_ca_public_key) do
+      :ok
+    else
+      false ->
+        {:error, "Invalid Origin transaction with invalid certificate"}
+
+      _ ->
+        {:error, "Invalid Origin transaction's content"}
+    end
+  end
+
+  defp do_accept_transaction(%Transaction{
          type: :node_shared_secrets,
          data: %TransactionData{
            content: content,
--- a/lib/archethic/mining/pending_transaction_validation.ex
+++ b/lib/archethic/mining/pending_transaction_validation.ex
@@ -167,7 +167,6 @@ defmodule Archethic.Mining.PendingTransactionValidation do
        }) do
     with {origin_public_key, rest} <-
            Utils.deserialize_public_key(content),
-         <<_curve_id::8, _origin_id::8, origin_public_key::binary>> <- origin_public_key,
          <<key_certificate_size::16, key_certificate::binary-size(key_certificate_size),
            _::binary>> <- rest,
          root_ca_public_key <-
--- a/lib/archethic/mining/pending_transaction_validation.ex
+++ b/lib/archethic/mining/pending_transaction_validation.ex
@@ -165,8 +165,10 @@ defmodule Archethic.Mining.PendingTransactionValidation do
            content: content
          }
        }) do
-    with {<<_curve_id::8, _origin_id::8, origin_public_key::binary>>, key_certificate} <-
+    with {<<_curve_id::8, _origin_id::8, origin_public_key::binary>>, rest} <-
            Utils.deserialize_public_key(content),
+         <<key_certificate_size::16, key_certificate::binary-size(key_certificate_size),
+           _::binary>> <- rest,
          root_ca_public_key <-
            Crypto.get_root_ca_public_key(origin_public_key),
          true <-
--- a/lib/archethic/mining/pending_transaction_validation.ex
+++ b/lib/archethic/mining/pending_transaction_validation.ex
@@ -165,8 +165,9 @@ defmodule Archethic.Mining.PendingTransactionValidation do
            content: content
          }
        }) do
-    with {<<_curve_id::8, _origin_id::8, origin_public_key::binary>>, rest} <-
+    with {origin_public_key, rest} <-
            Utils.deserialize_public_key(content),
+         <<_curve_id::8, _origin_id::8, origin_public_key::binary>> <- origin_public_key,
          <<key_certificate_size::16, key_certificate::binary-size(key_certificate_size),
            _::binary>> <- rest,
          root_ca_public_key <-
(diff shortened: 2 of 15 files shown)


lib/archethic_web/controllers/api/origin_key_controller.ex
```diff
@@ -3,35 +3,33 @@ defmodule ArchethicWeb.API.OriginKeyController do
...
+      signing_seed = SharedSecrets.get_origin_family_seed(origin_family)
+
+      tx =
+        Transaction.new(
+          :origin_shared_secrets,
+          %TransactionData{
+            content: <<origin_public_key::binary>>
+          },
+          signing_seed,
+          0
```
samuel-uniris, line 25: You should use the last index of the chain, as few devices can request authorization, and using 0 will cause "transaction already exists" error

lib/archethic_web/controllers/api/origin_key_controller.ex
```diff
@@ -3,35 +3,33 @@ defmodule ArchethicWeb.API.OriginKeyController do
...
-          Ownership.get_encrypted_key(ownership, origin_public_key) |> Base.encode16()
-      }
+         true <- Crypto.get_key_certificate(origin_public_key) == certificate,
+         <<_curve_id::8, origin_id::8, _rest::binary>> <- origin_public_key do
+      origin_family = Crypto.key_origin(origin_id)
+      signing_seed = SharedSecrets.get_origin_family_seed(origin_family)
+
+      tx =
+        Transaction.new(
+          :origin_shared_secrets,
```
samuel-uniris, line 20: I think you should also change the type to something like`origin` as it's not anymore a shared secret

This will have an impact on https://github.com/archethic-foundation/archethic-node/blob/dcee81f358b450f798f3f87e6cd8b34c527b0c4d/lib/archethic/shared_secrets/mem_tables_loader.ex#L87 


Also regarding the transaction changes you should adapt the transaction loading from the new data encoding

samuel-uniris (changes requested):

Neylix:
Nice for the API !
There is also some change to do during network init when initializing the first software origin key.
https://github.com/archethic-foundation/archethic-node/blob/76cefaceccc1cf0291622be10e473cd12d148475/lib/archethic/bootstrap/network_init.ex#L86
You may change the transaction as the one in the your API

lib/archethic_web/controllers/api/origin_key_controller.ex
blackode: Yes. Updated index from chain_length of last address

lib/archethic_web/views/explorer_view.ex
```diff
@@ -199,10 +199,10 @@ defmodule ArchethicWeb.ExplorerView do
     """
   end
 
-  def format_transaction_content(:origin_shared_secrets, content) do
+  def format_transaction_content(:origin, content) do
```
samuel-uniris, line 202: I think we have to change this as we have a single chain by family now so the content should be updated and simplified

lib/archethic_web/controllers/api/origin_key_controller.ex
```diff
@@ -3,35 +3,37 @@ defmodule ArchethicWeb.API.OriginKeyController do
...
+         {:ok, last_index} <- Archethic.get_transaction_chain_length(last_address) do
+      <<_curve_id::8, origin_id::8, _rest::binary>> = origin_public_key
+      origin_family = Crypto.key_origin(origin_id)
+      signing_seed = SharedSecrets.get_origin_family_seed(origin_family)
+
+      tx =
+        Transaction.new(
+          :origin,
+          %TransactionData{
+            content: <<origin_public_key::binary>>
```
samuel-uniris, line 26: We should add the certificate in the content as it should be validated by the network.
Also the check of the certificate is not correct as we have not the certificate in the node config but only the root CA.
So the best is to send the transaction and let the validation nodes makes the check like in there https://github.com/archethic-foundation/archethic-node/blob/76cefaceccc1cf0291622be10e473cd12d148475/lib/archethic/mining/pending_transaction_validation.ex#L132

samuel-uniris (changes requested):

lib/archethic_web/controllers/api/origin_key_controller.ex
blackode: This means we have to concat certificate along with public key right?
Do we have any standard byte size for publickey to set and retrieve second part as a content?

blackode: Yeah, updated the validation part for certificate adding inside the content. I have added new private function to match transaction type `origin`.

lib/archethic/mining/pending_transaction_validation.ex
```diff
@@ -159,6 +159,28 @@ defmodule Archethic.Mining.PendingTransactionValidation do
     end
   end
 
+  defp do_accept_transaction(%Transaction{
+         type: :origin,
+         data: %TransactionData{
+           content: content
+         }
+       }) do
+    with {<<_curve_id::8, _origin_id::8, origin_public_key::binary>>, key_certificate} <-
```
samuel-uniris, line 168: I think you can extract the certificate like this: https://github.com/archethic-foundation/archethic-node/blob/b8a0952e1a5e624e5c9f0ead0da898bdf8bbd195/lib/archethic/p2p/node.ex#L87 
by encoding the certificate size we can have a better control of the deserialized data

lib/archethic_web/views/explorer_view.ex
```diff
@@ -199,10 +199,10 @@ defmodule ArchethicWeb.ExplorerView do
...
     get_origin_public_keys(content, %{software: [], hardware: []})
```
samuel-uniris, line 203: I think you have to change this function as it's including the certificate as well.
Also this function have only one key now

samuel-uniris (changes requested):

lib/archethic_web/controllers/api/origin_key_controller.ex
```diff
@@ -3,35 +3,36 @@ defmodule ArchethicWeb.API.OriginKeyController do
...
-         ownership when ownership != nil <-
-           Enum.find(tx.data.ownerships, fn ownership ->
-             Ownership.authorized_public_key?(ownership, origin_public_key)
-           end) do
-      res = %{
-        encrypted_origin_private_keys: Base.encode16(ownership.secret),
-        encrypted_secret_key:
-          Ownership.get_encrypted_key(ownership, origin_public_key) |> Base.encode16()
-      }
+         address <- Crypto.derive_address(origin_public_key),
```
samuel-uniris, line 13: The address of the chain should not be based on the origin public key passed but from a well known key derivation as done before.

In your case the chain length will always be 0, if the given origin keys are different.

Using the signing seed, you will be able to generate the 0th key and then fetch the chain length.

So, more we will have clients using a given family, more the chain will evolve, which help to identify the keys to load.

samuel-uniris (changes requested):

samuel-uniris:
Overall this seems ok
I would like you add like this on the transaction controller api

https://github.com/archethic-foundation/archethic-node/blob/b8a0952e1a5e624e5c9f0ead0da898bdf8bbd195/lib/archethic_web/controllers/api/transaction_controller.ex#L30-L37

So a client can wait completion of the transaction asynchronously

Also you should catch error here:
https://github.com/archethic-foundation/archethic-node/pull/343/files#diff-4cff92ecfcf860d5104b94c8c0573af7926b9e088a1c4dcec00f242af807ecbaR35

blackode:
> Overall this seems ok I would like you add like this on the transaction controller api
> 
> https://github.com/archethic-foundation/archethic-node/blob/b8a0952e1a5e624e5c9f0ead0da898bdf8bbd195/lib/archethic_web/controllers/api/transaction_controller.ex#L30-L37
> 
> So a client can wait completion of the transaction asynchronously
> 
> Also you should catch error here: [#343 (files)](https://github.com/archethic-foundation/archethic-node/pull/343/files#diff-4cff92ecfcf860d5104b94c8c0573af7926b9e088a1c4dcec00f242af807ecbaR35)

I have updated the code with error handling and some refactoring. Please do check @samuel-uniris

lib/archethic_web/controllers/api/origin_key_controller.ex
```diff
@@ -3,63 +3,82 @@ defmodule ArchethicWeb.API.OriginKeyController do
...
+      |> SharedSecrets.get_origin_family_seed()
+
+    {first_origin_family_public_key, _} = Crypto.derive_keypair(signing_seed, 0)
 
-      conn
-      |> put_status(:ok)
-      |> json(res)
+    address = Crypto.derive_address(first_origin_family_public_key)
+
+    with {:ok, last_address} <- Archethic.get_last_transaction_address(address),
```
Neylix, line 38: You don't need to get the last address of the chain as `Archethic.get_transaction_chain_length` always return the chain length whatever is the address.
Also I think that `origin` transactions will be stored by all node, so you may not need to ask chain index from other node but directly from the running node itself by using `TransactionChain.size(address)`

Neylix (changes requested):

lib/archethic_web/controllers/api/origin_key_controller.ex
```diff
@@ -3,63 +3,82 @@ defmodule ArchethicWeb.API.OriginKeyController do
...
+         {:ok, last_index} <- Archethic.get_transaction_chain_length(last_address) do
+      tx_content = <<origin_public_key::binary, certificate::binary>>
```
Neylix, line 40: Durgin mining validation you deserialize the content using `key_certificate_size::16, key_certificate::binary-size(key_certificate_size)` but here you don't add the certificate size to the content. You should add `byte_size(certificate)::16`

lib/archethic_web/controllers/api/origin_key_controller.ex
samuel-uniris: Good point

lib/archethic/bootstrap/network_init.ex
```diff
@@ -83,33 +82,25 @@ defmodule Archethic.Bootstrap.NetworkInit do
...
     Transaction.new(
-      :origin_shared_secrets,
+      :origin,
       %TransactionData{
         code: """
           condition inherit: [
             # We need to ensure the type stays consistent
             # So we can apply specific rules during the transaction validation
-            type: origin_shared_secrets
+            type: origin
```
Neylix, line 100: There is a problem here, the next transaction are refused because it is expected to have a nil content.
The code should be this to avoid controlling the content : 
```
condition inherit: [
  # We need to ensure the type stays consistent
  # So we can apply specific rules during the transaction validation
    type: origin,

    # We discard the content and code verification
    content: true
  ]
```
Also this code have to be added in transaction created by the API

lib/archethic/shared_secrets/mem_tables_loader.ex
```diff
@@ -86,7 +86,7 @@ defmodule Archethic.SharedSecrets.MemTablesLoader do
 
   def load_transaction(%Transaction{
         address: address,
-        type: :origin_shared_secrets,
+        type: :origin,
         data: %TransactionData{content: content}
       }) do
     content
```
Neylix, line 92: There is also a bug here, as now we have only one key and a certificate in the content of the transaction, we should deserialize it as in the mining part to get the public key.

lib/archethic_web/controllers/api/origin_key_controller.ex
```diff
@@ -3,63 +3,85 @@ defmodule ArchethicWeb.API.OriginKeyController do
...
+    |> json(response)
+  end
+
+  defp prepare_transaction(origin_id, origin_public_key, certificate) do
+    signing_seed =
+      origin_id
+      |> Crypto.key_origin()
+      |> SharedSecrets.get_origin_family_seed()
+
+    {first_origin_family_public_key, _} = Crypto.derive_keypair(signing_seed, 0)
```
Neylix, line 35: Here you have to use the public key at index 1, because the first transaction address of the chain is the hash of the public key at index + 1.
SO the good way is `Crypto.derive_keypair(signing_seed, 1)`

lib/archethic/bootstrap/network_init.ex
Neylix: There is also this one to add `content: true`

lib/archethic_web/controllers/api/origin_key_controller.ex
blackode: This has been updated

Neylix (approved):

samuel-uniris (approved):

PR #351, opened 2022-06-02
Improve location using GeoLite2

Neylix:
# Description

Geo localisation is implemented with ip2location database, but I figured that this db wasn't really accurate for several IPs and often return main ISP router location than user router location.
Maxmind provide a free location DB (GeoLite2) which is more accurate than ip2location, so I propose to use this one instead

## Type of change

# How Has This Been Tested?

Tested with several IP in France and India, location are way better.

# Checklist:


config/config.exs
```diff
@@ -136,6 +136,15 @@ config :archethic, Archethic.Networking.IPLookup.NATDiscovery,
 config :archethic, Archethic.Networking.IPLookup.RemoteDiscovery,
   provider: Archethic.Networking.IPLookup.RemoteDiscovery.IPIFY
 
+config :geolix,
+  databases: [
+    %{
+      id: :city,
+      adapter: Geolix.Adapter.MMDB2,
```
apoorv-2204, line 143: This is really nice, I was not able to find the mamind db . Sam will be really happy : ) nice

apoorv-2204: We will now have very less offset . We can now think about Network patches ?

samuel-uniris: While it might improve location, network patch is something which might improve P2P calls, but not really priority for now.

Neylix:
> https://github.com/elixir-geolix/adapter_mmdb2 might be lighter choice so we could add an impl for MaxMind instead of Ip2location without changing too much configuration, and supervision tree, ...
> 
> As we can call the function directly from the lib.
> 
> Otherwise we could embrace Geolix as abstraction but might tight-coupled this part of the code to this library

Maybe directly use https://hexdocs.pm/mmdb2_decoder/MMDB2Decoder.html
Instead of using Geolix

samuel-uniris:
I think you might use directly the GenServer, ETS is great when you are dealing with high read concurrency but here I don't think you will face bottlenecks with a plain GenServer .

Before we used ETS because the way

mix.exs
```diff
@@ -75,6 +75,7 @@ defmodule Archethic.MixProject do
 
       # P2P
       {:ranch, "~> 2.1", override: true},
+      {:mmdb2_decoder, "~> 3.0"},
```
apoorv-2204, line 78: Is it better than maxmind geo database?

Neylix: This is the same database, but just use a lighter library to fetch the data.
Geolix and GeolixAdapterMmdb2 were to eavy and finaly use Mmdb2Decoder

samuel-uniris (approved):

PR #368, opened 2022-06-08
Change transaction value fee to a fixed one (base fee)

### LINKED ISSUE #349: Change transaction fee with base fee
We should adapt the transaction fee algorithm to use a base fee (flat tax) to avoid transaction's value fee .
This causing unfair fee calculation and not efficient with non financial transaction.

A first assumption should be 0.1 UCO fee (~1ct) (via OracleChain from the USD/EUR price)

blackode:
# Description
The transaction fee value for oracle transactions have been set with `0.01/uco_price_in_usd` 

Fixes #349 

## Type of change

### DIFF
--- a/lib/archethic/mining/fee.ex
+++ b/lib/archethic/mining/fee.ex
@@ -42,6 +42,9 @@ defmodule Archethic.Mining.Fee do
       true == Transaction.network_type?(type) ->
         0
 
+      type == :oracle ->
+        0.01 / uco_price_in_usd
+
       true ->
         transaction_value = get_transaction_value(tx) / @unit_uco
         nb_recipients = get_number_recipients(tx)
--- a/lib/archethic/mining/fee.ex
+++ b/lib/archethic/mining/fee.ex
@@ -42,11 +42,8 @@ defmodule Archethic.Mining.Fee do
       true == Transaction.network_type?(type) ->
         0
 
-      type == :oracle ->
-        0.01 / uco_price_in_usd
-
       true ->
-        transaction_value = get_transaction_value(tx) / @unit_uco
+        transaction_value = 0.01 / uco_price_in_usd
         nb_recipients = get_number_recipients(tx)
         nb_bytes = get_transaction_size(tx)
         nb_storage_nodes = get_number_replicas(tx)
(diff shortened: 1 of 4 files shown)


lib/archethic/mining/fee.ex
```diff
@@ -42,11 +42,8 @@ defmodule Archethic.Mining.Fee do
       true == Transaction.network_type?(type) ->
         0
 
-      type == :oracle ->
-        0.01 / uco_price_in_usd
-
       true ->
-        transaction_value = get_transaction_value(tx) / @unit_uco
+        transaction_value = 0.01 / uco_price_in_usd
```
samuel-uniris, line 46: I think you should make the change here:

https://github.com/archethic-foundation/archethic-node/blob/3c24793b3c00085b5f3ccfabafe8ca557a604e99/lib/archethic/mining/fee.ex#L99

As there is no more transaction value fee

lib/archethic/mining/fee.ex
```diff
@@ -95,14 +86,13 @@ defmodule Archethic.Mining.Fee do
...
          uco_price_in_usd,
-         transaction_value,
          nb_bytes,
          nb_storage_nodes,
          nb_recipients
        ) do
     # TODO: determine the fee for smart contract execution
 
-    value_cost = fee_for_value(uco_price_in_usd, transaction_value)
+    value_cost = 0.01 / uco_price_in_usd
```
samuel-uniris, line 95: I think you can rename it as "minimum fee"
You could also use a constant for the 0.01 , this will be easier to update.

test/archethic/mining/fee_test.exs
```diff
@@ -20,7 +20,7 @@ defmodule Archethic.Mining.FeeTest do
 
     test "should return a fee less than amount to send for a single transfer" do
       # 0.050048 UCO for 1 UCO at $0.2
-      assert 50_048_000 =
```
samuel-uniris, line 23: Can you also change the comments values

blackode: I have updated all fee comments with new value by dividing with `100000000`

samuel-uniris (approved):

PR #369, opened 2022-06-08
Implement read quorum

### LINKED ISSUE #332: Support read quorum for P2P
## Problem to solve

In distributed system when we are dealing with several replicas sharding the same information, during the read time we want to avoid as much as possible stale data, to be up to date.

Current implementation request most of the time the nearest node but this can make consistency issues and retrieved not accurate data.

### First proposal

We can implement a read quorum policy to query few nodes and except a set of nodes to respond accordingly to the policy.

Usually in distributed system the rule of quorum is: `reads + writes > replicas`

Where :
- the reads are the number of nodes which reply to the read request
- the writes are the number of nodes which confirm the write
- the replicas is the entire set of replicas

In Archethic's case, the writes are the confirmations coming from the atomic commitment from a validation node point of view, where the beacon chain and the client received a replication attestation. (subtree of the replication tree)

Then we can deduct a number reads from the election algorithm of the replicas and the validation nodes: `reads = writes - replicas`

But sometimes we also want to apply some acceptance rules from the client request, then we should also support extensibility by given an acceptance function for a set of replies (winner can be the longest chain, or the latest update)

### Second proposal 

While the first approach seems logical in most of distributed system and databases, as the number of replicas will be high, in term of performance the cost of consistent reads becomes high (can be N/2 or N/3).

Also the integration becomes more complex as we have to define conflict resolutions functions and setup to determine the number of replication attestations for each information. However sometimes we are requesting N request because there are not transaction validations for some queries.

Also because Archethic leverages atomic commitment we can use it in the queries as well.
Then we can introduce `atomic reads` to be sure all the nodes requested have to return the same results otherwise we pick others or return failure.

With this approach we can also leverage the network coordinates to ensure the data loading from the nearest nodes

**Remarks**: 
> Problem with this approach is we can have huge amount of failures, as the consistency is not guarantee in a distributed system because of lateness

### Third proposal

Another interesting approach is `monotonic quorum reads`  where in  2 successive quorum reads, it’s guaranteed the 2nd one won’t get something older than the 1st one.

### Integrations:

In all the cases, the function must be integrated in the following modules:
- [x] replication module
- [x] mining module
- [x] self repair module 
- [x] beacon chain module 
- [x] transaction chain module 
- [x] top module 
- [x] live chain explorers

Finally we need to:
- [ ] Ensure non regression

samuel-uniris:
# Description

Fixes #332 

## Type of change

# How Has This Been Tested?

- Test A
- Test B

# Checklist:


lib/archethic/transaction_chain.ex
```diff
@@ -553,4 +587,206 @@ defmodule Archethic.TransactionChain do
...
+  end
+
+  defp fetch_transaction_chain(
+         nodes,
+         address,
+         paging_state
+       ) do
+    conflict_resolver = fn results ->
+      results
+      |> Enum.sort(&(&1.more? and length(&1.transactions) > length(&2.transactions)))
```
Neylix, line 681: This resolver will not work, for exemple if you have `a={more?: false, length: 6}` and `b={more?: false, length: 5}`
as `a.more? == false` the sorted result will be [b,a] but we want [a,b] as a is the longuest chain.
I would recommand something like this
`Enum.sort(&((&1.more? and !&2.more?) or length(&1.transactions) > length(&2.transactions))`

Neylix (changes requested):

Neylix (approved):

PR #375, opened 2022-06-13
Integrate MiniUPNP

### LINKED ISSUE #346: Support PCP for UPnP discovery and port forwarding
## Problem to solve

Our current implementation of UPnP is using UPnP v1/v2 and NAT PMP.
If it's working in some router, there is gap between miniupnpc and our implementation.
The gap would be the PCP UPnP support, which is the replacement of NAT PMP.

## Solution

We should support this implementation described here: http://miniupnp.free.fr/nat-pmp.html, to have better support UPnP networking discovery and port forwarding .

We should implement PCP to handle:

- [x] Integrate miniupnpc

samuel-uniris:
# Description

Integrate miniupnp as C dependency and build locally the binary to use it via system calls.
This way we are reproducing the same behavior.

This is quick fix to do not have to recreate in Erlang or Elixir the UPnP PCP implementation while keeping the support for NAT PMP and UPnP v1/v2

Fixes #346 

## Type of change

# How Has This Been Tested?

Tested locally using `mix archethic.nettools` on Mac OS X (without upnpc installed) 

# Checklist:


Makefile
```diff
@@ -8,6 +8,9 @@ compile_c_programs:
 	mkdir -p priv/c_dist
 	$(CC) src/c/crypto/stdio_helpers.c src/c/crypto/ed25519.c -o priv/c_dist/libsodium_port -I src/c/crypto/stdio_helpers.h -lsodium
 	$(CC) src/c/hypergeometric_distribution.c -o priv/c_dist/hypergeometric_distribution -lgmp
+	
+	$(MAKE) -C src/c/nat/miniupnp/miniupnpc
```
apoorv-2204, line 12:
```/usr/bin/gcc src/c/hypergeometric_distribution.c -o priv/c_dist/hypergeometric_distribution -lgmp
make -C src/c/nat/miniupnp/miniupnpc
make[1]: *** src/c/nat/miniupnp/miniupnpc: No such file or directory.  Stop.
make: *** [Makefile:12: compile_c_programs] Error 2```

File: .gitmodules [hunk 2]
```diff
@@ -0,0 +1,3 @@
+[submodule "src/c/nat/miniupnp"]
+	path = src/c/nat/miniupnp
+	url = https://github.com/miniupnp/miniupnp
```
apoorv-2204, line 3: I think git failed to fetch, this lib

apoorv-2204:
It is working locally for my pc.

- [ ] image
- [ ]

.gitmodules
apoorv-2204: - [ ] git submodule update --force --recursive --init --remote

samuel-uniris: I think we can insert it into the Makefile

lib/archethic/networking/ip_lookup/nat_discovery/miniupnp.ex
```diff
@@ -0,0 +1,65 @@
...
+          |> to_charlist()
+          |> :inet.parse_address()
+          |> elem(1)
+
+        {_, status} ->
+          {:error, status}
+      end
+
+    opts = [
+      # Add redirection
```
apoorv-2204, line 42: `-e "ArchethicPort"`
 Is it a good thing to have this?

samuel-uniris: why not

apoorv-2204: Do I do a that by committing on add_miniupnp?

samuel-uniris: I will do it ;)

PR #376, opened 2022-06-15
Detection of unresponsive trigger nodes

### LINKED ISSUE #338: Detect unresponsive trigger node
### Problem to solve

We are using some trigger node based on election algorithm to send transaction in case of:
- Oracle
- Node shared secrets
- Contract triggers
...

For instance, the first node of the election will be the one to emit the transaction.

However in case of this node is failing or malicious, if the transaction is not sent, we would have the next one, but as the election may not change, the same issue can occurs.

We then have to find a solution to support fallback of trigger nodes

### Solution

We can implement some kind of process which is waiting some events to occur to determine if a transaction or an action have been submitted.

This solution should be enough generic to support multiple implementation based on the several context we are using.

An idea could be to modelize it trough finite state machine like that:
```mermaid
stateDiagram-v2

[*] --> Waiting
Waiting --> SoftTimeout: after some time
SoftTimeout --> Waiting: resend transaction
Waiting --> Confirmed:  submission notified
Confirmed --> [*]
Waiting --> HardTimeout: no response for all the workers
HardTimeout --> [*]
```
Tasks
- [x] - Add FSM with coding steps below 
- [x] - Add send start/exit  message signals to indicate start end states of FSM in NewTransactions
- [x] - Implement resend until confirm or hardtimeout if went to softtimeout from waiting state and if hardtimeout reset

### Coding Steps

 Implement a FSM with the following: 
- [x] - 1. 4 states i.e Waiting, SoftTimeout, Confirmed, HardTimeout
- [x] - 2. Internally perfrom resend based on timouts

imnik11:
# Description

It detects the unresponsive nodes and resends the transaction to the next trigger next node.

Fixes #338

## Type of change

# How Has This Been Tested?
Test with 3 three nodes
`>ARCHETHIC_CRYPTO_SEED=node1 ARCHETHIC_P2P_PORT=3001 ARCHETHIC_HTTP_PORT=4001 iex -S mix`
similary change the seed, ports for other nodes.
If we stop the node1 the node shared secret transactions , oracle etc must be fall back to node2, node2 as trigger node and shall be validated.

# Checklist:

### DIFF
--- a/lib/archethic/contracts/worker.ex
+++ b/lib/archethic/contracts/worker.ex
@@ -32,7 +32,7 @@ defmodule Archethic.Contracts.Worker do
   alias Archethic.TransactionChain.Transaction.ValidationStamp.LedgerOperations.TransactionMovement
 
   alias Archethic.Utils
-
+  alias Archethic.Utils.DetectNodeResponsiveness
   require Logger
 
   use GenServer
@@ -256,21 +256,35 @@ defmodule Archethic.Contracts.Worker do
   defp schedule_trigger(_), do: :ok
 
   defp handle_new_transaction(next_transaction = %Transaction{}) do
-    [%Node{first_public_key: key} | _] =
-      next_transaction
-      |> Transaction.previous_address()
-      |> Election.chain_storage_nodes(P2P.authorized_and_available_nodes())
-
+    validation_nodes = P2P.authorized_and_available_nodes()
     # The first storage node of the contract initiate the sending of the new transaction
-    if key == Crypto.first_node_public_key() do
-      validation_nodes = P2P.authorized_and_available_nodes()
-
+    if trigger_node?(next_transaction) do
       P2P.broadcast_message(validation_nodes, %StartMining{
         transaction: next_transaction,
         validation_node_public_keys: Enum.map(validation_nodes, & &1.last_public_key),
         welcome_node_public_key: Crypto.last_node_public_key()
       })
+    else
+      DetectNodeResponsiveness.start_link(next_transaction.address, fn count ->
+        if trigger_node?(count) do
+          P2P.broadcast_message(validation_nodes, %StartMining{
+            transaction: next_transaction,
+            validation_node_public_keys: Enum.map(validation_nodes, & &1.last_public_key),
+            welcome_node_public_key: Crypto.last_node_public_key()
+          })
+        end
+      end)
     end
+  end
+
+  defp trigger_node?(next_transaction = %Transaction{}, count \\ 0) do
+    %Node{first_public_key: key} =
+      next_transaction
+      |> Transaction.previous_address()
+      |> Election.chain_storage_nodes(P2P.authorized_and_available_nodes())
+      |> Enum.at(count)
+
+    key == Crypto.first_node_public_key()
   end
 
   defp chain_transaction(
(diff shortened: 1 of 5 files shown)


lib/archethic/contracts/worker.ex
```diff
@@ -256,23 +256,37 @@ defmodule Archethic.Contracts.Worker do
...
-
+    if trigger_node?(next_transaction) do
       P2P.broadcast_message(validation_nodes, %StartMining{
         transaction: next_transaction,
         validation_node_public_keys: Enum.map(validation_nodes, & &1.last_public_key),
         welcome_node_public_key: Crypto.last_node_public_key()
       })
+    else
+      DetectNodeResponsiveness.start_link(next_transaction.address, fn count ->
+        if trigger_node?(count) do
```
samuel-uniris, line 269: I think you need the `next_transaction` variable

lib/archethic/contracts/worker.ex
```diff
@@ -256,23 +256,37 @@ defmodule Archethic.Contracts.Worker do
...
+        end
+      end)
     end
   end
 
+  defp trigger_node?(next_transaction = %Transaction{}, count \\ 0) do
+    %Node{first_public_key: key} =
+      next_transaction
+      |> Transaction.previous_address()
+      |> Election.chain_storage_nodes(P2P.authorized_and_available_nodes())
```
samuel-uniris, line 284: you might reuse: `validation_nodes` here

lib/archethic/oracle_chain/scheduler.ex
```diff
@@ -140,20 +140,59 @@ defmodule Archethic.OracleChain.Scheduler do
...
       if Enum.empty?(new_oracle_data) do
         Logger.debug("Oracle transaction skipped - no new data")
       else
-        send_polling_transaction(new_oracle_data, index, summary_date)
+        send_polling_transaction(tx)
       end
     else
-      Logger.debug("Oracle transaction skipped - not the trigger node")
+      DetectNodeResponsiveness.start_link(tx.address, fn count ->
+        if Enum.empty?(new_oracle_data) and trigger_node?(summary_date, index + 1, count) do
```
samuel-uniris, line 189: You can reuse the emptiness check before the trigger node detection to avoid multiple calls

imnik11: OK, but generally 
` if a==b and c==d ` # second condition check happens only after first is true unlike in case of `if a==b or c==d` where both the conditions are checked. Is this not the case in elixir ?

lib/archethic/contracts/worker.ex
imnik11: shall I pass validation nodes as arg to trigger_node? fn then ?

lib/archethic/contracts/worker.ex
imnik11: As arg of the anonymous fn ? I thought it was like global var

samuel-uniris: if you look at the trigger_node? function , the argument is transaction, count but you are not passing.
in functional programming there is not global var

lib/archethic/oracle_chain/scheduler.ex
samuel-uniris: It does.
But you can simply:
- check if need to generate a new transaction (emptiness check)
- if yes
   - build tx 
   - if trigger node
        - send tx
    - if not trigger node
        - start watcher
- if no
   - skip

imnik11:
changes moved to PR #395

PR #389, opened 2022-06-20
Poll metrics Bug. Resolves #370.

### LINKED ISSUE #370: Poll metrics Error in metrics dashboard during node start
### What happened?

Screenshot 2022-06-09 125910

Maybe the metrics polling process started before the transaction_validation and it got empty so, it is unable to poll metrics

### What is expected ?

It expected to start normally and give the `archethic_mining_full_transaction_validation_duration => val `, similarly others metrics etc

apoorv-2204:
# Description

Includes acc in Enum.reduce
Fixes # (#370)

## Type of change

# How Has This Been Tested?

- [x] No error reported during /metrics/dashboard

# Checklist:


samuel-uniris (commented):
I think you can simplify to avoid creation of a module for a single function, and just create a function in `utils` to be something like: `merge_list_of_maps/1`, where you could apply doc tests as it's public function. So this could be used by other components

samuel-uniris:
> > I think you can simplify to avoid creation of a module for a single function, and just create a function in `utils` to be something like: `merge_list_of_maps/1`, where you could apply doc tests as it's public function. So this could be used by other components
> 
> I created this module , in future prospect for data mutation and to resolve to_map and from_map.

Do not overengineer and foresee too much, we might not even remember it, so you can either use private fn or use already existing module like utils. So it's simple to manage.

apoorv-2204:
> > > I think you can simplify to avoid creation of a module for a single function, and just create a function in `utils` to be something like: `merge_list_of_maps/1`, where you could apply doc tests as it's public function. So this could be used by other components
> > 
> > 
> > I created this module , in future prospect for data mutation and to resolve to_map and from_map.
> 
> Do not overengineer and foresee too much, we might not even remember it, so you can either use private fn or use already existing module like utils. So it's simple to manage.
> 
> We could refactor later in any case

Agreed Noted,
But actually, I saw we had too many methods for data mutation spread over different modules.

samuel-uniris:
> > > > I think you can simplify to avoid creation of a module for a single function, and just create a function in `utils` to be something like: `merge_list_of_maps/1`, where you could apply doc tests as it's public function. So this could be used by other components
> > > 
> > > 
> > > I created this module , in future prospect for data mutation and to resolve to_map and from_map.
> > 
> > 
> > Do not overengineer and foresee too much, we might not even remember it, so you can either use private fn or use already existing module like utils. So it's simple to manage.
> > We could refactor later in any case
> 
> Agreed Noted, But actually, I saw we had too many methods for data mutation spread over different modules.

Data mutation/transformation is the core of functional programming, so it's kind of obvious to use them in different places. The issue of many places, it's when something is really recurrent and with the same behavior. But for data transformation, often, need to be adapted for a given context, it might the interpretation or the expected result

apoorv-2204:
> > > > > I think you can simplify to avoid creation of a module for a single function, and just create a function in `utils` to be something like: `merge_list_of_maps/1`, where you could apply doc tests as it's public function. So this could be used by other components
> > > > 
> > > > 
> > > > I created this module , in future prospect for data mutation and to resolve to_map and from_map.
> > > 
> > > 
> > > Do not overengineer and foresee too much, we might not even remember it, so you can either use private fn or use already existing module like utils. So it's simple to manage.
> > > We could refactor later in any case
> > 
> > 
> > Agreed Noted, But actually, I saw we had too many methods for data mutation spread over different modules.
> 
> Data mutation/transformation is the core of functional programming, so it's kind of obvious to use them in different places. The issue of many places, it's when something is really recurrent and with the same behavior. But for data transformation, often, we need to adapt it for a given context, it might change the interpretation or the expected result

Is it a good idea to generalize from_map and to_map for structs?

Issue #303, opened 2022-05-05
Add ID to the NFT UTXO model


samuel-uniris:
### Problem to solve

In Bitcoin UTXO model, to identify which UTXO from which transaction we are using, a transaction ID and UTXO ID is used.

Also, in order to support multiple NFT UTXO and specific consumption of them, we need a way to identify the one to use.

This is particularly necessary in NFT collection minting, where we are using UTXO for each token and want to send them individually to other wallets.

### Solution

The proposal would be to add an UTXO's ID and NFT's ID in the transaction data structure to resolve this point.
So a transaction could be built to use a specific NFT UTXO to consume.

This would involve:
- [x] Update transaction data structure for NFT Ledger and UTXO
- [x] Update the ledger operations consumption algorithm

AmodD:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Product-and-Design-Workspace-61efe71b9083a90018250d89/issues/archethic-foundation/archethic-node/303?planning-poker) with ZenHub @apoorv-2204 @blackode @Neylix

AmodD:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Product-and-Design-Workspace-61efe71b9083a90018250d89/issues/archethic-foundation/archethic-node/303?planning-poker) with ZenHub @samuel-uniris

apoorv-2204:
Given a situation, 
I am trying to mint nft color , which are just 256 color from (erc721 implemented smart contract)
So One utxo ID and one NFT ID (0to255) mapping ??
instead of One utxo id including all 256.?

samuel-uniris:
> Given a situation, I am trying to mint nft color , which are just 256 color from (erc721 implemented smart contract) So One utxo ID and one NFT ID (0to255) mapping ?? instead of One utxo id including all 256.?

Yes

apoorv-2204:
https://misc.archethic.net/fLXItn9yTU2P0-pb-rT9lw

apoorv-2204:
> In Bitcoin UTXO model, to identify which UTXO from which transaction we are using, a transaction ID and UTXO ID is used.

tx_id -> txn hash, pointer to old txn whose output was this utxo
vout -> The index number of the UTXO to be spent; first one is 0

> The proposal would be to add an UTXO's ID and NFT's ID in the transaction data structure to resolve this point.

In ae -> the utxo_id will equivalent to vout(like in BTC)? 
[...]

PR #390, opened 2022-06-20
Token standard

### LINKED ISSUE #378: Token standard implementation
According to archethic-foundation/aeip#1, we now have to implement it.

The token will be minted as before, but using the rules from the standard.

- [X] Validate the JSON in the transaction's content
- [x] Parse the json and extract minting policies (supply, properties identification for the number of utxos)
- [x] For non-fungible token create multi UTXO with ID

samuel-uniris:
# Description

Implement the token standard

Fixes #378 #303

## Type of change

# How Has This Been Tested?

To create a token, we can use the libjs example transaction builder and create transaction of type 'nft' and enter your token specification, for example:
```json
{
   "supply": 100,
   "name": "My Token",
   "symbol": "MTK",
   "type": "fungible"
}
```

This should create a new transaction where the unspent outputs should be 100.
This test can also be done with fungible tokens:
```json
{
  "supply": 3,
  "name": "MyToken",
  "symbol": "MTK",
  "type": "non-fungible",
  "properties": [
      [{ "name": "image", "value": "mylink"}],
      [{ "name": "image", "value": "mylink"}],
      [{ "name": "image", "value": "mylink"}]
   ]
}
```

This PR also implement a fee for the NFT creation and minting being logarithmic, so token with high supply will cost more than other with small supply as the UTXO must be created and will cost space and bandwidth.
But providing a fair transaction's fee compared to UCO transfers.

# Checklist:


lib/archethic/mining/fee.ex
```diff
@@ -53,10 +53,26 @@ defmodule Archethic.Mining.Fee do
...
   end
 
+  defp get_additional_fee(
+         %Transaction{type: :nft, data: %TransactionData{content: content}},
+         uco_price_in_usd
+       ) do
+    with {:ok, json} <- Jason.decode(content),
+         supply when is_integer(supply) and supply > 0 <- Map.get(json, "supply", 0) do
+      base_fee = minimum_fee(uco_price_in_usd)
+      trunc(:math.log10(supply) + 1 * supply * base_fee)
```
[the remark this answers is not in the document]
samuel-uniris: Good point for the length of properties, instead of the supply, for fungible tokens this make sense

lib/archethic/mining/fee.ex
```diff
@@ -53,10 +53,26 @@ defmodule Archethic.Mining.Fee do
             nb_storage_nodes,
             nb_recipients
           ) * @unit_uco
-        )
+        ) + get_additional_fee(tx, uco_price_in_usd)
```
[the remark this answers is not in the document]
samuel-uniris: Right

lib/archethic/mining/pending_transaction_validation.ex
```diff
@@ -307,10 +307,23 @@ defmodule Archethic.Mining.PendingTransactionValidation do
...
+      |> File.read!()
+      |> Jason.decode!()
+      |> ExJsonSchema.Schema.resolve()
+
+    with {:ok, json_token} <- Jason.decode(content),
+         :ok <- ExJsonSchema.Validator.validate(schema, json_token) do
       :ok
     else
-      {:error, "Invalid NFT content"}
+      :error ->
```
Neylix, line 321: To catch `Jason.decode()` error you should use `{:error, %Jason.DecodeError{}}`

samuel-uniris: 👍

lib/archethic/p2p/message.ex
```diff
@@ -429,8 +429,8 @@ defmodule Archethic.P2P.Message do
   def encode(%Balance{uco: uco_balance, nft: nft_balances}) do
     nft_balances_binary =
       nft_balances
-      |> Enum.reduce([], fn {nft_address, amount}, acc ->
-        [<<nft_address::binary, amount::float>> | acc]
+      |> Enum.reduce([], fn {{nft_address, nft_id}, amount}, acc ->
+        [<<nft_address::binary, amount::float, nft_id::8>> | acc]
```
prix-uniris, line 433: NFT Id is 8 bits so, it's range will be between 0-255? or is it 8 bytes?

lib/archethic/mining/pending_transaction_validation.ex
```diff
@@ -307,10 +307,23 @@ defmodule Archethic.Mining.PendingTransactionValidation do
...
-    if Regex.match?(~r/(?<=initial supply:).*\d/mi, content) do
+    schema =
+      :archethic
+      |> Application.app_dir("priv/json-schemas/token-core.json")
+      |> File.read!()
+      |> Jason.decode!()
+      |> ExJsonSchema.Schema.resolve()
+
+    with {:ok, json_token} <- Jason.decode(content),
+         :ok <- ExJsonSchema.Validator.validate(schema, json_token) do
```
Neylix, line 318: You should add a verification to attest that if the properties list have more than 1 item, the supply must be the same than the properties list length

lib/archethic_web/controllers/api/schema/nft_ledger.ex
```diff
@@ -26,8 +27,10 @@ defmodule ArchethicWeb.API.Schema.NFTLedger do
 
   defp changeset_transfers(changeset, params) do
     changeset
-    |> cast(params, [:to, :amount, :nft])
-    |> validate_required([:to, :amount, :nft])
+    |> cast(params, [:to, :amount, :nft, :nft_id])
+    |> validate_required([:to, :amount, :nft, :nft_id])
     |> validate_number(:amount, greater_than: 0)
+    |> validate_number(:nft_id, greater_than_or_equal_to: 0)
+    |> validate_number(:nft_id, less_than_or_equal_to: 255)
```
prix-uniris, line 34: validate inclusion would be better option to check value in range 
validate_inclusion(changeset, :nft_id, 0..255)
https://hexdocs.pm/ecto/Ecto.Changeset.html#validate_inclusion/4

prix-uniris (commented):
Added My feedback/doubts

lib/archethic/p2p/message.ex
samuel-uniris: This is erlang bitstring notation.
So you can define how bits or binary are encoded.
Here it means the id is encoded in 8 bits

priv/json-schemas/token-core.json
```diff
@@ -0,0 +1,51 @@
...
+       "type": "string",
+       "description": "Symbol of the token"
+     },
+     "properties": {
+        "description": "List of all the token properties (each one being for a token item)",
+        "type": "array",
+        "items": { "$ref": "#/$defs/property-list" }
+      }
+   },
+   "required": [ "supply", "type" ],
```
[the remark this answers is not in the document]
samuel-uniris: We thought about it, because token can exist without name or symbol and don't impact how the nodes are interpreting the specification.

Neylix (approved):

lib/archethic/p2p/message.ex
prix-uniris: So, In a NFT can have maximum of 256 properties,  aka NFT Tokens, if taken sequentially 0..255. If It's a hard restriction that is not mentioned in aeip #2 then we should mention it there as well.

prix-uniris (approved):
LGTM

lib/archethic/p2p/message.ex
Neylix: Yes good observation !
Actually the nft_id is limited to 256 values because we can have a maximum of 256 inputs in a transaction.
But we wanted to created a new issue to make the number of inputs dynamic ot have un unlimited number of input (or at least a huge number)

prix-uniris: Screenshot from 2022-06-23 00-09-55
Here Id: 2022 can't fit in 8bits so, we can change this in our aeip too. 
And I think there are some project who have 6000+ NFT tokens for a NFT, so having a bigger range would be definitely better, and yes with an encoding scheme mentioning a length and after that, all properties would make it huge. 
Got your point also with the maximum of 256 Input transactions, thanks for explaining.

lib/archethic/mining/pending_transaction_validation.ex
```diff
@@ -307,10 +307,29 @@ defmodule Archethic.Mining.PendingTransactionValidation do
...
+      :archethic
+      |> Application.app_dir("priv/json-schemas/token-core.json")
+      |> File.read!()
+      |> Jason.decode!()
+      |> ExJsonSchema.Schema.resolve()
+
+    with {:ok, json_token} <- Jason.decode(content),
+         :ok <- ExJsonSchema.Validator.validate(schema, json_token),
+         %{"type" => "non-fungible", "supply" => supply, "properties" => properties}
+         when length(properties) == supply <- json_token do
```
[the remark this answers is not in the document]
Neylix: Also, as the token id is use and makes sense for token and nft, I think we should rename id to `id` instead of `nft_id`

lib/archethic/transaction_chain/transaction/validation_stamp/ledger_operations.ex
```diff
@@ -64,24 +84,42 @@ defmodule Archethic.TransactionChain.Transaction.ValidationStamp.LedgerOperation
...
+
+  defp get_token_utxos(
+         %{"type" => "non-fungible", "supply" => supply, "properties" => properties},
+         address
+       )
+       when length(properties) == supply do
+    properties
+    |> Enum.with_index()
+    |> Enum.map(fn {_item_properties, index} ->
+      %UnspentOutput{from: address, amount: 1 * @unit_uco, type: {:NFT, address, index + 1}}
```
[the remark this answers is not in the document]
samuel-uniris: Yes right I missed this one 👍

lib/archethic/mining/pending_transaction_validation.ex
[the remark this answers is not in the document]
samuel-uniris: Ok I will add the check

PR #393, opened 2022-06-21
Initialize miner reward pool

### LINKED ISSUE #379: Create miner reward pool during network initialization

No related to problem but to support the new miner reward mechanism we have dedicate a transaction chain with for the bootstrap funds of rewards.

During the network initialization we have to create a transaction chain to allocate the amount for the bootstrap of the reward tokens.
So, only the first node which initializes the network will do it.
The chain should use a seed (randomly generated) which will be encoded and encrypted in the node shared secrets (replacement of the `network_seed`).
Hence authorized nodes will be able to build transactions to send rewards to the nodes.

During this first transaction, the token will minted (as token) with the initial supply as 1000000

### Additional context

_No response_

Neylix:
# Description

This PR implement the initialization of the minser reward pool.
It uses the network pool seed of the shared secret chain to create a new transaction chain.
The first transaction is the creation of 1M NFT

Fixes #379 

## Type of change

# How Has This Been Tested?

Run the node and there is the creation of the transaction during network init.
Also uses test files

# Checklist:

### DIFF
--- a/config/test.exs
+++ b/config/test.exs
@@ -6,7 +6,7 @@ config :logger, level: :error
 config :archethic, :mut_dir, "data_test"
 
 config :archethic, Archethic.Account.MemTablesLoader, enabled: false
-config :archethic, Archethic.Account.MemTables.NFTLedger, enabled: false
+config :archethic, Archethic.Account.MemTables.TokenLedger, enabled: false
 config :archethic, Archethic.Account.MemTables.UCOLedger, enabled: false
 
 config :archethic, Archethic.BeaconChain.Subset, enabled: false


lib/archethic/bootstrap/network_init.ex
```diff
@@ -139,6 +141,29 @@ defmodule Archethic.Bootstrap.NetworkInit do
...
+    }
+
+    tx = %Transaction{address: address} = Transaction.new_nft_reward(data)
+
+    tx
+    |> self_validation([
+      %UnspentOutput{
+        from: address,
+        amount: @genesis_network_pool_amount,
+        type: {:NFT, address}
```
apoorv-2204: Are reward tokens are non-fungible NFT's?
If they are fungible , we might need to have nft_id 0. as per aeip2

Neylix: This is an old commit, go on the last to see the new token standard implementation https://github.com/archethic-foundation/archethic-node/pull/393/commits/c9c195a45d0c894b63108318b8bd9764c92388bf

samuel-uniris (approved):

Issue #282, opened 2022-04-19
OracleChain DB File non-existing 


samuel-uniris:
### What happened?

<img width="1247" alt="image" src="https://user-images.githubusercontent.com/42943690/163973781-2f0967c0-51aa-4e5a-b865-c6e3ddeb48d5.png">

After few minutes of running , the oracle chain crashes because the transaction cannot be read from the disk.
The index lookup mentioned the existing of the genesis address, but the file was not present, which is causing the error

### What is expected ?

A transaction's chain file should be readable if the index mention its existence.

apoorv-2204:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Product-and-Design-Workspace-61efe71b9083a90018250d89/issues/archethic-foundation/archethic-node/282?planning-poker) with ZenHub @samuel-uniris @imnik11 @blackode

samuel-uniris:
I have noticied the chain folder to disappear but hard to reproduce

apoorv-2204:
Task.
- [ ] Gather knowledge  
- [ ] Create Subtasks
- [ ] Pinpoint the Bug
- [ ] Propose Solution

apoorv-2204:
Update: 
I was able to replicate the issue but not deterministically.
I generated thousands of txns , in mix archethic. Regression --bench localhost  
But an another error was that ets tables were not loading up.
 
we may change the _build/devorprod/archethic/datanode1 to 
`~/aebot/aenode/datenode

apoorv-2204:
I think we can close this , issue. It is already resolved.

It was a issue of writing to same genesis address file concurrently. 
:file erlang does not write to same file in mutual exclusion. So a 
a queue was created to serialise writing to same genesis address file.
While we write concurrently to different files,

PR #395, opened 2022-06-23
Detection of unresponsive trigger node Contd of 376 

### LINKED ISSUE #338: Detect unresponsive trigger node

imnik11:
# Description
It detects the unresponsive nodes and resends the transaction to the next trigger next node.

Continuation of PR #376 
Fixes #338 

## Type of change

# How Has This Been Tested?

Test with 4 three nodes
`>ARCHETHIC_CRYPTO_SEED=node1 ARCHETHIC_P2P_PORT=3002 ARCHETHIC_HTTP_PORT=4000 iex -S mix`
`>ARCHETHIC_CRYPTO_SEED=node2 ARCHETHIC_P2P_PORT=3003 ARCHETHIC_HTTP_PORT=4001 iex -S mix`
`>ARCHETHIC_CRYPTO_SEED=node3 ARCHETHIC_P2P_PORT=3004 ARCHETHIC_HTTP_PORT=4002 iex -S mix`
`>ARCHETHIC_CRYPTO_SEED=node4 ARCHETHIC_P2P_PORT=3005 ARCHETHIC_HTTP_PORT=4003 iex -S mix`

If we stop the node1 the node shared secret transactions , oracle etc must be fall back to node2, node2 as trigger node and shall be validated.

But for me if a Node is removed the Nodes raised No cross validator node available error and let to #282 kind of error after sometime, otherwise the if the oracle polling transaction was not yet confirmed then other nodes were detecting and resent the tx.
# Checklist:

### DIFF
--- a/lib/archethic/oracle_chain/scheduler.ex
+++ b/lib/archethic/oracle_chain/scheduler.ex
@@ -149,7 +149,7 @@ defmodule Archethic.OracleChain.Scheduler do
 
     storage_nodes =
       summary_date
-      |> Crypto.derive_oracle_address(index)
+      |> Crypto.derive_oracle_address(index + 1)
       |> Election.storage_nodes(authorized_nodes)
 
     tx = build_oracle_transaction(summary_date, index, new_oracle_data)
@@ -168,13 +168,15 @@ defmodule Archethic.OracleChain.Scheduler do
         {:trigger, false} ->
           {:ok, pid} =
             DetectNodeResponsiveness.start_link(tx.address, fn count ->
-              if trigger_node?(storage_nodes, count) do
+              new_oracle_data = get_new_oracle_data(summary_date, index)
+              new_data? = !Enum.empty?(new_oracle_data)
+
+              if trigger_node?(storage_nodes, count) and new_data? do
                 Logger.info("Oracle polling transaction ...attempt #{count}",
                   transaction_address: Base.encode16(tx.address),
                   transaction_type: :oracle
                 )
 
-                new_oracle_data = get_new_oracle_data(summary_date, index)
                 tx = build_oracle_transaction(summary_date, index, new_oracle_data)
                 send_polling_transaction(tx)
               end
--- a/lib/archethic/oracle_chain/scheduler.ex
+++ b/lib/archethic/oracle_chain/scheduler.ex
@@ -173,6 +173,7 @@ defmodule Archethic.OracleChain.Scheduler do
                   transaction_address: Base.encode16(tx.address),
                   transaction_type: :oracle
                 )
+
                 new_oracle_data = get_new_oracle_data(summary_date, index)
                 tx = build_oracle_transaction(summary_date, index, new_oracle_data)
                 send_polling_transaction(tx)
@@ -606,7 +607,6 @@ defmodule Archethic.OracleChain.Scheduler do
   end
 
   defp get_new_oracle_data(summary_date, index) do
-
     summary_date
     |> Crypto.derive_oracle_address(index)
     |> get_oracle_data()
--- a/lib/archethic/oracle_chain/scheduler.ex
+++ b/lib/archethic/oracle_chain/scheduler.ex
@@ -146,60 +146,52 @@ defmodule Archethic.OracleChain.Scheduler do
       |> get_oracle_data()
       |> Services.fetch_new_data()
 
-    if generate_oracle_tx?(new_oracle_data) do
-      # build transaction
-      {prev_pub, prev_pv} = Crypto.derive_oracle_keypair(summary_date, index)
+    authorized_nodes =
+      summary_date
+      |> P2P.authorized_nodes()
+      |> Enum.filter(& &1.available?)
 
-      {next_pub, _} = Crypto.derive_oracle_keypair(summary_date, index + 1)
+    storage_nodes =
+      summary_date
+      |> Crypto.derive_oracle_address(index)
+      |> Election.storage_nodes(authorized_nodes)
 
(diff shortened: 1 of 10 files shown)


imnik11:
Screenshot 2022-06-23 180617

Here if u look the tx at 12:34:13 is getting validation at first attempt but the other one 12:34:23 is going for second retry and is being getting validated at 12:34:26 so, if we have long time retrying(>10 sec) it will give invalid tx based on validation_context.ex

samuel-uniris:
> Screenshot 2022-06-23 224622 Now I change the timeout to @default_timeout to 11 * 1000 so, it will be 11 sec and tried to simulate what happens during long timeouts but I encountered this unwanted error in the log which says that No cross validation node but the transaction is already done.

Probably ~10s is better because the default timeouts during mining are:
- coordinator notification (validation stamp) -> 5s
- cross validation nodes (add context) -> 3s

imnik11:
Screenshot 2022-06-24 112925

This is strange I am able to get the transaction on the explorer but the log reports, no cross validation node respond to confirm mining context.

Also the address is for oracle summary transaction type whereas I found the same address for oracle tx.

test/archethic/utils/detect_node_responsiveness_test.exs
```diff
@@ -0,0 +1,290 @@
...
+  end
+
+  test "start_link/2 for hard timeout state" do
+    address = <<0::8, 0::8, :crypto.strong_rand_bytes(32)::binary>>
+    # dummy replaying fn with count as argument
+    replaying_fn = fn count ->
+      count
+    end
+
+    {:ok, pid} = DetectNodeResponsiveness.start_link(address, replaying_fn)
```
samuel-uniris, line 29: I think you should make the timeouts configurable, as 10s for all tests is too long in a unit testing suite.

test/archethic/utils/detect_node_responsiveness_test.exs
```diff
@@ -0,0 +1,290 @@
...
+
+    MockDB
+    |> stub(:transaction_exists?, fn ^address ->
+      false
+    end)
+
+    #  first soft_timeout
+    assert true == Process.alive?(pid)
+    Process.sleep(10_500)
```
samuel-uniris, line 108: I don't think you need to sleep as you are expecting to receive the message before to continue

samuel-uniris (changes requested):

lib/archethic/oracle_chain/scheduler.ex
```diff
@@ -140,21 +140,58 @@ defmodule Archethic.OracleChain.Scheduler do
...
+        {:trigger, false} ->
+          {:ok, pid} =
+            DetectNodeResponsiveness.start_link(tx.address, fn count ->
+              if trigger_node?(storage_nodes, count) do
+                Logger.info("Oracle polling transaction ...attempt #{count}",
+                  transaction_address: Base.encode16(tx.address),
+                  transaction_type: :oracle
+                )
+
+                send_polling_transaction(tx)
```
samuel-uniris, line 181: @imnik11 to avoid consensus not reached as the data might change, I think you should remake the data polling from the service providers

lib/archethic/shared_secrets/node_renewal_scheduler.ex
```diff
@@ -117,9 +119,26 @@ defmodule Archethic.SharedSecrets.NodeRenewalScheduler do
...
+        :crypto.strong_rand_bytes(32),
+        :crypto.strong_rand_bytes(32)
+      )
+
+    validation_nodes = get_validation_nodes()
+
+    if trigger_node?(validation_nodes) do
       Logger.info("Node shared secrets renewal creation...")
-      make_renewal()
+      make_renewal(tx)
```
Neylix, line 133: I think you can pass the `validation_nodes` in parameter as so the `Archethic.send_new_transaction` will not recalculate the validation nodes

Neylix (changes requested):

lib/archethic/shared_secrets/node_renewal_scheduler.ex
imnik11: ok, will do that anything else

Neylix (approved):

PR #402, opened 2022-06-24
Mint mining rewards token on self repair

### LINKED ISSUE #380: Mint miner rewards token on self-repair cycle

This issue is to provide a way to use the transaction fees burnt and allocate fill the miner reward pool from them.
This should prevent the drain of the miner reward pool and regulate its amount with the fees already paid.

During the self-repair, we already fetching beacon chain summaries to identify the transactions to synchronize.
All the transaction summaries contain the transaction's fee used during validation.
Then we should:
- aggregate the amount of fee burnt during this cycle
- build transaction to mint new tokens on the miner reward chain
- this transaction should be triggered by a given node (and use the trigger node detection to prevent failure)

### Additional context

This is related to #379

Neylix:
# Description

This PR create a new scheduler for minting NFT rewards.
During self repair, all nodes store calculate and store the amount of burned fees during the past summary.
Then the scheduler trigger a node to create a mint rewards transaction.
The transaction is created with the amount of burned fees stored.
To avoid any corruption, we control during mining that the supply match the last burned feed amount and that there is no mint reward transaction since last schedule

Fixes #380 

## Type of change

# How Has This Been Tested?

Tested on mutliple node to match the election.
Every minute at 30 sec a new transaction is generated with the good supply amount.
If trying to create a malicious transaction using the network pool seed, it returns an error

Add these test in test files

# Checklist:

### DIFF
--- a/lib/archethic/reward.ex
+++ b/lib/archethic/reward.ex
@@ -5,6 +5,13 @@ defmodule Archethic.Reward do
 
   alias Archethic.OracleChain
 
+  alias Archethic.Crypto
+
+  alias Archethic.Election
+
+  alias Archethic.P2P
+  alias Archethic.P2P.Node
+
   alias __MODULE__.NetworkPoolScheduler
 
   alias Archethic.TransactionChain.Transaction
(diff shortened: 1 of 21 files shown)


lib/archethic/reward.ex
```diff
@@ -28,6 +35,21 @@ defmodule Archethic.Reward do
 
   @doc """
   Create a transaction for minting new rewards
+
+  ## Examples
+
+    iex> case Reward.new_rewards_mint(2_000_000_000) do
```
samuel-uniris, line 41: In the doctest you directly put the expected result, so no need to make the check if's working on not in

```elixir
iex> Reward.new_reward_mint(2_000_000_000)
%Transaction{ 
  type: :mint_rewards,
  data: %TransactionData{
      content: "..."
  }
}
```

lib/archethic/reward/reward_scheduler.ex
```diff
@@ -1,4 +1,4 @@
-defmodule Archethic.Reward.NetworkPoolScheduler do
+defmodule Archethic.Reward.RewardScheduler do
```
samuel-uniris, line 1: I think you can rename it Scheduler, to avoid name repetition

PR #418, opened 2022-06-28
Fix db timeouts

### LINKED ISSUE #416: Database write are causing timeout

After running a node few hours, I faced this issue related to database.

image

It seems the writing didn't reach confirmation before 5s, caused probably by some congestion in beacon chain.

Right now the chain writer is using `GenServer.call` waiting complete storage before given acknowledgement. 

We have few approaches to tackle this issue:
- using `GenServer.cast` would avoid timeouts, but without the acknowledgement of the writes
- using `Task.async` inside the GenServer `handle_call` to avoid blocking calls - even if some call would have timeout, others will not be impacted
- using backpressure - we can prevent overloading of the writes and manage by the demand. But the P2P messages might be delayed.

samuel-uniris:
# Description

Fixes #416 

## Type of change

# How Has This Been Tested?

To validate this, we should try to simulate load and run for a certain amount of time

# Checklist:

### DIFF
--- a/lib/archethic/db/embedded_impl/chain_writer.ex
+++ b/lib/archethic/db/embedded_impl/chain_writer.ex
@@ -34,7 +34,7 @@ defmodule Archethic.DB.EmbeddedImpl.ChainWriter do
     db_path = Keyword.get(arg, :path)
     setup_folders(db_path)
 
-    {:ok, %{db_path: db_path}}
+    {:ok, %{db_path: db_path, clients: %{}}}
   end
 
   defp setup_folders(path) do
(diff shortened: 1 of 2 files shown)


lib/archethic/db/embedded_impl/chain_writer.ex
```diff
@@ -45,9 +45,29 @@ defmodule Archethic.DB.EmbeddedImpl.ChainWriter do
...
         {:append_tx, genesis_address, tx},
-        _from,
+        from,
         state = %{db_path: db_path}
       ) do
+    %Task{ref: ref} = Task.async(fn -> write_transaction(genesis_address, tx, db_path) end)
+    {:noreply, Map.update!(state, :clients, &Map.put(&1, ref, from))}
+  end
+
+  def handle_info({ref, :ok}, state) do
```
apoorv-2204, line 55: This is a pretty good Idea.

lib/archethic/db/embedded_impl/chain_writer.ex
```diff
@@ -45,9 +45,29 @@ defmodule Archethic.DB.EmbeddedImpl.ChainWriter do
...
+    case pop_in(state, [:clients, ref]) do
+      {nil, ^state} ->
+        {:noreply, state}
+
+      {from, new_state} ->
+        GenServer.reply(from, :ok)
```
apoorv-2204, line 61: I think, we should add  test cases for it.

samuel-uniris: If the existing test cases are working, it mean we didn't break something. 
We could add more load testing

prix-uniris (approved):

PR #432, opened 2022-07-06
Fixed multiple pub sub causing crash due to new replication attestations

### LINKED ISSUE #429: Multiple pubsub subscriber on beacon explorer

I launched a node and went on beacon chain explorer page and stayed on it for a while.
The error hapened is that sometimes the pages is refreshed and so a new pubsub for replication attestationis created, but the previous pubsub is not removed.
So when a new replication attestation is sent, there is mutiple beacon explorer that handle it and after a while there is this kind of error : 

image

With all the subsribers: 

image

Beacon explorer should not subscribe to a PubSub if it is already subscribed.

imnik11:
# Description
we were using same pub sub for replication attestation during transaction validation which caused crash due to beacon explorer was also using the same. Now we added a separate PubSub for beacon explorer to get the transaction attestations and push it to UI 

Fixes #429 

## Type of change

# How Has This Been Tested?

Run 3 nodes and checked the logs, Now I don't get a lot of logger msg showing same info. And the transaction duplication is also fixed in beacon explorer

# Checklist:

### DIFF
--- a/lib/archethic/beacon_chain/subset.ex
+++ b/lib/archethic/beacon_chain/subset.ex
@@ -193,11 +193,13 @@ defmodule Archethic.BeaconChain.Subset do
     end
   end
 
-  defp notify_subscribed_nodes(nodes, attestation) do
+  defp notify_subscribed_nodes(nodes, %ReplicationAttestation{
+         transaction_summary: tx_summary
+       }) do
     nodes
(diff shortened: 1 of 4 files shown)


Neylix (approved):

Neylix:
Nice catch !

Issue #425, opened 2022-07-01
Self-Repair timeouts with the new batching of summaries


samuel-uniris:
### Describe the problem you discovered

![image](https://user-images.githubusercontent.com/42943690/176845717-64257f1f-e410-441e-a436-018445332e50.png)

This issue is a critical one making the self-repair not working, causing the network to be in unstable state.
Without self-repair working, the entire synchronization is not working, resulting in no transaction validation at the end (cf node shared secrets)

### Describe the solution you'd like

We should support batch of summaries and avoid timeouts regarding the data within.
We have to check if the issue is coming from the read of the data on the beacon chain or due to the amount of data to return.

samuel-uniris:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Workspace-61efec9c93391c001e69a6dc/issues/archethic-foundation/archethic-node/425?planning-poker) with ZenHub @Neylix @prix-uniris @apoorv-2204 @imnik11

samuel-uniris:
You can run benchmarks trying to reproduce to leverage many transactions

PR #441, opened 2022-07-11
Improve P2P

Neylix:
# Description

During mining process, there is several request to get a previous transaction or a the while tranaction chain.
Request and mining process have some timeout of 3sec.
But transactions can have a maximum size of 3MB and if the node connection is low (especially in upload), send 3MB can take more time than 3 sec

So to avoid te timeout behavior for low connection speed, we can dynamically adapt the timeout depending on the maximum request size expected.

Fixes poart of #425 

## Type of change

# How Has This Been Tested?

To simulate low connection, we can update the connection.ex file with this on line 299: 
```
time = trunc(byte_size(msg) / @floor_upload_speed * 1_000)
Process.send_after(from, {ref, {:ok, message}}, time)
# send(from, {ref, {:ok, message}})
```
and this in top of the file
```
@floor_upload_speed Application.compile_env!(:archethic, [Message, :floor_upload_speed])
```

To test with big transaction size we can use Aeweb cli

Transaction will take a lot of time to be validated, but will not fall in timeout due to low speed connection

# Checklist:

### DIFF
--- a/lib/archethic/mining/distributed_workflow.ex
+++ b/lib/archethic/mining/distributed_workflow.ex
@@ -26,6 +26,7 @@ defmodule Archethic.Mining.DistributedWorkflow do
   alias Archethic.Mining.WorkflowRegistry
 
   alias Archethic.P2P
+  alias Archethic.P2P.Message
   alias Archethic.P2P.Message.AcknowledgeStorage
   alias Archethic.P2P.Message.AddMiningContext
   alias Archethic.P2P.Message.CrossValidate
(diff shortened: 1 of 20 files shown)


lib/archethic/p2p/message.ex
```diff
@@ -150,6 +152,28 @@ defmodule Archethic.P2P.Message do
...
+      3_000
+    end
+  end
+
+  @doc """
+  Return the maximum timeout for a full sized transaction
+  """
+  @spec get_max_timeout() :: non_neg_integer()
+  def get_max_timeout() do
+    trunc(@content_max_size / @floor_upload_speed * 1_000)
```
apoorv-2204, line 174: I think it means  trunc(
(@content_max_size / (@floor_upload_speed * 1_000))
)

Neylix: No it's `trunc((@content_max_size / @floor_upload_speed) * 1_000)`
but formater remove parenthesis as they are not needed in this case

lib/archethic/mining/distributed_workflow.ex
```diff
@@ -126,6 +127,17 @@ defmodule Archethic.Mining.DistributedWorkflow do
     GenStateMachine.cast(pid, {:add_cross_validation_stamp, stamp})
   end
 
+  defp get_context_timeout(type) when type == :hosting, do: Message.get_max_timeout()
```
apoorv-2204, line 130: `function(:timeout)`
`function(:oracle)`
Hello, Does this pattern matching works?

apoorv-2204 (approved):

Issue #443, opened 2022-07-12
Replicate only missing transactions on Replication transaction chain


Neylix:
### Describe the problem you discovered

Whenever we send the message `GetTransactionChain` we request all the chain and depending on the transactions size it can take a long time.

### Describe the solution you'd like

Before requesting the chain, we should know if we already have some transactions of this chain already stored in the node DB.
For this, before sending the message, we can ask to the storage node what is the genesis address of this chain.
With the genesis address, we can get the latest tranaction in the node DB and passing it as paging address in the message and so we will get only the missing transactions.

An other improvement would be to request some part of the chain to multiple node concurrently. For exemple if we miss 20 transactions, we can request 10 to node1 and 10 to node2 at the same time and then concatenate the results.

apoorv-2204:
> `we can request 10 to node1 and 10 to node2 at the same time and then concatenate the results.`
But it will require to have the 10th txn address as paging address for the 11th to 20th transactions.

internet-zero:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Product-and-Design-Workspace-61efe71b9083a90018250d89/issues/archethic-foundation/archethic-node/443?planning-poker) with ZenHub @apoorv-2204 @imnik11 @Neylix @prix-uniris

PR #453, opened 2022-07-17
Replicate only missing transactions #443

apoorv-2204:
# Description

Fixes # (#443)

## Type of change

- 
# How Has This Been Tested?

# Checklist:

### DIFF
--- a/lib/archethic.ex
+++ b/lib/archethic.ex
@@ -12,6 +12,8 @@ defmodule Archethic do
 
   alias __MODULE__.P2P
 
+  alias __MODULE__.DB
+
   alias __MODULE__.P2P.Message.Balance
   alias __MODULE__.P2P.Message.GetBalance
   alias __MODULE__.P2P.Message.NewTransaction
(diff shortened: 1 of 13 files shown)


lib/archethic.ex
```diff
@@ -137,6 +137,35 @@ defmodule Archethic do
     TransactionChain.resolve_last_address(address)
   end
 
+  @doc """
+  Retrieve the last transaction address for a chain locally
+  It queries the the netowork for genesis address
+  """
+  @spec get_last_transaction_address_locally(address :: binary()) ::
+          {:ok, binary()} | {:error, :not_found} | {:error, :network_issue}
+  def get_last_transaction_address_locally(address) when is_binary(address) do
```
Neylix, line 146: This function may be in TransactionChain module, same for `get_genesis_transaction_address`, `fetch_transaction_chain_locally` and  `available_loccaly?` as we request an information of a transaction chain

lib/archethic.ex
```diff
@@ -176,11 +205,23 @@ defmodule Archethic do
     TransactionChain.fetch_inputs_remotely(address, nodes, DateTime.utc_now())
   end
 
+  @spec available_locally?(binary) :: {boolean(), binary | nil}
+  def available_locally?(address) when is_binary(address) do
+    case get_last_transaction_address_locally(address) do
+      {:ok, last_address} -> {true, last_address}
+      {:error, :not_found} -> {false, nil}
```
Neylix, line 212: You should also catch `{:error, :network_issue}`
Also in case of wrong result, you can juste return false, nil is not used after

lib/archethic.ex
```diff
@@ -201,6 +242,43 @@ defmodule Archethic do
...
+    {:ok, txn_list}
+  end
+
+  defp fetch_transaction_chain_locally(paging_address) do
+    try do
+      nodes = [P2P.get_node_info()]
+
+      chain_page =
+        paging_address
+        |> TransactionChain.stream_remotely(nodes)
```
Neylix, line 272: This function aims to get the local transaction chain, but you get it remotely so it's not the local chain.
You should use `TransactionChain.get`

lib/archethic/transaction_chain.ex
```diff
@@ -506,6 +510,41 @@ defmodule Archethic.TransactionChain do
     |> Enum.map(fn {:ok, res} -> res end)
   end
 
+  @doc """
+  Retrieve the genesis address of a chain
+  """
+  @spec resolve_last_address(binary()) :: {:ok, binary()} | {:error, :network_issue}
```
Neylix, line 516: Spec definition so not martch the function name.
Same for `fetch_genesis_address_remotely`

Neylix (changes requested):
The process to get the last local transaction address should also be done in the replication process. Maybe in `TransactionContext.stream_transaction_chain`

lib/archethic.ex
apoorv-2204: I am querying the same nodes via P2P.message abstraction. 
It twill be more secure rather than directly querying the DB.

lib/archethic.ex
apoorv-2204: right

Neylix (approved):

PR #462, opened 2022-07-19
Fix self repair error

### LINKED ISSUE #460: Node shared secrets chain write in loop during self repair

When a node disconnect for a long time and reconnect, it launches the self repair. But if the node was previously authorized it still receive the network transaction (oracle, node shared secrets ...).
There is a problem while the self repair is not finished, when the node receive a node_shared_secrets, it write all the past transaction again and again at each node_shared_secrets transaction

When the first node_shared_secrets transaction is received, all past transaction of the chains should be written in DB, but only once

Neylix:
# Description

This PR fixes 2 errors : 
- During replication, when retrieving the previous chain, Embedded_impl.write_transaction_chain is called, but this function do not retrieve the genesis chain if it exist to append the new transactions to the same file. And so mutuiple file was created for the same chain.
- If an authorized node disconnect and is globally unavailable, of it reconnect at the end of a cycle, it will send the en of node sync flag, but it will stay unavailable because it has not enough connection time in the cycle to be globally available. And so the synced flag stay to false.

Correction: 
- Retrieve the previous address to get the genesis addres if the chain
- Send the end of node sync flag until the node is available

Fixes #460 

## Type of change

# How Has This Been Tested?

- Run 2 nodes and shut down one after it gets authorized and let the other run for a while. Then reconnect the other node and the first node shared secrets transaction should retrieve the while chain and write it in the same file and not in a new file each 10 transactions
- Run 2 nodes and shut down one after it gets authorized. Wait the node goes globally unavailable and reconnect it just before a summary time, the end of node sync should be sent on the next self repair

# Checklist:

### DIFF
--- a/lib/archethic/application.ex
+++ b/lib/archethic/application.ex
@@ -26,6 +26,7 @@ defmodule Archethic.Application do
   alias Archethic.Networking.Supervisor, as: NetworkingSupervisor
 
   alias Archethic.P2P.Supervisor, as: P2PSupervisor
+  alias Archethic.P2P.ListenerSupervisor
 
   alias Archethic.OracleChain
   alias Archethic.OracleChain.Supervisor, as: OracleChainSupervisor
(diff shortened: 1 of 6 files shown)


imnik11 (approved):
I checked with graphql 
With 5 nodes
In earlier codebase 
Screenshot 2022-07-21 134218
Due to quorum read it takes the other nodes

Screenshot 2022-07-21 140544

after this PR

Screenshot 2022-07-21 141145

PR #482, opened 2022-08-01
#447

### LINKED ISSUE #447: Faucet - Provide control over the transaction chain and all nodes

Actually, if a user uses the faucet for an address, he couldn't receive more than 3 transactions in an hour with a node.
We should extend this control for the addresses of the transaction chain et for all nodes

_No response_

### Additional context

_No response_

prix-uniris:
# Description
This changes in following details:
- Make use of TransactionChain.fetch_genesis_address_remotely(address) in faucet rate limiter
- Triggering FaucetRateLimiter's register after transaction has been replicated and triggered from transaction_subscriber.ex

Fixes #447

## Type of change

# How Has This Been Tested?

Manual testing with single node

# Checklist:

### DIFF
--- a/lib/archethic/transaction_chain.ex
+++ b/lib/archethic/transaction_chain.ex
@@ -836,7 +836,7 @@ defmodule Archethic.TransactionChain do
     end
   end
 
-  defp fetch_genesis_address_remotely(address) when is_binary(address) do
+  def fetch_genesis_address_remotely(address) when is_binary(address) do
     nodes =
       address
       |> Election.chain_storage_nodes(P2P.available_nodes())
--- a/lib/archethic/transaction_chain.ex
+++ b/lib/archethic/transaction_chain.ex
@@ -836,6 +836,12 @@ defmodule Archethic.TransactionChain do
     end
   end
 
+  @doc """
+  Retrieve the genesis address for a chain from P2P Quorom
+  It queries the the network for genesis address
+  """
+  @spec fetch_genesis_address_remotely(address :: binary()) ::
+          {:ok, binary()} | {:error, :network_issue}
   def fetch_genesis_address_remotely(address) when is_binary(address) do
     nodes =
       address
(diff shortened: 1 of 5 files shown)


lib/archethic/transaction_chain.ex
```diff
@@ -836,7 +836,7 @@ defmodule Archethic.TransactionChain do
     end
   end
 
-  defp fetch_genesis_address_remotely(address) when is_binary(address) do
+  def fetch_genesis_address_remotely(address) when is_binary(address) do
```
samuel-uniris, line 839: Can you add some doc and typespec as well (as it's public now) ?

lib/archethic_web/faucet_rate_limiter.ex
```diff
@@ -46,14 +48,23 @@ defmodule ArchethicWeb.FaucetRateLimiter do
...
   def handle_call({:block_status, address}, _from, state) do
+    address =
+      case TransactionChain.fetch_genesis_address_remotely(address) do
+        {:ok, genesis_address} -> genesis_address
+        _ -> address
+      end
+
     reply =
-      if address_state = Map.get(state, address) do
+      if address_state = Map.get(state, String.upcase(address)) do
```
samuel-uniris, line 67: What is the purpose of `String.upcase` here ? 
Addresses should be of binary form

samuel-uniris (changes requested):

lib/archethic_web/faucet_rate_limiter.ex
prix-uniris: This is just to store it in the map which is only the facuet_rate_limiter state. When user requests It comes as string so storing in facuet_rate_limiter as address -> some map values , where address is a string key and Upcase to keep consistency

lib/archethic/transaction_chain.ex
prix-uniris: Okay

lib/archethic_web/faucet_rate_limiter.ex
samuel-uniris: map keys can be binary, so the best would be to decode the input in the controller as binary and use it later, so you will remove the case issues

prix-uniris: Okay makes sense.

samuel-uniris (approved):

PR #487, opened 2022-08-02
Correct self repair notifier

### LINKED ISSUE #348: Fix self repair notifier
A notifier is implemented when the topology changes triggering the a self repair because the shards change

However the process is crashing causing errors from the DB mentioning the transactions don't exist.

This should be fixed to reactive it (being disabled) and support this shard repairing.

samuel-uniris:
Trigger repair on the right P2P network change

# Description

Fixes #348 

## Type of change

# How Has This Been Tested?

When an authorized node is considered as unauthorized or globally offline, this means a topology change occurred, hence the notifier will repair transaction shards but proposing transactions to new storage nodes.

By disconnecting an authorized node the notifier should replicate all the transactions this node was in charge of.
Logs are also added to help to visualize the events.

# Checklist:

### DIFF
--- a/lib/archethic/self_repair/notifier.ex
+++ b/lib/archethic/self_repair/notifier.ex
@@ -1,24 +1,45 @@
 defmodule Archethic.SelfRepair.Notifier do
-  @moduledoc false
+  @moduledoc """
+  Process to handle repair in case of topology change by trying to replicate transactions to new shard composition.
+
+  When a node receive a topology change due to the unavailability of a node,
+  we compute the new election for the already stored transactions.
+
+  Hence, a new shard might me formed as we notify the new transactions to the
+  new storage nodes
+
+  ```mermaid
+  flowchart TD
+      A[Node 4] --x|Topology change notification| B[Node1]
+      B --> | List transactions| B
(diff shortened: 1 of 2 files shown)


lib/archethic/self_repair/notifier.ex
```diff
@@ -1,104 +1,172 @@
...
+  require Logger
 
   def start_link(args \\ []) do
     GenServer.start_link(__MODULE__, args)
   end
 
   def init(_) do
     PubSub.register_to_node_update()
-    {:ok, []}
+    {:ok, %{notified: %{}}}
```
apoorv-2204, line 50: if there is process crash??, and in a situation where all nodes  waiting for each other for node_update:  message?

samuel-uniris: This state is just to avoid multiple node_update message for a given node, to avoid multiple repair in a close time.

samuel-uniris: If the process crashes, it will restart with an empty state, and wait another message of unavailability of the node

prix-uniris (approved):
LGTM

Issue #407, opened 2022-06-27
Create GraphQL API to get token information


samuel-uniris:
### Is your feature request related to a problem?

Because tokens are created on transaction chain, we might change the supply along the way, causing the addresses to change;
Hence we don't know if the new token is on the right chain.

### Describe the solution you'd like

We can create an API to get the token information as long as the genesis address to provide a better identification of the token.

```
token(address: "0fA02C9....") {
   genesisAddress,
   supply,
   name,
   symbol,
   properties ...
}
```

The API will fetch the genesis and token metadata and send them back in the response

apoorv-2204:
An api also required for getting Oracle Transaction by after time stamp

samuel-uniris:
Closed by #490

PR #490, opened 2022-08-03
GraphQL API to get token information #407

prix-uniris:
# Description
Adds a new GraphQL Schema to the API to fetch token information

Fixes # 407

## Type of change
- Feature Addition

# How Has This Been Tested?

Manual testing with single node

# Checklist:

### DIFF
--- a/lib/archethic_web/graphql_schema/resolver.ex
+++ b/lib/archethic_web/graphql_schema/resolver.ex
@@ -41,16 +41,18 @@ defmodule ArchethicWeb.GraphQLSchema.Resolver do
     ]
 
     # Using 5000 ms default timeout on tasks
-    genesis_address = case Task.yield(List.first(tasks)) do
-      {:ok, {:ok, genesis_address}} -> genesis_address
-      {:ok, {:error, :network_issue}} -> :network_issue
-      _ -> nil
-    end
+    genesis_address =
+      case Task.yield(List.first(tasks)) do
+        {:ok, {:ok, genesis_address}} -> genesis_address
+        {:ok, {:error, :network_issue}} -> :network_issue
+        _ -> nil
+      end
 
-    transaction_content = case Task.yield(List.last(tasks)) do
-      {:ok, {:ok, content}} -> content
-      _ -> nil
-    end
+    transaction_content =
+      case Task.yield(List.last(tasks)) do
+        {:ok, {:ok, content}} -> content
+        _ -> nil
+      end
 
     if transaction_content != nil and genesis_address != nil do
       {:ok,
--- a/lib/archethic_web/graphql_schema/resolver.ex
+++ b/lib/archethic_web/graphql_schema/resolver.ex
@@ -12,6 +12,8 @@ defmodule ArchethicWeb.GraphQLSchema.Resolver do
   alias Archethic.TransactionChain.TransactionData
   alias Archethic.TransactionChain.TransactionInput
 
+  require Logger
+
   @limit_page 10
 
   def get_balance(address) do
@@ -53,11 +55,22 @@ defmodule ArchethicWeb.GraphQLSchema.Resolver do
       {:ok, {:error, :network_issue}} ->
         {:error, "Network issue"}
 
+      {:ok, {:error, :decode_error}} ->
+        {:error, "Error in decoding transaction"}
+
       {:ok, {:error, :invalid_transaction}} ->
         {:error, "Transaction is not a token"}
 
       {:ok, {:error, :transaction_not_found}} ->
         {:error, "Transaction does not exist!"}
+
+      {:exit, reason} ->
+        Logger.debug("Task exited with reason")
+        Logger.debug(reason)
+        {:error, "Task Exited!"}
+
+      nil ->
+        {:error, "Task didn't responded within timeout!"}
     end
   end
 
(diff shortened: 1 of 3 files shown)


lib/archethic_web/graphql_schema/resolver.ex
```diff
@@ -33,6 +34,54 @@ defmodule ArchethicWeb.GraphQLSchema.Resolver do
     end
   end
 
+  def get_token(address) do
+    genesis_address =
```
samuel-uniris, line 38: You should parallelize into two tasks and await the result to make it more efficient

lib/archethic_web/graphql_schema/resolver.ex
```diff
@@ -33,6 +34,54 @@ defmodule ArchethicWeb.GraphQLSchema.Resolver do
...
+      case TransactionChain.fetch_genesis_address_remotely(address) do
+        {:ok, genesis_address} -> genesis_address
+        _ -> nil
```
samuel-uniris, line 41: if you are not getting the genesis address, this is a network issue, so we should handle this

lib/archethic_web/graphql_schema/resolver.ex
```diff
@@ -33,6 +34,54 @@ defmodule ArchethicWeb.GraphQLSchema.Resolver do
...
+      end
+
+    transaction_content =
+      case Archethic.search_transaction(address) do
+        {:ok, %Transaction{data: %TransactionData{content: content}, type: type}} ->
```
samuel-uniris, line 46: Instead of doing if `type == :token`, you can pattern match the type of the transaction.
This will reduce the complexity of the code and make it easier to read/understand

samuel-uniris (changes requested):

lib/archethic_web/graphql_schema/resolver.ex
prix-uniris: Yup that solves it

lib/archethic_web/graphql_schema/resolver.ex
```diff
@@ -33,6 +34,64 @@ defmodule ArchethicWeb.GraphQLSchema.Resolver do
...
+        _ -> nil
+      end
+
+    transaction_content =
+      case Task.yield(List.last(tasks)) do
+        {:ok, {:ok, content}} -> content
+        _ -> nil
+      end
+
+    if transaction_content != nil and genesis_address != nil do
```
samuel-uniris, line 57: I think you could handle it with the `with` statement.

You could do something like this
```elixir

t1 = Task.async(fn -> TransactionChain.fetch_genesis_address_remotely(address) end)
t2 = Task.async(fn -> get_transaction_content(address) end)

with {:ok, genesis_address} <- Task.yield(t1),
     {:ok, tx_content} <- Task.yield(t2) do
    {:ok,
       %{
         genesis: genesis_address,
         name: Map.get(transaction_content, "name"),
         supply: Map.get(transaction_content, "supply"),
         symbol: Map.get(transaction_content, "symbol"),
         type: Map.get(transaction_content, "type"),
         properties: do_reduce_properties(Map.get(transaction_content, "properties"))
       }}
else
  {:error, :network_issue} ->
     {:error, "Network issue"}
  {:error, :invalid_transaction} ->
     {:error, "Transaction is not a token"}
  {:error, :transaction_not_exists} ->
     {:error, "Transaction doesn't exists"}
end
```

samuel-uniris: Just to improve the readability

prix-uniris: Yes that's much more elegant and better, Thanks. Made the changes

lib/archethic_web/graphql_schema/resolver.ex
```diff
@@ -33,6 +34,56 @@ defmodule ArchethicWeb.GraphQLSchema.Resolver do
...
+        {:error, "Transaction does not exist!"}
+    end
+  end
+
+  defp get_transaction_content(address) do
+    case Archethic.search_transaction(address) do
+      {:ok, %Transaction{data: %TransactionData{content: content}, type: :token}} ->
+        case Jason.decode(content) do
+          {:ok, map} -> {:ok, map}
+          _ -> {:error, :decode_error}
```
samuel-uniris, line 69: I didn't see the `decode_error` in the `with/else` case

samuel-uniris (changes requested):

lib/archethic_web/graphql_schema/resolver.ex
prix-uniris: Added additional cases.

lib/archethic_web/graphql_schema/resolver.ex
```diff
@@ -33,6 +36,67 @@ defmodule ArchethicWeb.GraphQLSchema.Resolver do
...
+         properties: do_reduce_properties(Map.get(transaction_content, "properties"))
+       }}
+    else
+      {:ok, {:error, :network_issue}} ->
+        {:error, "Network issue"}
+
+      {:ok, {:error, :decode_error}} ->
+        {:error, "Error in decoding transaction"}
+
+      {:ok, {:error, :invalid_transaction}} ->
```
samuel-uniris, line 61: This one is not used right ?
If you want to return this error, you should find a case where this will be happen

prix-uniris: Which one? :invalid_transaciton one

samuel-uniris: yes

prix-uniris: Yup there's no case for that one, removing it then.

samuel-uniris (approved):

Issue #468, opened 2022-07-22
Faucet - add a trim in the address field


redDwarf03:
### Is your feature request related to a problem?

if the address contains blank at the end of the address, a control says that the address is malformated

### Describe the solution you'd like

Could you add a trim function before the control please

imnik11:
we can just add this before checking 
 https://github.com/archethic-foundation/archethic-node/blob/559e3d699ac428df4f5f5c7cbd18401e2e60b3e9/lib/archethic_web/controllers/faucet_controller.ex#L45

`    address = String.trim(address)
`

samuel-uniris:
Closed by #476

PR #491, opened 2022-08-03
Minor fix

apoorv-2204:
# Description
resolve some test warnings

Fixes # (#468)

## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/test/archethic/beacon_chain/update_test.exs
+++ b/test/archethic/beacon_chain/update_test.exs
@@ -1,90 +1,115 @@
 defmodule Archethic.BeaconChain.UpdateTest do
   use ArchethicCase
 
-  alias Archethic.BeaconChain.Update
+  alias Archethic.BeaconChain.Update, as: BeaconUpdate
 
   alias Archethic.P2P.Node
 
   import Mox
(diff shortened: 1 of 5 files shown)


test/archethic/beacon_chain/update_test.exs
```diff
@@ -6,7 +6,7 @@ defmodule Archethic.BeaconChain.UpdateTest do
   alias Archethic.P2P.Node
 
   import Mox
-  import GenServer
+  alias GenServer
```
samuel-uniris, line 9: I don't see the need of the alias or the import
Also you should use wrapped functions in the module instead (`subscribe`, ...)

apoorv-2204: multiple Update processes(are spawned) with send and assert_recieve , 
maybe due to that we require specific pid GenServer cast

samuel-uniris: I don't think it's the reason, but better is to abstract genserver calls through the API even for tests

samuel-uniris (approved):

PR #494, opened 2022-08-05
Improve latency handling

### LINKED ISSUE #493: Better support of low latency
When node latency starts the rise - for instance 200ms (europe -> to australia or US) and some important packet loss(ie. 10%), the network does not behave normally.

Some requests are no asynchronous while they could be
The timeouts should have a better management (to see if we have to implement a retry in the queries)
We have to identify where are the bottleneck (latency can vary from fiber to DSL within a WAN network with high distances)

Related issues are:
- #110 
- #498 
- #499
- #510 
- #568

### LINKED ISSUE #498: Make chain replication asynchronous

When the peers have a high latency > 200ms, due to some distance or network issue, we have to consider the fact to not waiting the chain replication in the validation workflow. This is causing timeout as the call is synchronous.

By using asynchronous messaging, the validation will submit the replication message and await the end of the replication by the chain storage nodes. This will remove the timeouts and synchronicity of the flow.

### Additional context

_No response_

samuel-uniris:
# Description

Improve latency handling to make the system more resilient

Fixes #498 #497 #499

## Type of change

# How Has This Been Tested?

To test with high latency, we can use a tool like [comcast](https://github.com/tylertreat/comcast) and with some latencies: 

```sh
comcast --device=lo0 --latency=200 --target-bw=1000 --packet-loss=10%
```

- We should not have timeouts in the ReplicateTransactionChain
- We should not have timeouts in the self-repair

# Checklist:


apoorv-2204:
https://github.com/archethic-foundation/archethic-node/blob/4507fb8f2349bd64a6979a8295fe78db1897b735/lib/archethic/mining/distributed_workflow.ex#L351

we can also leverage this, as Mentioned by Samuel. 
```
# TODO: Provide a better waiting time management
          # for example rolling percentile latency could be way to achieve this
          # (https://cs.stackexchange.com/a/129178)
```

lib/archethic/mining/standalone_workflow.ex
```diff
@@ -1,11 +1,11 @@
 defmodule Archethic.Mining.StandaloneWorkflow do
   @moduledoc """
-  Transaction validation is performed in a single node processing.
+  Transactpion validation is performed in a single node processing.
```
prix-uniris, line 3: A missed typo.

lib/archethic/p2p/message.ex
```diff
@@ -59,6 +59,7 @@ defmodule Archethic.P2P.Message do
   alias __MODULE__.RegisterBeaconUpdates
```
prix-uniris, line 59: A better way for import can be like this syntax : alias __MODULE__.{ModuleName1, ModuleName2, etc.}.
https://elixir-lang.org/getting-started/alias-require-and-import.html#multi-aliasimportrequireuse

test/archethic/mining/distributed_workflow_test.exs
```diff
@@ -914,6 +918,26 @@ defmodule Archethic.Mining.DistributedWorkflowTest do
             Workflow.add_cross_validation_stamp(coordinator_pid, stamp)
           end
 
+          receive do
+            {:ack_replication, sig, pub} ->
```
prix-uniris, line 922: Won't this case always be true and bottom receive will not work?

prix-uniris (commented):
Few typos and improvements can be made.

lib/archethic/mining/standalone_workflow.ex
samuel-uniris: ok

lib/archethic/p2p/message.ex
samuel-uniris: I think this can be improved for the entire codebase, so the scope of the PR would be too big I guess.
But yes you are right, it might be better.

test/archethic/mining/distributed_workflow_test.exs
samuel-uniris: Here you are awaiting the replication to be finished, once it's done you are notifying the validation process about the completion of the replication.
So we have to do it for all the storage nodes involved.

prix-uniris (approved):
Tested with comcast with 2 nodes locally, worked fine.
comcast --device=lo --latency=200 --target-bw=1000 --packet-loss=10%
LGTM

Issue #423, opened 2022-06-29
Ensure network transaction type only passed in a network transaction chain


samuel-uniris:
### Is your feature request related to a problem?

Network transactions are transaction which have a direct impact on the network and replications from shared secrets, to oracle and miner rewards.
Hence not other transaction chain should have the same type, ex: single shared secrets, single miner reward chain, ...

### Describe the solution you'd like

We can check the previous public key of the transaction to get previous and perform verification about the last address known by its type of transaction, and hence validate it or not.

So the network transaction types cannot be used by anybody.

samuel-uniris:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Workspace-61efec9c93391c001e69a6dc/issues/archethic-foundation/archethic-node/423?planning-poker) with ZenHub @apoorv-2204 @imnik11 @Neylix @prix-uniris

PR #495, opened 2022-08-05
Resolves: Ensure network transaction type only passed in a network transaction chain#423

apoorv-2204:
# Description
Validate previous txn should be of type network only and public key exists in it.

Fixes # (#423)

## Type of change

# How Has This Been Tested?
- [x] Distributed Testing 
dependent on #508 #515 

# Checklist:

### DIFF
--- a/lib/archethic/bootstrap.ex
+++ b/lib/archethic/bootstrap.ex
@@ -3,19 +3,20 @@ defmodule Archethic.Bootstrap do
   Manage Archethic Node Bootstrapping
   """
 
-  alias __MODULE__.NetworkInit
-  alias __MODULE__.Sync
-  alias __MODULE__.TransactionHandler
+  alias Archethic.Bootstrap.{
+    NetworkInit,
+    Sync,
+    TransactionHandler
+  }
 
-  alias Archethic.Crypto
-
-  alias Archethic.Networking
-
-  alias Archethic.P2P
-  alias Archethic.P2P.Node
-  alias Archethic.P2P.Listener
-
-  alias Archethic.SelfRepair
+  alias Archethic.{
+    Crypto,
+    Networking,
+    P2P,
+    P2P.Node,
+    P2P.Listener,
+    SelfRepair
+  }
 
   require Logger
 
@@ -207,6 +208,7 @@ defmodule Archethic.Bootstrap do
       Logger.info("Synchronization finished")
     end
 
+    Archethic.Bootstrap.NetworkConstraints.persist_genesis_address()
     Sync.publish_end_of_sync()
     SelfRepair.start_scheduler()
 
--- a/lib/archethic/bootstrap/network_constraints.ex
+++ b/lib/archethic/bootstrap/network_constraints.ex
@@ -11,7 +11,13 @@ defmodule Archethic.Bootstrap.NetworkConstraints do
 
   @spec persist_genesis_address() :: :ok
   def persist_genesis_address() do
-    res = process_gen_addr()
+    persist(:oracle)
+
+    res = %{
+      reward: persist(:reward),
+      node_shared_secrets: persist(:node_shared_secrets),
+      origin: persist(:origin)
+    }
 
     Logger.debug(
       "Gen Addr: reward: #{res.reward}, origin: #{res.origin}, nss: #{res.node_shared_secrets}"
@@ -24,17 +30,6 @@ defmodule Archethic.Bootstrap.NetworkConstraints do
     end
 
     :ok
-  end
-
-  @spec process_gen_addr() :: map()
-  def process_gen_addr() do
-    persist(:oracle)
-
-    %{
-      reward: persist(:reward),
-      node_shared_secrets: persist(:node_shared_secrets),
-      origin: persist(:origin)
-    }
   end
 
   @spec persist(:oracle | :reward | :origin | :node_shared_secrets) :: :ok | :error
--- a/lib/archethic/bootstrap/network_constraints.ex
+++ b/lib/archethic/bootstrap/network_constraints.ex
@@ -1,0 +1,89 @@
+defmodule Archethic.Bootstrap.NetworkConstraints do
+  @moduledoc false
+
+  alias Archethic.{
+    Reward,
+    SharedSecrets,
+    OracleChain
+  }
+
+  require Logger
+
+  @spec persist_genesis_address() :: :ok
+  def persist_genesis_address() do
+    res = process_gen_addr()
+
+    Logger.debug(
+      "Gen Addr: reward: #{res.reward}, origin: #{res.origin}, nss: #{res.node_shared_secrets}"
+    )
+
+    if Enum.all?(res, fn {_k, v} -> v == :ok end) do
+      Logger.info("Genesis Address Loading: Successful")
+    else
+      Logger.info("Genesis Address Loading: Failed, Resheduled to Next-Self-Repair")
+    end
+
+    :ok
+  end
+
+  @spec process_gen_addr() :: map()
+  def process_gen_addr() do
+    persist(:oracle)
+
+    %{
+      reward: persist(:reward),
+      node_shared_secrets: persist(:node_shared_secrets),
+      origin: persist(:origin)
+    }
+  end
+
+  @spec persist(:oracle | :reward | :origin | :node_shared_secrets) :: :ok | :error
+  def persist(:reward) do
+    case Reward.get_gen_addr() do
+      nil ->
+        Reward.persist_gen_addr()
+
+      gen_addr when is_binary(gen_addr) ->
+        :ok
+    end
+  end
+
+  def persist(:origin) do
+    case SharedSecrets.get_gen_addr(:origin) do
+      nil ->
+        SharedSecrets.persist_gen_addr(:origin)
+
+      _gen_addr_list ->
+        :ok
+    end
+  end
+
+  def persist(:node_shared_secrets) do
+    case SharedSecrets.get_gen_addr(:node_shared_secrets) do
+      nil ->
+        SharedSecrets.persist_gen_addr(:node_shared_secrets)
+
+      gen_addr when is_binary(gen_addr) ->
+        :ok
+    end
+  end
+
+  def persist(:oracle) do
+    try do
+      OracleChain.update_summ_gen_addr()
+      Logger.info("Oracle Gen Addr Table: Loaded")
+
+      if gen_addr = OracleChain.get_gen_addr() do
+        Logger.debug("New Oracle Gen Addr")
+        Logger.debug(gen_addr)
+      end
+
+      :ok
+    rescue
+      _e ->
+        Logger.info("Oracle Gen Addr Table: Failed ")
+
+        :error
+    end
+  end
+end
--- a/lib/archethic/mining/pending_transaction_validation.ex
+++ b/lib/archethic/mining/pending_transaction_validation.ex
@@ -1,40 +1,35 @@
 defmodule Archethic.Mining.PendingTransactionValidation do
   @moduledoc false
 
-  alias Archethic.Contracts
-  alias Archethic.Contracts.Contract
+  alias Archethic.{
+    Contracts,
+    Contracts.Contract,
+    Crypto,
+    DB,
+    SharedSecrets,
+    Election,
+    Governance,
+    Networking,
+    OracleChain,
+    P2P,
+    P2P.Message.FirstPublicKey,
+    P2P.Message.GetFirstPublicKey,
+    P2P.Node,
+    Reward,
+    SharedSecrets.NodeRenewal,
+    Utils,
+    TransactionChain
+  }
 
-  alias Archethic.Crypto
+  alias Archethic.TransactionChain.{
+    Transaction,
+    TransactionData,
+    TransactionData.Ledger,
+    TransactionData.Ownership,
+    TransactionData.TokenLedger
+  }
 
-  alias Archethic.DB
-
-  alias Archethic.SharedSecrets
-
-  alias Archethic.Election
-
-  alias Archethic.Governance
   alias Archethic.Governance.Code.Proposal, as: CodeProposal
-  alias Archethic.Networking
-  alias Archethic.OracleChain
-
-  alias Archethic.P2P
-  alias Archethic.P2P.Message.FirstPublicKey
-  alias Archethic.P2P.Message.GetFirstPublicKey
-  alias Archethic.P2P.Node
-
-  alias Archethic.Reward
-
-  alias Archethic.SharedSecrets.NodeRenewal
-
-  alias Archethic.TransactionChain
-  alias Archethic.TransactionChain.Transaction
-  alias Archethic.TransactionChain.TransactionData
-  alias Archethic.TransactionChain.TransactionData.Ledger
-  alias Archethic.TransactionChain.TransactionData.Ownership
-  alias Archethic.TransactionChain.TransactionData.TokenLedger
-
-  alias Archethic.Utils
-
   require Logger
 
   @doc """
@@ -50,7 +45,8 @@ defmodule Archethic.Mining.PendingTransactionValidation do
     with true <- Transaction.verify_previous_signature?(tx),
          :ok <- validate_contract(tx),
          :ok <- validate_content_size(tx),
-         :ok <- do_accept_transaction(tx, validation_time) do
+         :ok <- do_accept_transaction(tx, validation_time),
+         :ok <- validate_previous_transaction_type?(tx) do
       :telemetry.execute(
         [:archethic, :mining, :pending_transaction_validation],
         %{duration: System.monotonic_time() - start},
@@ -59,6 +55,14 @@ defmodule Archethic.Mining.PendingTransactionValidation do
 
       :ok
     else
+      {:error, :invalid_tx_type} ->
+        Logger.error("Invalid Transaction Type",
+          transaction_address: Base.encode16(address),
+          transaction_type: type
+        )
+
+        {:error, "Invalid Transaction Type"}
+
       false ->
         Logger.error("Invalid previous signature",
           transaction_address: Base.encode16(address),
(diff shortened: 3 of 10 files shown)


lib/archethic/mining/pending_transaction_validation.ex
```diff
@@ -70,6 +71,37 @@ defmodule Archethic.Mining.PendingTransactionValidation do
...
+        true
+
+      true ->
+        # when network tx , check with previous transaction
+        do_validate_previous_transaction_type?(tx)
+    end
+  end
+
+  def do_validate_previous_transaction_type?(tx) do
+    case TransactionChain.get_last_transaction(tx.address, [:address, :type]) do
```
samuel-uniris, line 88: I think you can just get the previous address from: `Transaction.previous_address(tx)` which is calculating the hash from the previous public key, then apply a verification if the type is not the same

samuel-uniris: You just need to be careful because some network transaction chain can have multiple types:
- oracles (oracle, oracle_summary)
- rewards (mint_rewards, node_rewards)

samuel-uniris (changes requested):

lib/archethic/mining/pending_transaction_validation.ex
```diff
@@ -70,6 +76,62 @@ defmodule Archethic.Mining.PendingTransactionValidation do
...
+        false
+
+      {:ok, last_txn = %Transaction{}} ->
+        match_last_tx_type?(curr_txn.type, last_txn.type) &&
+          Transaction.verify_previous_signature?(curr_txn)
+    end
+  end
+
+  @spec match_last_tx_type?(:atom, :atom) :: boolean
+  def match_last_tx_type?(curr_tx_type, last_txn_type) do
```
samuel-uniris, line 118: The logic is here but using pattern matching would be a bit clearer in my opinion

samuel-uniris (changes requested):

lib/archethic/mining/pending_transaction_validation.ex
```diff
@@ -47,7 +47,8 @@ defmodule Archethic.Mining.PendingTransactionValidation do
     with true <- Transaction.verify_previous_signature?(tx),
          :ok <- validate_contract(tx),
          :ok <- validate_content_size(tx),
-         :ok <- do_accept_transaction(tx) do
+         :ok <- do_accept_transaction(tx),
+         :valid_chain_type <- validate_previous_transaction_type?(tx) do
```
samuel-uniris, line 51: You should use :ok instead

lib/archethic/mining/pending_transaction_validation.ex
```diff
@@ -64,12 +65,71 @@ defmodule Archethic.Mining.PendingTransactionValidation do
...
+        # Last txn is
+        # TODO imp: impl handling of last KO txn
+        false
+
+      {:ok, last_txn = %Transaction{}} ->
+        match_last_tx_type?(curr_txn.type, last_txn.type)
+    end
+  end
+
+  @spec match_last_tx_type?(:atom, :atom) :: boolean
```
samuel-uniris, line 122: instead of atom you should use Transaction.transaction_type()

samuel-uniris (changes requested):

lib/archethic/mining/pending_transaction_validation.ex
```diff
@@ -64,12 +65,72 @@ defmodule Archethic.Mining.PendingTransactionValidation do
...
+        false
+
+      {:ok, last_txn = %Transaction{}} ->
+        match_last_tx_type?(curr_txn.type, last_txn.type)
+    end
+  end
+
+  @spec match_last_tx_type?(Transaction.transaction_type(), Transaction.transaction_type()) ::
+          boolean
+  def match_last_tx_type?(:oracle, :oracle_summary), do: true
```
samuel-uniris, line 124: I don't think the cyclic mode is always true, for instance oracle can be followed by oracle and oracle summary but nothing after oracle_summary

apoorv-2204: we check for current txn type and last txn type only, and we are relying on network bootstrap , for last txn type.

Neylix:
Hi !
I think we can have a better way to ensure that network transaction comes from network chain.
Actually, using previous address require to do a DB read to get the previous address, and if there is no previous address we need to do complex control.

An other idea can be to: 
- store the genesis address of all network chain in a persistent term during bootstrap (except for the oracle chain and beacon chain)
- store the genesis address of the current chain for oracle and beacon chain at each new summary in an ets table
- in the pending validation, get the genesis address of the incoming transaction and depending of the transaction type, retrieve the good genesis address from the persistent term or ets table and compare both genesis address

The advantage is that we control whenever if it is the first transaction of the chain or not, and we mainly use ets table read instead of DB read

lib/archethic/mining/pending_transaction_validation.ex
```diff
@@ -70,7 +78,93 @@ defmodule Archethic.Mining.PendingTransactionValidation do
...
+    pre_stored_genesis_address.current
+    |> elem(0) == first_addr_from_prev_addr ||
+      pre_stored_genesis_address.prev
+      |> elem(0) == first_addr_from_prev_addr
+  end
+
+  def validate_network_chain?(type, curr_txn = %Transaction{})
+      when type in [:mint_rewards, :node_rewards] do
+    # singleton tx chain in network lifespan
+    pre_stored_genesis_address = :persistent_term.get(:reward_gen_addr)
```
samuel-uniris, line 124: You have to try catch the `get` calls if the entry is not present

lib/archethic/shared_secrets.ex
```diff
@@ -116,4 +116,22 @@ defmodule Archethic.SharedSecrets do
...
+    :persistent_term.put(@ss_gen_key, addr)
+  end
+
+  @origin_gen_key :origin_gen_addr
+  def persist_gen_addr(addr, :origin) do
+    :persistent_term.put(@origin_gen_key, addr)
+  end
+
+  def get_gen_addr(:origin) do
+    :persistent_term.get(@origin_gen_key)
```
samuel-uniris, line 131: You have to try catch the `get` calls if the entry is not present

lib/archethic/shared_secrets.ex
```diff
@@ -116,4 +116,22 @@ defmodule Archethic.SharedSecrets do
...
+  end
+
+  def get_gen_addr(:node_shared_secrets) do
+    :persistent_term.get(@ss_gen_key)
```
samuel-uniris, line 135: You have to try catch the `get` calls if the entry is not present

lib/archethic/reward.ex
```diff
@@ -215,4 +215,13 @@ defmodule Archethic.Reward do
...
     RewardTokens.exists?(token_address)
   end
+
+  @key :reward_gen_addr
+  def persist_gen_addr(value) do
+    :persistent_term.put(@key, value)
+  end
+
+  def get_gen_addr() do
+    :persistent_term.get(@key)
```
samuel-uniris, line 225: You have to try catch the `get` calls if the entry is not present

lib/archethic/mining/distributed_workflow.ex
```diff
@@ -1003,5 +1003,6 @@ defmodule Archethic.Mining.DistributedWorkflow do
     }
 
     P2P.broadcast_message(storage_nodes, message)
```
samuel-uniris, line 1005: It returns :ok by default

samuel-uniris (changes requested):

lib/archethic/bootstrap.ex
```diff
@@ -206,12 +213,91 @@ defmodule Archethic.Bootstrap do
...
     SelfRepair.start_scheduler()
 
     :persistent_term.put(:archethic_up, :up)
   end
 
+  # case1: When a node joins a already initialized network
+  # after syncing all transactions persist the gen_addr
+  # case2: node rejoins and requires sync, at end of sync persist gen_addr
+  # case3: first time network initialization
+  def persist_gen_addr() do
```
Neylix, line 227: I think that you should not list the transaction here, it will takes a lot of ressources.
As we already knows the seed, you can derive the genesis address directly from the seed.
For exemple, node shared secret would be :
`SharedSecrets.persist_gen_addr()`
-> 
```
def persist_gen_addr() do
    ss_adrr =
      :ets.lookup(@keystore_table, :shared_secrets_index)
      |> Crypto.derive_keypair(0)
      |> elem(0)
      |> Crypto.derive_address()

    origin_software_adrr =
      Archethic.SharedSecrets.get_origin_family_seed(:software)
      |> Crypto.derive_keypair(0)
      |> elem(0)
      |> Crypto.derive_address()

    :persistent_term.put(@ss_gen_key, ss_adrr)
    :persistent_term.put(@origin_gen_key, origin_software_adrr)
  end
```

This can be apply for other chain (exept beacon and oracle chain)

lib/archethic/bootstrap.ex
```diff
@@ -206,12 +213,91 @@ defmodule Archethic.Bootstrap do
...
+      end)
+
+      Task.Supervisor.async_nolink(Archethic.TaskSupervisor, fn ->
+        TransactionChain.list_addresses_by_type(:origin)
+        |> Enum.at(0)
+        |> TransactionChain.get_genesis_address()
+        |> SharedSecrets.persist_gen_addr(:origin)
+      end)
+
+      Task.Supervisor.async_nolink(Archethic.TaskSupervisor, fn ->
```
Neylix, line 283: I think we don't need to control `node` transaction as there will be many node transaction chain.

lib/archethic/bootstrap/network_init.ex
```diff
@@ -84,6 +84,8 @@ defmodule Archethic.Bootstrap.NetworkInit do
     tx
     |> self_validation()
     |> self_replication()
+
+    SharedSecrets.persist_gen_addr(tx.address, :node_shared_secrets)
```
Neylix, line 88: I think you don't need to persist the term here as you do it after the bootstrap (function post_bootstrap is always call even after network init)

lib/archethic/oracle_chain/scheduler.ex
```diff
@@ -279,6 +291,10 @@ defmodule Archethic.OracleChain.Scheduler do
 
     current_time = DateTime.utc_now() |> DateTime.truncate(:second)
     next_summary_date = next_date(summary_interval, current_time)
+    # put next summarry date genesis address
+    next_summary_date
```
Neylix, line 295: I don't understand this line, it maybe a misstake

Neylix (changes requested):

lib/archethic/bootstrap/network_constraints.ex
```diff
@@ -0,0 +1,136 @@
+defmodule Archethic.Bootstrap.NetworkConstraints do
```
samuel-uniris, line 1: I don't think this module is necessary.
It's even too complex.

samuel-uniris (changes requested):
What you could do it's to manage it straight after bootstrap. For shared secrets you could get the last transaction of the chain and retrieve it's genesis.

lib/archethic/bootstrap/network_constraints.ex
apoorv-2204: I tried those method, many like getting last txn , enforce loading of nss txs, this method evolved lot of experimentations, of retrieving. we can derive gen addr of nss and reward. until the node is authorized .because  in nss we require to have seed , which is warped secret and not accessible until authorised. I dont know how to proceed futher.

apoorv-2204: Its logic is that keep trying to persist gen addr with node update msg, once persists exit normally, so transient.?

apoorv-2204:
> What you could do it's to manage it straight after bootstrap. For shared secrets you could get the last transaction of the chain and retrieve it's genesis.

:ok

lib/archethic/bootstrap/network_constraints.ex
samuel-uniris: You don't even need to access the NSS seed I think just access to the genesis address

apoorv-2204: Resovles:Enforce resync of last network transactions after bootstrap#508 #515 .
After forced resync we can load gen addr from DB

lib/archethic/bootstrap/network_constraints.ex
```diff
@@ -0,0 +1,134 @@
...
+      true
+    rescue
+      _error -> false
+    end
+  end
+
+  def persist(:reward, false) do
+    try do
+      case TransactionChain.list_addresses_by_type(:mint_rewards)
+           |> Enum.at(0) do
```
Neylix, line 89: Here you can do some improvement. As we only want the first address we don't need to wait for all the addresses to be fetched.
`TransactionChain.list_addresses_by_type` will return a Stream and `Enum.at(0)` will wait for all the Stream to be finished before taking the element 0.
You can add `Stream.take(1)` before the Enum. This will stop the Stream after the first element !
```
TransactionChain.list_addresses_by_type(:mint_rewards)
|> Stream.take(1)
|> Enum.at(0)
```

lib/archethic/bootstrap/network_constraints.ex
```diff
@@ -0,0 +1,134 @@
+defmodule Archethic.Bootstrap.NetworkConstraints do
+  @moduledoc """
+  This GenServer make sure that Node load network constraints before any transaction validation.
+  Updates gen_addr constraints post bootstrap, if fails wait for node to get authorized
+  """
+
+  use GenServer, restart: :transient
```
Neylix, line 7: I think you don't need to use a GenServer anymore as you don't get the NSS seed anymore.
Now you fetch the NSS and reward chain from DB and get the genesis address, so you can diretly persist addresses at the end of bootstrap no ?
I tried with the same function than in NetworkConstraints, it works well

Neylix (changes requested):

lib/archethic/bootstrap/network_constraints.ex
apoorv-2204: nice

lib/archethic/bootstrap/network_constraints.ex
Neylix: How can we get nil from DB as all these chain are initialized during network init, or downloaded during bootstrap self repair ?

apoorv-2204: Yes I had the same question, so we  loop until we persist.(on node udpate)
If we dont loop it will crash , may also crash bootstrap process , and restrict node in validation of tx.
So need of independent transient process.

lib/archethic/bootstrap.ex
```diff
@@ -206,12 +209,54 @@ defmodule Archethic.Bootstrap do
...
 
+  #  persist gen_addr
+  #  New node joining, sync all txs
+  #  node rejoins and requires sync
+  #  first time network initialization
+  @spec persist_gen_addr :: Task.t()
+  def persist_gen_addr() do
+    # Implemented
+    # :node_shared_secrets, :origin, :oracle , :oracle_summary, :node_rewards ,:mint_rewards
+    Task.Supervisor.async_nolink(Archethic.TaskSupervisor, fn ->
```
samuel-uniris, line 227: You don't think to leverage async process for simple persistent_term or ETS inserts

lib/archethic/mining/pending_transaction_validation.ex
```diff
@@ -70,6 +79,115 @@ defmodule Archethic.Mining.PendingTransactionValidation do
...
+        first_addr =
+          tx
+          |> Transaction.previous_address()
+          |> TransactionChain.get_genesis_address()
+
+        gen_addr == first_addr
+    end
+  end
+
+  def validate_network_chain?(type, _tx = %Transaction{})
```
samuel-uniris, line 147: I don't think you need this one as well as L:186.
You should use a default one to be by-passed

lib/archethic/shared_secrets.ex
```diff
@@ -116,4 +116,47 @@ defmodule Archethic.SharedSecrets do
         :biometric
     end
   end
+
+  @ss_gen_key :node_shared_secrets_gen_addr
+  @spec persist_gen_addr(:node_shared_secrets | :origin) :: :ok
+  def persist_gen_addr(:node_shared_secrets) do
+    addr =
+      Archethic.Crypto.node_shared_secrets_public_key(0)
```
samuel-uniris, line 124: This will crash as the daily nonce is not loaded.
You should as mentioned earlier to leverage post self-repair operation to look at the first genesis address of the type shared secrets from the DB.

samuel-uniris (changes requested):

lib/archethic/bootstrap.ex
apoorv-2204: my bad, Old commits showed up

lib/archethic/oracle_chain.ex
```diff
@@ -186,4 +190,82 @@ defmodule Archethic.OracleChain do
...
+
+  @doc """
+    Methods Serves to resolve Situation
+    time  : event
+    12:00 : auth node rejoins
+    12:01:40 : completes sync ,recieves oracle summary to be verified by this auth node
+    12:01:55 : will result failure as last oracle_gen_addr not there
+  """
+  @spec update_last_summ_gen_addr :: :ok
+  def update_last_summ_gen_addr() do
```
Neylix, line 202: Instead of calculating the last summary_date, you can just use `OracleChain.previous_summary_date()`

lib/archethic/oracle_chain.ex
```diff
@@ -186,4 +190,82 @@ defmodule Archethic.OracleChain do
...
+    :ok
+  end
+
+  @doc """
+  Updates ets table with current_summary_gen_addr
+  Address: genesis address of current(summarydate) oracle chain
+  DateTime: current summary datetime
+  """
+  @spec update_curr_summ_gen_addr :: :ok
+  def update_curr_summ_gen_addr() do
```
Neylix, line 236: Instead of calculating the next summary_date, you can just use `OracleChain.next_summary_date()`

lib/archethic/bootstrap.ex
```diff
@@ -207,13 +209,35 @@ defmodule Archethic.Bootstrap do
       Logger.info("Synchronization finished")
     end
 
+    persist(:oracle)
```
Neylix, line 212: Here you also have to persist all the genesis address.
You can add the function to persist oracle in the networkConstraint module, and add a flag in the function to determine if you are in bootstrap time or self repair time like `bootstrap?`
If you are on bootstrap you persist oracle, if not you don't

lib/archethic/mining/pending_transaction_validation.ex
```diff
@@ -73,6 +77,115 @@ defmodule Archethic.Mining.PendingTransactionValidation do
...
+        first_addr_from_prev_addr =
+          tx
+          |> Transaction.previous_address()
+          |> TransactionChain.get_genesis_address()
+
+        first_addr_from_prev_addr in origin_gen_addr_list
+    end
+  end
+
+  def validate_network_chain?(type, _tx = %Transaction{})
```
Neylix, line 178: Here you don't have to check all the type, you can just implement a default function that returns ok

Neylix (changes requested):

lib/archethic/mining/pending_transaction_validation.ex
apoorv-2204: Hi, Impl all suggestions.

lib/archethic/reward.ex
```diff
@@ -242,4 +248,32 @@ defmodule Archethic.Reward do
       |> DateTime.from_naive!("Etc/UTC")
     end
   end
+
+  @key :reward_gen_addr
+  @spec persist_gen_addr() :: :ok
+  def persist_gen_addr() do
+    try do
```
samuelmanzanera, line 255: In which case something will throw ?

lib/archethic/bootstrap.ex
```diff
@@ -207,6 +208,7 @@ defmodule Archethic.Bootstrap do
       Logger.info("Synchronization finished")
     end
 
+    Archethic.Bootstrap.NetworkConstraints.persist_genesis_address()
```
samuelmanzanera, line 211: You are already doing it at the end of the self-repair so why would you want to do it now ?

lib/archethic/bootstrap/network_constraints.ex
```diff
@@ -0,0 +1,89 @@
...
+      Logger.info("Genesis Address Loading: Successful")
+    else
+      Logger.info("Genesis Address Loading: Failed, Resheduled to Next-Self-Repair")
+    end
+
+    :ok
+  end
+
+  @spec process_gen_addr() :: map()
+  def process_gen_addr() do
```
samuelmanzanera, line 30: I think you could either make it private or embed it into `persiste_genesis_address`

lib/archethic/bootstrap.ex
samuelmanzanera: ok

lib/archethic/reward.ex
apoorv-2204: When first time node boots up, it will not have reward chain, post bootstrap. db can return nil, causing to crash

samuelmanzanera: But you are already catching the nil case

Neylix (approved):

PR #514, opened 2022-08-16
Resolves #507 : Start listening P2P connections after complete bootstrap 

### LINKED ISSUE #507: Start listening P2P connections after complete bootstrap

When a node, already exists and is authorized and available, restart quickly, this node will receive new mining requests hence its bootstrap might be out of date and the mining requests would cause errors.

To resolve this, we can refuse any incoming connections requests until the bootstrap is not finished.
Because after bootstrap we are setting a persistent_term `archethic_up`, we can use it in the `ListenerProtocol` to close connection until the persistent_term exists and be true.

This will make sure the node is ready to receive incoming requests

### Additional context

_No response_

apoorv-2204:
# Description

Whenever a node is bootstrapping , it must reject Request messages .  
   ListenerProtocol handles Incoming new messages from other nodes and
   no response is processed here.
   Connection modules handles this nodes request get messages and
   then its response is processed.

Fixes # (#507)

## Type of change

# How Has This Been Tested?

- iex -S mix with multiple nodes

# Checklist:

### DIFF
--- a/lib/archethic/p2p/listener_protocol.ex
+++ b/lib/archethic/p2p/listener_protocol.ex
@@ -1,5 +1,9 @@
 defmodule Archethic.P2P.ListenerProtocol do
   @moduledoc false
+  # ListenerProtocol handles Incoming new messages from other nodes and
+  # no response is processed here.
+  # Connection modules handles this nodes request get messages and
+  # then its response is processed.
 
   require Logger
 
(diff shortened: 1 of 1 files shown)


lib/archethic/p2p/listener_protocol.ex
```diff
@@ -38,22 +38,31 @@ defmodule Archethic.P2P.ListenerProtocol do
...
-        %Archethic.P2P.MessageEnvelop{
-          message: response,
-          message_id: message_id,
-          sender_public_key: Archethic.Crypto.first_node_public_key()
-        }
-        |> Archethic.P2P.MessageEnvelop.encode(sender_public_key)
-
-      transport.send(socket, encoded_response)
-    end)
+    <<msg_nb::8, _rest::bitstring>> = Archethic.P2P.Message.encode(message)
```
samuel-uniris, line 41: I don't think this is necessary

lib/archethic/p2p/listener_protocol.ex
```diff
@@ -38,22 +38,31 @@ defmodule Archethic.P2P.ListenerProtocol do
...
+    request_msg = [1..10, 15..21, 23, 26, 28, 31]
+
+    if :persistent_term.get(:archethic_up, nil) == nil && msg_nb in request_msg do
+      # node is Bootstrapping, reject any incoming request messages
+      # close conn return error
+      send(self(), {transport, socket})
```
samuel-uniris, line 47: You could use the `transport.close(socket)` instruction instead

samuel-uniris: But maybe you can manage this in the init, before to receive any message

samuel-uniris (changes requested):

lib/archethic/p2p/listener_protocol.ex
```diff
@@ -13,17 +17,22 @@ defmodule Archethic.P2P.ListenerProtocol do
   end
 
   def init({ref, transport, opts}) do
-    {:ok, socket} = :ranch.handshake(ref)
-    :ok = transport.setopts(socket, opts)
+    if :persistent_term.get(:archethic_up, nil) == nil do
```
samuel-uniris, line 20: This could be refactored to reduce the two times `nil`usage.
Just by inverting the position:
```elixir
if :persistent_tem.get(:archethic_up, nil) do
    # Handle the connection
else
   {:stop, :node_bootstrapping}
end
```

This is just for clarity

samuel-uniris (approved):

Issue #508, opened 2022-08-11
Enforce resync of last network transactions after bootstrap


ghost:
### Is your feature request related to a problem?

When a node which available and authorized restarts in a short period of time, this node will receive incoming requests for transaction validation.
However, it might miss some network transactions such as `node_shared_secrets` or `oracle`
So the validation will fail since its shared state will be out to date.

### Describe the solution you'd like

To resolve the out of sync issue for network transactions, we can try to re-sync at the end of the bootstrap by:
- fetching and loading the last node shared transaction
- fetching and loading the last oracle transaction

Hence, the node will be able to validate those transactions in the same state as others will do

samuel-uniris:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Product-and-Design-Workspace-61efe71b9083a90018250d89/issues/archethic-foundation/archethic-node/508?planning-poker) with ZenHub @apoorv-2204 @imnik11 @Neylix @prix-uniris

apoorv-2204:
- we can get paging address from DB, then request node to return tx s after  that.
- we may require to have DB.EmbeddedImpl for returning last 10 txns or last txs by paging address in future (for blind case scenario where paging address not available)

PR #515, opened 2022-08-17
Resovles:Enforce resync of last network transactions after bootstrap#508 

apoorv-2204:
# Description

If there is quick disconnection and re-connection of a (authorised and available) node, it is a possibility that it will miss some oracle and node shared secrets transactions , which results in inconsistent node state.Post Bootstrap ,we must resync  nss and oracle txs so that current node state is same as other nodes state. So during validation it will not report invalid tx

Fixes # (#508 )

## Type of change

# How Has This Been Tested?

- [x] unit test cases
- [x] Distributed testing with 3 nodes

## How to test?
- Start 3 nodes, wait for all of them to get authorised.
- Shutdown one of the node.
- wait for other nodes for one or two cycle of (nss ,oc) txs.
- restart the node And look of enforced re-sync in logs.
- after quick restart of that node 

Screenshot_20220907_163030

# Checklist:

### DIFF
--- a/lib/archethic/bootstrap.ex
+++ b/lib/archethic/bootstrap.ex
@@ -15,7 +15,9 @@ defmodule Archethic.Bootstrap do
     P2P,
     P2P.Node,
     P2P.Listener,
-    SelfRepair
+    SelfRepair,
+    TransactionChain,
+    Replication
   }
 
   require Logger
@@ -209,11 +211,70 @@ defmodule Archethic.Bootstrap do
     end
 
     Archethic.Bootstrap.NetworkConstraints.persist_genesis_address()
+    resync_network_chain()
+
     Sync.publish_end_of_sync()
     SelfRepair.start_scheduler()
 
     :persistent_term.put(:archethic_up, :up)
     Listener.listen()
+  end
+
+  def resync_network_chain() do
+    Logger.info("Enforced Resync: Started!")
+
+    if P2P.authorized_node?() && P2P.available_node?() do
+      # evict this node
+      nodes =
+        Enum.reject(
+          P2P.authorized_and_available_nodes(),
+          &(&1.first_public_key == Crypto.first_node_public_key())
+        )
+
+      do_resync_network_chain([:oracle, :node_shared_secrets], nodes)
+    end
+  end
+
+  @spec do_resync_network_chain(list(atom), list(P2P.Node.t()) | []) :: :ok
+  def do_resync_network_chain(_type_list, _nodes = []),
+    do: Logger.info("Enforce Reync of Network Txs: failure, No-Nodes")
+
+  # by type: Get gen addr, get last address (remotely  & locally)
+  # compare, if dont match, fetch last tx remotely
+  def do_resync_network_chain(type_list, nodes) when is_list(nodes) do
+    Task.Supervisor.async_stream_nolink(Archethic.TaskSupervisor, type_list, fn type ->
+      with addr when is_binary(addr) <- get_genesis_addr(type),
+           {:ok, rem_last_addr} <- TransactionChain.resolve_last_address(addr),
+           {local_last_addr, _} <- TransactionChain.get_last_address(addr),
+           false <- rem_last_addr == local_last_addr,
+           {:ok, tx} <- TransactionChain.fetch_transaction_remotely(rem_last_addr, nodes),
+           :ok <- Replication.validate_and_store_transaction_chain(tx) do
+        Logger.info("Enforced Resync: Success", transaction_type: type)
+        :ok
+      else
+        true ->
+          Logger.info("Enforced Resync: No new transaction to sync", transaction_type: type)
+          :ok
+
+        e when e in [nil, []] ->
+          Logger.debug("Enforced Resync: Transaction not available", transaction_type: type)
+          :ok
+
+        e ->
+          Logger.debug("Enforced Resync: Unexpected Error", transaction_type: type)
+          Logger.debug(e)
+      end
+    end)
+    |> Stream.run()
+  end
+
+  @spec get_genesis_addr(:node_shared_secrets | :oracle) :: binary() | nil
+  defp get_genesis_addr(:oracle) do
+    Archethic.OracleChain.get_gen_addr().current |> elem(0)
+  end
+
+  defp get_genesis_addr(:node_shared_secrets) do
+    Archethic.SharedSecrets.get_gen_addr(:node_shared_secrets)
   end
 
   defp first_initialization(
--- a/test/archethic/bootstrap_test.exs
+++ b/test/archethic/bootstrap_test.exs
@@ -1,51 +1,56 @@
 defmodule Archethic.BootstrapTest do
   use ArchethicCase
 
-  alias Archethic.Crypto
+  alias Archethic.{
+    Bootstrap,
+    Crypto,
+    P2P,
+    P2P.BootstrappingSeeds,
+    P2P.Node,
+    Replication,
+    SharedSecrets,
+    SharedSecrets.NodeRenewalScheduler,
+    TransactionChain,
+    TransactionFactory
+  }
+
+  alias Archethic.P2P.Message.{
+    GetTransactionChainLength,
+    TransactionChainLength,
+    BootstrappingNodes,
+    EncryptedStorageNonce,
+    GetBootstrappingNodes,
+    GetLastTransactionAddress,
+    GetStorageNonce,
+    GetTransaction,
+    GetTransactionChain,
+    GetTransactionSummary,
+    GetTransactionInputs,
+    GetFirstAddress,
+    FirstAddress,
+    LastTransactionAddress,
+    ListNodes,
+    NewTransaction,
+    NodeList,
+    NotFound,
+    NotifyEndOfNodeSync,
+    TransactionList,
+    TransactionInputList,
+    Ok,
+    GetFirstAddress,
+    NotFound
+  }
+
+  alias TransactionChain.{
+    Transaction,
+    TransactionSummary,
+    Transaction.ValidationStamp,
+    Transaction.ValidationStamp.LedgerOperations
+  }
 
   alias Archethic.BeaconChain.SlotTimer, as: BeaconSlotTimer
   alias Archethic.BeaconChain.SummaryTimer, as: BeaconSummaryTimer
-
-  alias Archethic.Bootstrap
-
-  alias Archethic.P2P
-  alias Archethic.P2P.BootstrappingSeeds
-  alias Archethic.P2P.Message.GetTransactionChainLength
-  alias Archethic.P2P.Message.TransactionChainLength
-  alias Archethic.P2P.Message.BootstrappingNodes
-  alias Archethic.P2P.Message.EncryptedStorageNonce
-  alias Archethic.P2P.Message.GetBootstrappingNodes
-  alias Archethic.P2P.Message.GetLastTransactionAddress
-  alias Archethic.P2P.Message.GetStorageNonce
-  alias Archethic.P2P.Message.GetTransaction
-  alias Archethic.P2P.Message.GetTransactionChain
-  alias Archethic.P2P.Message.GetTransactionSummary
-  alias Archethic.P2P.Message.GetTransactionInputs
-  alias Archethic.P2P.Message.LastTransactionAddress
-  alias Archethic.P2P.Message.ListNodes
-  alias Archethic.P2P.Message.NewTransaction
-  alias Archethic.P2P.Message.NodeList
-  alias Archethic.P2P.Message.NotFound
-  alias Archethic.P2P.Message.NotifyEndOfNodeSync
-  alias Archethic.P2P.Message.TransactionList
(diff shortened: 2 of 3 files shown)


lib/archethic/bootstrap.ex
```diff
@@ -206,12 +208,32 @@ defmodule Archethic.Bootstrap do
...
+
     Sync.publish_end_of_sync()
     SelfRepair.start_scheduler()
-
     :persistent_term.put(:archethic_up, :up)
   end
 
+  def resync_network_chain(type_list) do
+    Task.Supervisor.async_stream_nolink(Archethic.TaskSupervisor, type_list, fn type ->
+      with nodes <- P2P.authorized_and_available_nodes(),
```
samuel-uniris, line 220: Maybe something like to avoid `error` == nil

```elixir
if TransactionChain.list_addresses_by(type) |> Enum.at(0) do
  with {:ok, last_addr} <- TransactionChain.fetch_last_address_remotely(addr, nodes),
       {:ok, last_tx} <- TransactionChain.fetch_transaction_remotely(last_addr, nodes),
       :ok <- Replication.validate_and_store_transaction_chain(last_tx) do
      Logger.info("Resync Success", transaction_type: type)
   end
end
```

lib/archethic/bootstrap.ex
```diff
@@ -206,12 +208,36 @@ defmodule Archethic.Bootstrap do
...
     :persistent_term.put(:archethic_up, :up)
   end
 
+  def resync_network_chain(type_list) do
+    nodes = P2P.authorized_and_available_nodes()
+
+    Task.Supervisor.async_stream_nolink(Archethic.TaskSupervisor, type_list, fn type ->
+      if addr = TransactionChain.list_addresses_by_type(type) |> Enum.at(0) do
+        with {:ok, last_addr} <- TransactionChain.fetch_last_address_remotely(addr, nodes),
+             {:ok, last_tx} <- TransactionChain.fetch_transaction_remotely(last_addr, nodes),
```
samuel-uniris, line 224: maybe you should check if the transaction exists already to avoid the download and this warning

image

lib/archethic/bootstrap.ex
```diff
@@ -206,12 +208,36 @@ defmodule Archethic.Bootstrap do
...
+             :ok <- Replication.validate_and_store_transaction_chain(last_tx) do
+          Logger.info("Enforced Resync of Network Txs: Success", type: type)
+        else
+          error ->
+            Logger.info("Enforced Resync of Network Txs: failure",
```
samuel-uniris, line 229: The error should be in the message and not as metadata

samuel-uniris (changes requested):
Few remarks to improve it

samuel-uniris:
Also please describe more the flow how to test it

lib/archethic/bootstrap.ex
apoorv-2204: In write_transaction DB we are checking for existing txns.
I will redo code with fetch transactions remotely with paging addr

lib/archethic/bootstrap.ex
```diff
@@ -206,12 +208,93 @@ defmodule Archethic.Bootstrap do
...
+  # when no node loaded in memtable
+  @spec resync_network_chain(any, maybe_improper_list) :: :ok
+  def resync_network_chain(_type_list, _nodes = []),
+    do: Logger.info("Enforced Resync of Network Txs: failure, No-Nodes")
+
+  def resync_network_chain(type_list, nodes) when is_list(nodes) do
+    # spawn process for each type of txn to sync
+    Task.Supervisor.async_stream_nolink(Archethic.TaskSupervisor, type_list, fn type ->
+      # acquire a(any) address of type txn, (for querying DB)
+      case TransactionChain.list_addresses_by_type(type) |> Enum.at(0) do
```
Neylix, line 229: Here I think we can reuse the persistent genesis address from PR #495.
So it will not load 2 times the `list_addresses_by_type` function

lib/archethic/bootstrap.ex
```diff
@@ -206,12 +208,93 @@ defmodule Archethic.Bootstrap do
...
+
+          :ok
+
+        addr ->
+          # get last txn locally available
+          {last_addr, _dt} = TransactionChain.get_last_address(addr)
+
+          try do
+            # fetch txs remotely post last_addr
+            TransactionChain.stream_remotely(last_addr, nodes, last_addr)
```
Neylix, line 245: I think you can just use the function `Replication.validate_and_store_transaction_chain()` instead of `Replication.validate_and_store_transaction()` as it already fetch the missing part of the chain.
So to use it you can fetch remotely the last address of the chain and call the function

Neylix (changes requested):

lib/archethic/bootstrap.ex
apoorv-2204: true

lib/archethic/bootstrap.ex
apoorv-2204: No we want to avoid warnings that transaction already exist

Neylix: You just have to compare if the local last address is different than the remote last address. If so you will not write a transaction that already exists

lib/archethic/bootstrap.ex
```diff
@@ -207,13 +209,83 @@ defmodule Archethic.Bootstrap do
...
+
     Sync.publish_end_of_sync()
     SelfRepair.start_scheduler()
 
     :persistent_term.put(:archethic_up, :up)
     Listener.listen()
   end
 
+  # when no node loaded in memtable
+  @spec resync_network_chain(any, maybe_improper_list) :: :ok
```
samuel-uniris, line 222: Why did you mention it as improper list ?
Also the list of types would every time be the same so the first parameter is not required I think

lib/archethic/bootstrap.ex
```diff
@@ -207,13 +209,83 @@ defmodule Archethic.Bootstrap do
...
+  def resync_network_chain(_type_list, _nodes = []),
+    do: Logger.info("Enforced Resync of Network Txs: failure, No-Nodes")
+
+  def resync_network_chain(type_list, nodes) when is_list(nodes) do
+    # spawn process for each type of txn to sync
+    Task.Supervisor.async_stream_nolink(Archethic.TaskSupervisor, type_list, fn type ->
+      # acquire a(any) address of type txn, (for querying DB)
+      case TransactionChain.list_addresses_by_type(type) |> Stream.take(1) |> Enum.at(0) do
+        nil ->
+          Logger.info("Enforced Resync: failure ", transaction_type: type)
```
samuel-uniris, line 232: Why do you qualify it as failure ? If it's pre self repair any new node would have this log

lib/archethic/bootstrap.ex
samuel-uniris: > You just have to compare if the local last address is different than the remote last address. If so you will not write a transaction that already exists

👍

samuel-uniris (changes requested):

lib/archethic/bootstrap.ex
apoorv-2204: > > You just have to compare if the local last address is different than the remote last address. If so you will not write a transaction that already exists
> 
> +1


We are taking last addr and fetching with last addr as paging addr . What ever txs we will receive, will be post last addr, I added it if we ever had got those txns/

lib/archethic/bootstrap.ex
```diff
@@ -212,10 +214,44 @@ defmodule Archethic.Bootstrap do
     Sync.publish_end_of_sync()
     SelfRepair.start_scheduler()
 
+    if P2P.authorized_node?() do
```
Neylix, line 217: Here you should also controll if node is available

lib/archethic/bootstrap.ex
```diff
@@ -212,10 +214,44 @@ defmodule Archethic.Bootstrap do
...
+      resync_network_chain([:oracle, :node_shared_secrets], P2P.authorized_and_available_nodes())
```
Neylix, line 218: This function should be called before sending end of sync notification I think

lib/archethic/bootstrap.ex
```diff
@@ -212,10 +214,44 @@ defmodule Archethic.Bootstrap do
...
+  @spec resync_network_chain(list(atom), list(P2P.Node.t()) | []) :: :ok
+  def resync_network_chain(_type_list, _nodes = []),
+    do: Logger.info("Enforced Resync of Network Txs: failure, No-Nodes")
+
+  def resync_network_chain(type_list, nodes) when is_list(nodes) do
+    # spawn process for each type of txn to sync
+    Task.Supervisor.async_stream_nolink(Archethic.TaskSupervisor, type_list, fn type ->
+      # acquire a(any) address of type txn, (for querying DB)
+      with addr when is_binary(addr) <-
+             TransactionChain.list_addresses_by_type(type) |> Stream.take(1) |> Enum.at(0),
```
Neylix, line 234: Here yous should use the persistent term provided by PR #495 because for oracle, as there is multiple chain, `list_addresses_by_type` will not return you the last chain

Neylix (changes requested):

lib/archethic/bootstrap.ex
apoorv-2204: done!

lib/archethic/bootstrap.ex
apoorv-2204: Done!

lib/archethic/bootstrap.ex
apoorv-2204: noted!

lib/archethic/bootstrap.ex
```diff
@@ -209,13 +211,55 @@ defmodule Archethic.Bootstrap do
...
 
+  @spec resync_network_chain(list(atom), list(P2P.Node.t()) | []) :: :ok
+  def resync_network_chain(_type_list, _nodes = []),
+    do: Logger.info("Enforced Resync of Network Txs: failure, No-Nodes")
+
+  def resync_network_chain(type_list, nodes) when is_list(nodes) do
+    # spawn process for each type of txn to sync
+    Task.Supervisor.async_stream_nolink(Archethic.TaskSupervisor, type_list, fn type ->
+      # acquire a(any) address of type txn, (for querying DB)
+      with addr when is_binary(addr) <- get_addr(type),
```
Neylix, line 234: You should rename `addr` to `genesis_address` to be more understandable. Same for `get_addr()` -> `get_genesis_address()`

lib/archethic/bootstrap.ex
```diff
@@ -209,13 +211,55 @@ defmodule Archethic.Bootstrap do
...
+           {:ok, rem_last_addr} <- TransactionChain.resolve_last_address(addr),
+           false <- TransactionChain.get_last_address(addr) |> elem(0) == rem_last_addr,
+           {:ok, tx} <- TransactionChain.fetch_transaction_remotely(rem_last_addr, nodes),
+           :ok <- Replication.validate_and_store_transaction_chain(tx) do
+        Logger.info("Enforced Resync: Success", transaction_type: type)
+      else
+        e when e in [nil, true, []] ->
```
Neylix, line 241: When it return true it means that you already have the last address in DB, so you should'nt log an "error" but log that resync is not needed

test/archethic/bootstrap_test.exs
```diff
@@ -431,4 +431,87 @@ defmodule Archethic.BootstrapTest do
...
+
+      MockClient
+      |> stub(:send_message, fn
+        _, %GetTransactionChainLength{}, _ ->
+          %TransactionChainLength{length: 4}
+
+        _, %GetLastTransactionAddress{}, _ ->
+          %LastTransactionAddress{address: addr4}
+
+        _, %GetTransactionChain{}, _ ->
```
Neylix, line 497: Here mocking getTransactionChain is useless as it is not called

lib/archethic/bootstrap.ex
```diff
@@ -209,13 +211,55 @@ defmodule Archethic.Bootstrap do
...
+          Logger.debug("Enforced Resync: Unexpected Error Transaction Address: nil/exists",
+            transaction_type: type
+          )
+
```
Neylix, line 245: Here you should return :ok

Neylix (changes requested):
Some updates to do in testfile, I think you have to control if the transaction has been writen in DB.
You can add a MockDB on write_transaction_chain and send to the process a message that to inform we entered in this function.
Then you can add an `assert_receive` in the test to control that you correctly writted the transaction

lib/archethic/bootstrap.ex
```diff
@@ -209,13 +211,60 @@ defmodule Archethic.Bootstrap do
     end
 
     Archethic.Bootstrap.NetworkConstraints.persist_genesis_address()
+
+    if P2P.authorized_node?() && P2P.available_node?() do
+      resync_network_chain([:oracle, :node_shared_secrets], P2P.authorized_and_available_nodes())
```
samuelmanzanera, line 216: You should maybe evict yourself, as you are authorized, so to avoid stale download you should target only remote nodes

Neylix (approved):

PR #527, opened 2022-08-18
Resolve last address conflicts

### LINKED ISSUE #520: Handle conflicts in the resolving of the last transaction addres

To resolve the last transaction address, we can get multiple answers from the nodes and we are not providing some conflict resolutions.
If a node have not the right address in case of out of sync, the resolve is not working and the data become stale.

We should indicate the timestamp of the last transaction address answer so we could manage and take the latest.

samuel-uniris:
# Description

Provide last address conflict resolution and return timestamp in the LastTransactionAddress.
So changes in the chain index database are required. 

Fixes #520 

## Type of change

# How Has This Been Tested?

To test this feature, you start a node. Create a transaction on chain (for example AEWeb)
Then add another node, await the node replicate the transaction in the self-repair and become available (yet not authorized)
Then create another transaction on the chain. (the transaction should be replicated on the first node)
Then request the last update of the chain, the latest value should be retrieved while being called to the both nodes and the conflict resolutions should happen.

# Checklist:


lib/archethic/db/embedded_impl/chain_index.ex
```diff
@@ -401,15 +421,15 @@ defmodule Archethic.DB.EmbeddedImpl.ChainIndex do
...
-        do_search_last_address_until(fd, until, address)
+        do_search_last_address_until(fd, until, {address, timestamp})
       else
         cond do
           timestamp == until ->
             :file.close(fd)
-            address
+            {address, timestamp}
 
           timestamp < until ->
```
Neylix, line 431: Is this control usefull as it is already done before ?

samuel-uniris: L431 ?

Neylix: Yes, we are doing 
```
if timestamp < until do 
...
else
  cond do 
    timestamp < until -> 
      ...
  end
end
```
the second timestamp < until will never be true

samuel-uniris: ok

Neylix (approved):

PR #528, opened 2022-08-18
Fix get transaction chain lookup

samuel-uniris:
# Description

Fix the transaction chain reading to make sure the chain is one expected, so bigger.
Since d244b1dad8562e8ec614c963527d3274529d24df, the chain was downloaded using a diff between local and remote fetch.

The problem being, if a local chain was bigger than the address of the chain requested (pointer) we would include more transaction than expected. And also we might cause database issues, because the pagination address was the latest one, which don't provide much transactions to load.

This fix makes sure we only download the right transactions until the desired transaction's address.

## Type of change

# Checklist:

### DIFF
--- a/lib/archethic.ex
+++ b/lib/archethic.ex
@@ -189,28 +189,53 @@ defmodule Archethic do
       |> P2P.nearest_nodes()
       |> Enum.filter(&Node.locally_available?/1)
 
-    try do
-      {local_chain, paging_address} =
-        case TransactionChain.get_last_local_address(address) do
-          nil -> {[], nil}
-          last_address -> {TransactionChain.get_locally(last_address), last_address}
-        end
-
-      remote_chain =
-        if address != paging_address do
+    # We directly check if the transaction exists and retrieve the genesis
+    # Otherwise we are requesting the genesis address remotly
+    genesis_address =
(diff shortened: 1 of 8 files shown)


apoorv-2204:
LGTM

lib/archethic.ex
```diff
@@ -189,28 +189,53 @@ defmodule Archethic do
...
           address
-          |> TransactionChain.stream_remotely(nodes, paging_address)
-          |> Enum.to_list()
-          |> List.flatten()
+      end
+
+    %{transactions: local_chain, last_address: last_local_address} =
+      genesis_address
+      |> TransactionChain.scan_chain(address)
+      |> Enum.reduce_while(%{transactions: [], last_address: nil}, fn transaction, acc ->
```
Neylix, line 206: Do we need to reduce the chain and check the desired address, as the function `scan_chain` already stop at the desired address ?

samuel-uniris: The reduce was mainly used to get the latest address while leveraging streams , so one item at the time. 
With List.last we would have to traverse the entire list. Also it's not supporting Stream, so we should convert it to list first. 
I didn't find yet a better way

PR #533, opened 2022-08-19
Get token id in the API

### LINKED ISSUE #503: Add unique identification in the token API

As we can create multiple tokens in a chain, we want to provide a way to uniquely identify a token not only based on the genesis address but also with some additional data.

The reason is, clients need an easy and fast way to identify unique assets.

By combining name, symbol, genesis address we can compute a hash and get an ID or digest with all those values.
This will identify the token.

### Additional context

_No response_

samuel-uniris:
# Description

Provide an unique identification of a token for genesis address.
This make possible to host multiple tokens on the same chain while being able to identify them.

Fixes #503 

## Type of change

# Checklist:

### DIFF
--- a/lib/archethic_web/graphql_schema/resolver.ex
+++ b/lib/archethic_web/graphql_schema/resolver.ex
@@ -41,15 +41,40 @@ defmodule ArchethicWeb.GraphQLSchema.Resolver do
     t2 = Task.async(fn -> get_transaction_content(address) end)
 
     with {:ok, {:ok, genesis_address}} <- Task.yield(t1),
-         {:ok, {:ok, transaction_content}} <- Task.yield(t2) do
+         {:ok,
+          {:ok,
+           %{
+             "name" => name,
+             "supply" => supply,
+             "symbol" => symbol,
+             "type" => type,
+             "properties" => properties
+           }}} <- Task.yield(t2) do
+      data_to_digest =
+        case type do
+          "fungible" ->
+            %{genesis_address: Base.encode16(genesis_address), name: name, symbol: symbol}
+
+          "non-fungible" ->
+            %{
+              genesis_address: Base.encode16(genesis_address),
+              name: name,
+              symbol: symbol,
+              properties: properties
+            }
+        end
+
+      token_id = :crypto.hash(:sha256, Jason.encode!(data_to_digest)) |> Base.encode16()
+
       {:ok,
        %{
          genesis: genesis_address,
-         name: Map.get(transaction_content, "name"),
-         supply: Map.get(transaction_content, "supply"),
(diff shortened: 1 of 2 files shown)


lib/archethic_web/graphql_schema/resolver.ex
```diff
@@ -41,15 +41,40 @@ defmodule ArchethicWeb.GraphQLSchema.Resolver do
...
     with {:ok, {:ok, genesis_address}} <- Task.yield(t1),
-         {:ok, {:ok, transaction_content}} <- Task.yield(t2) do
+         {:ok,
+          {:ok,
+           %{
+             "name" => name,
+             "supply" => supply,
+             "symbol" => symbol,
+             "type" => type,
+             "properties" => properties
```
Neylix, line 51: Here properties may not be present and if we create a token without properties it raise a no match error.
I suggest to use something like `Map.get(map, "properties", []) to return an empty list by default

lib/archethic_web/graphql_schema/resolver.ex
```diff
@@ -41,15 +41,40 @@ defmodule ArchethicWeb.GraphQLSchema.Resolver do
...
+           }}} <- Task.yield(t2) do
+      data_to_digest =
+        case type do
+          "fungible" ->
```
Neylix, line 55: I think we should do in the other way, add the properties in the hash of fungible tokens and not in the non-fungible token as non_fungible token will probably have different properties

lib/archethic_web/graphql_schema/transaction_type.ex
```diff
@@ -239,7 +239,8 @@ defmodule ArchethicWeb.GraphQLSchema.TransactionType do
     field(:symbol, :string)
     field(:supply, :integer)
     field(:type, :string)
-    field(:properties, list_of(:tokenProperty))
+    field(:properties, list_of(list_of(:tokenProperty)))
+    field(:id, :string)
```
Neylix, line 243: You should also add the id description in the text above

Neylix (changes requested):

lib/archethic_web/graphql_schema/resolver.ex
samuel-uniris: Ok

lib/archethic_web/graphql_schema/resolver.ex
samuel-uniris: Oh yes you re right ! I inverted the logic 😅

lib/archethic_web/graphql_schema/transaction_type.ex
samuel-uniris: Ok

Neylix (approved):

PR #536, opened 2022-08-22
Verify if transaction exists before validation #502

### LINKED ISSUE #502: Verify if transaction exists before validation

Currently, the welcome node doesn't verify if a transaction exists before to send the validation requests.
Even if the validation nodes get the errors where the tx already exists, to avoid wasteful resources, we can prevent upfront the tx validation.

We can add a verification in the API when we receive a transaction to submit to the network for validation to check if the transaction exists.
This will imply to request from the storage node if the transaction exists.
We can use the `GetTransactionSummary` P2P message to check if the transaction exists or not, returning `'NotFound`in that case.

### Additional context

_No response_

prix-uniris:
# Description
Feature Addition, Check on Transaction Existence before sending it for mining.

Fixes #502 

## Type of change

# How Has This Been Tested?

Manually Tested

# Checklist:

### DIFF
--- a/lib/archethic_web/controllers/api/transaction_controller.ex
+++ b/lib/archethic_web/controllers/api/transaction_controller.ex
@@ -40,22 +40,7 @@ defmodule ArchethicWeb.API.TransactionController do
             |> json(%{status: "error - transaction is invalid"})
 
           {:error, :transaction_not_exists} ->
-            case Archethic.send_new_transaction(tx) do
-              :ok ->
-                TransactionSubscriber.register(tx.address, System.monotonic_time())
-
-                conn
-                |> put_status(201)
-                |> json(%{
-                  transaction_address: Base.encode16(tx.address),
-                  status: "pending"
-                })
-
-              {:error, :network_issue} ->
-                conn
-                |> put_status(422)
-                |> json(%{status: "error - transaction may be invalid"})
-            end
+            send_transaction(conn, tx)
         end
 
       changeset ->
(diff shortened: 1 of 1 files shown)


lib/archethic_web/controllers/api/transaction_controller.ex
```diff
@@ -25,21 +25,22 @@ defmodule ArchethicWeb.API.TransactionController do
...
-            })
+            |> put_status(422)
+            |> json(%{status: "error - network issue"})
 
-          {:error, :network_issue} ->
+          {:error, :transaction_invalid} ->
             conn
             |> put_status(422)
-            |> json(%{status: "error - transaction may be invalid"})
+            |> json(%{status: "error - transaction is invalid"})
```
samuel-uniris, line 40: I think we should allow the re-send of invalid transaction

lib/archethic_web/controllers/api/transaction_controller.ex
```diff
@@ -25,21 +25,22 @@ defmodule ArchethicWeb.API.TransactionController do
           |> TransactionPayload.to_map()
           |> Transaction.cast()
 
-        case Archethic.send_new_transaction(tx) do
-          :ok ->
-            TransactionSubscriber.register(tx.address, System.monotonic_time())
+        case Archethic.search_transaction(tx.address) do
```
samuel-uniris, line 28: By using search transaction we might fetch an entire transaction if exists. 
We should do a lightweight verification like using GetTransactionSummary message (acting like a header)

samuel-uniris (changes requested):

lib/archethic_web/controllers/api/transaction_controller.ex
prix-uniris: Okay makes sense, for this then we should query storage nodes for the transaction summary. 
We can skip the quorom read if we are focusing on lightweight fast responses.

samuel-uniris: I don't think we still need the quorum read, as we don't to avoid stale or inconsistent data.
The lightweight would be to reduce the payload to return.

lib/archethic_web/controllers/api/transaction_controller.ex
```diff
@@ -25,21 +38,27 @@ defmodule ArchethicWeb.API.TransactionController do
...
-              status: "pending"
-            })
-
-          {:error, :network_issue} ->
-            conn
-            |> put_status(422)
-            |> json(%{status: "error - transaction may be invalid"})
+        tx_address = tx.address
+
+        storage_nodes = Election.chain_storage_nodes(tx_address, P2P.available_nodes())
```
samuel-uniris, line 43: I think you should use authorized and available nodes to make it consistent with the shard election.

prix-uniris: Yes that would be better

samuel-uniris (approved):

PR #538, opened 2022-08-23
Prevent scheduling network tx

### LINKED ISSUE #420: Prevent scheduled transaction to be sent in the same time by multiple nodes

We are using some self-triggered transactions based on scheduling like cronjob.
Hence, we should avoid two nodes to send the same transaction with few milliseconds of interval.
This can happen if the transaction is validated too fastly, creating some race condition.

The solution would be to either:
- store the timestamp of the last transaction in the scheduler state (coming from the replication) to not relaunch if timestamp > last scheduler.
- manage through FSM using event notifications

### Additional context

```mermaid
stateDiagram-v2
    [*] --> IDLE
    IDLE --> CONFIRMED: replication confirmation
    CONFIRMED --> IDLE: schedule new cron
    IDLE --> SCHEDULED: recv cron event
    SCHEDULED --> SCHEDULED: replication confirmation
    SCHEDULED --> CONFIRMED
 ```

### PR DESCRIPTION
**Pull request by samuel-uniris**
# Description

This should prevent scheduler to trigger transactions too fastly computed before the scheduling time.
Using FSM we are ensuring to enter in the state of scheduling only when the tx is replicated or after timeouts.

Fixes #420 

## Type of change

# Checklist:

### DIFF
--- a/lib/archethic/reward.ex
+++ b/lib/archethic/reward.ex
@@ -55,8 +55,9 @@ defmodule Archethic.Reward do
     ...>  }
     ...> } = Reward.new_rewards_mint(2_000_000_000)
   """
-  @spec new_rewards_mint(amount :: non_neg_integer()) :: Transaction.t()
-  def new_rewards_mint(amount) do
+  @spec new_rewards_mint(amount :: non_neg_integer(), index :: non_neg_integer()) ::
+          Transaction.t()
+  def new_rewards_mint(amount, index) do
     data = %TransactionData{
       code: """
         condition inherit: [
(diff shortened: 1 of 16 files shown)

### CODE REVIEW

File: lib/archethic/reward.ex [hunk 1]
```diff
@@ -55,8 +55,9 @@ defmodule Archethic.Reward do
     ...>  }
     ...> } = Reward.new_rewards_mint(2_000_000_000)
```
Neylix, line 56: You just forgot to update the doctest function which is crashing tests

Neylix (changes requested):
It works well !
Just need to update doctest and it's ok for me

lib/archethic/reward.ex
samuel-uniris: ok

Neylix (approved):

PR #545, opened 2022-08-26
Fix replication transaction chain ordering writing

ghost:
# Description

Fix the replication transaction chain (due to regression)

## Type of change

# Checklist:

### DIFF
--- a/lib/archethic/replication.ex
+++ b/lib/archethic/replication.ex
@@ -115,6 +115,14 @@ defmodule Archethic.Replication do
 
           TransactionChain.write_transaction(tx)
 
+          :telemetry.execute(
+            [:archethic, :replication, :full_write],
+            %{
+              duration: System.monotonic_time() - start_time
+            },
+            %{role: :chain}
+          )
+
           :ok
 
         {:error, reason} ->
--- a/lib/archethic/replication.ex
+++ b/lib/archethic/replication.ex
@@ -90,8 +90,6 @@ defmodule Archethic.Replication do
              Enum.to_list(inputs)
            ) do
         :ok ->
-          # Store the previous and the new one, as being the reference for next transactions 
-          :ok = TransactionChain.write(Enum.filter([previous_tx, tx], & &1))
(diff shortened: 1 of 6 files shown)


lib/archethic/replication.ex
```diff
@@ -108,17 +106,14 @@ defmodule Archethic.Replication do
...
-            end)
-          end
+          # Stream the insertion of the chain
+          tx
+          |> stream_previous_chain()
+          |> Stream.reject(&Enum.empty?/1)
+          |> Stream.each(&TransactionChain.write/1)
+          |> Stream.run()
+
+          TransactionChain.write(Enum.filter([previous_tx, tx], & &1))
```
Neylix, line 116: writing previous transaction is useless here as it is done before by stream previous chain

Neylix (approved):

lib/archethic/replication.ex
samuelmanzanera: 👍

lib/archethic/replication.ex
```diff
@@ -108,17 +106,14 @@ defmodule Archethic.Replication do
...
-            end)
-          end
+          # Stream the insertion of the chain
+          tx
+          |> stream_previous_chain()
+          |> Stream.reject(&Enum.empty?/1)
+          |> Stream.each(&TransactionChain.write/1)
+          |> Stream.run()
+
+          TransactionChain.write_transaction(tx)
```
Neylix, line 116: telemetry should be calculated after DB writing no ?

Neylix (commented):
There is also a test error.
in sync test a mockDB on `write_transaction_chain` is defined to send an event, but we don't always use `write_transaction_chain`. In the test case it should be `write_transaction`

lib/archethic/replication.ex
samuelmanzanera: in fact this telemetry it's about the replication validation.
But I can add one for the full write

Neylix (approved):

PR #549, opened 2022-08-29
TokenAddress naming #525

### LINKED ISSUE #525: TokenAddress naming

In TokenTransfer, the token's address field is "token"
In TransactionInput, the token's address field is "tokenAddress"
Why ?

Could you uniformise with "tokenAddress" ?

_No response_

prix-uniris:
# Description
API Feature Improvement, In TokenTransfer naming changed to tokenAddress

Fixes #525  #556

## Type of change

Might Require Changes in libjs transaction_builder json conversion key from token to tokenAddress.

# How Has This Been Tested?
Mix Testing 
Manual Testing

# Checklist:


samuelmanzanera (commented):
I think #525 goes even beyond APIs and I think should be taken for entire codebase, to create  more consistency.

prix-uniris:
> I think #525 goes even beyond APIs and I think should be taken for entire codebase, to create more consistency.

Which files needed to be changed, as when I was changing, There are many internal function which uses :token as their key and that makes sense internally as that key usually keeps a address in the structure. If we change the key in the ledger transfer struct it wouldn't make much sense. Your thoughts.

If we need to change then where should we change it :
E.g - Like here
https://github.com/archethic-foundation/archethic-node/blob/58491b69f566350dfae7b49690816f063faf28ca/lib/archethic/transaction_chain/transaction/data/ledger/token/transfer.ex#L5

Neylix:
You can either update in token transfer but it will need to update also other dependent files
https://github.com/archethic-foundation/archethic-node/blob/58491b69f566350dfae7b49690816f063faf28ca/lib/archethic/transaction_chain/transaction/data/ledger/token/transfer.ex#L5

or just for the graphql API here 
https://github.com/archethic-foundation/archethic-node/blob/58491b69f566350dfae7b49690816f063faf28ca/lib/archethic/transaction_chain/transaction/data/ledger/token/transfer.ex#L155

lib/archethic_web/controllers/api/schema/token_ledger.ex
```diff
@@ -10,7 +10,7 @@ defmodule ArchethicWeb.API.Schema.TokenLedger do
     embeds_many :transfers, Transfer do
       field(:to, Address)
       field(:amount, :integer)
-      field(:token, Address)
+      field(:tokenAddress, Address)
```
samuelmanzanera, line 13: you should the `token_address` case

lib/archethic_web/controllers/api/schema/token_ledger.ex
```diff
@@ -10,7 +10,7 @@ defmodule ArchethicWeb.API.Schema.TokenLedger do
...
       field(:token_id, :integer)
```
samuelmanzanera, line 14: We should unify the `token_id` in camelCase for the API, as it's will be translated in snake case after casting

prix-uniris: yup that makes more senser, I was also thinking the same. Since atom parse does change it for node internal functions.

samuelmanzanera (approved):

lib/archethic/account/mem_tables_loader.ex
```diff
@@ -130,6 +130,9 @@ defmodule Archethic.Account.MemTablesLoader do
 
       {{to, _token_address}, utxo} ->
         TokenLedger.add_unspent_output(to, utxo, timestamp)
+
+      {{to, _token_address, _token_id}, utxo} ->
```
samuelmanzanera, line 134: I think the best would be to add the token_id to 0 by default so you would avoid to have 2 pattern matching.

prix-uniris: I think it shouldn't cause issue, as we are checking for 2 values and 3 values in struct or map. any example code to reproduce conflict.

samuelmanzanera: You have to change is also in the aggregate function

prix-uniris: I think it already have that pattern match,
https://github.com/prix-uniris/archethic-node/blob/3b0ad7f106fac54709076de3676f34808f0a4b7b/lib/archethic/account/mem_tables_loader.ex#L204
Where do i have to modify?

samuelmanzanera: There I think https://github.com/prix-uniris/archethic-node/blob/3b0ad7f106fac54709076de3676f34808f0a4b7b/lib/archethic/account/mem_tables_loader.ex#L223

prix-uniris: Alright add 0 as part of key for the map, like this
 L#223 {to, token_address, 0},
 Like this

apoorv-2204: Or may be you can drop it, we have an another pr 
https://github.com/archethic-foundation/archethic-node/pull/557
, with test cases, which addresses this issue , more clearly.

samuelmanzanera: While Prince is on it, it will be better to handle it and clsoe the #557

lib/archethic/account/mem_tables_loader.ex
```diff
@@ -130,6 +130,9 @@ defmodule Archethic.Account.MemTablesLoader do
 
       {{to, _token_address}, utxo} ->
         TokenLedger.add_unspent_output(to, utxo, timestamp)
+
+      {{to, _token_address, 0}, utxo} ->
```
samuelmanzanera, line 134: You should remove the previous pattern matching and replace `0` by `_token_id`

prix-uniris: Yes that makes more sense, I have done that.

samuelmanzanera (approved):

Issue #543, opened 2022-08-25
Prevent schedulers to be launched before end of bootstrap


ghost:
### Describe the solution you'd like

We can leverage persistent term `archethic_up` in the init and during node_update to start the schedulers

internet-zero:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Product-and-Design-Workspace-61efe71b9083a90018250d89/issues/archethic-foundation/archethic-node/543?planning-poker) with Zenhub @samuel-uniris @Neylix @apoorv-2204 @prix-uniris

samuelmanzanera:
Closed by #552

PR #552, opened 2022-08-30
Start Schedulers Post Bootstrap  #543

apoorv-2204:
# Description

Scheduler must start only after post Bootstrap, So that no new next txn are sent during bootstrap .  
During crash scheduler must read arch-ethic_up  before starting.

Affected Schedulers
- Reward
- Oracle
- Node Renewal
- Slot timer

Fixes # (#543)

## Type of change

# How Has This Been Tested?

- [x] Unit Test cases. 
- [x] Distributed Testing.

Logs were checked to ensure the scheduler started post Bootstrap Finished message.

Screenshot_20220906_161508

explicitly killing a scheduler
Screenshot_20220906_162836

## How to test?

Use Logs to confirm that schedulers started post bootstrap finished message.

# Checklist:

### DIFF
--- a/lib/archethic/bootstrap.ex
+++ b/lib/archethic/bootstrap.ex
@@ -116,6 +116,8 @@ defmodule Archethic.Bootstrap do
     else
       post_bootstrap(sync?: false)
     end
+
+    Logger.info("Bootstrapping finished!")
   end
 
   defp should_bootstrap?(_ip, _port, _http_port, _, nil), do: true
@@ -195,8 +197,6 @@ defmodule Archethic.Bootstrap do
         end
       end
     end
-
-    Logger.info("Bootstrapping finished!")
   end
 
   defp post_bootstrap(opts) do
@@ -213,6 +213,7 @@ defmodule Archethic.Bootstrap do
     SelfRepair.start_scheduler()
 
     :persistent_term.put(:archethic_up, :up)
+    Archethic.PubSub.notify_node_up()
     Listener.listen()
   end
 
--- a/lib/archethic/oracle_chain/scheduler.ex
+++ b/lib/archethic/oracle_chain/scheduler.ex
@@ -49,38 +49,63 @@ defmodule Archethic.OracleChain.Scheduler do
     polling_interval = Keyword.fetch!(args, :polling_interval)
     summary_interval = Keyword.fetch!(args, :summary_interval)
 
+    state_data =
+      %{}
+      |> Map.put(:polling_interval, polling_interval)
+      |> Map.put(:summary_interval, summary_interval)
+
+    case :persistent_term.get(:archethic_up, nil) do
+      nil ->
+        # node still bootstrapping , wait for it to finish Bootstrap
+        Logger.info(" Oracle Scheduler: Waiting for Node to complete Bootstrap. ")
+
+        PubSub.register_to_node_up()
+
+        {:ok, :idle, state_data}
+
+      # wait for node UP
+      :up ->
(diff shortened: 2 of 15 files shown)


lib/archethic/account/mem_tables_loader.ex
```diff
@@ -130,6 +130,9 @@ defmodule Archethic.Account.MemTablesLoader do
 
       {{to, _token_address}, utxo} ->
         TokenLedger.add_unspent_output(to, utxo, timestamp)
+
+      {{to, _token_address, _token_id}, utxo} ->
```
samuelmanzanera, line 134: Why do you have the two cases ?

lib/archethic/bootstrap.ex
```diff
@@ -213,6 +213,7 @@ defmodule Archethic.Bootstrap do
     SelfRepair.start_scheduler()
 
     :persistent_term.put(:archethic_up, :up)
+    Archethic.PubSub.notify_node_up(:persistent_term.get(:archethic_up, nil))
```
samuelmanzanera, line 216: Why are you passing parameter to this function ?
If you are persistent_term.put(:archethic_up, :up), it will be everytime up.
So you should just send message :node_up

lib/archethic/oracle_chain/scheduler.ex
```diff
@@ -64,8 +84,10 @@ defmodule Archethic.OracleChain.Scheduler do
         index = chain_size(summary_date)
         Logger.info("Oracle Scheduler: Scheduled during init - (index: #{index})")
 
-        {:ok, :idle,
+        {:idle,
          %{
+           # server data must have old key
+           args: nil,
```
samuelmanzanera, line 90: what is this variable ?

samuelmanzanera (changes requested):

lib/archethic/account/mem_tables_loader.ex
apoorv-2204: we use this two different cases in aggregating the movements. A map is used to aggregate , where key is 
`to , address, tokenid`  for nft 
and f
` to, token_address` for fungible tokens.
Bug was reported by Prince, it fixes that.

lib/archethic/oracle_chain/scheduler.ex
apoorv-2204: With introduction, of PubSub.register_for_node_up, process_data modification were required.
 If  `archethic_up` is `nil` we pass the args of `init(args)` into the process_data of FSM as `%{args: args}`,
Where it waits for `:node_up` as soon as message receives, start scheduler by reading data which has key :args.

samuelmanzanera: You can set the interval in the state even without archethic_up event. This should be common.

lib/archethic/bootstrap.ex
```diff
@@ -213,7 +213,7 @@ defmodule Archethic.Bootstrap do
     SelfRepair.start_scheduler()
 
     :persistent_term.put(:archethic_up, :up)
-    Archethic.PubSub.notify_node_up(:persistent_term.get(:archethic_up, nil))
+    Archethic.PubSub.notify_node_up(:up)
```
samuelmanzanera, line 216: No need to specify the `up` information, we want just want notification it's all

apoorv-2204:
Testing: we looked for Bootstrap message and post that scheduler must start.
Screenshot_20220905_120007

if Node is still bootstrapping scheduler/timer must wait .
Screenshot_20220905_120354

Neylix (approved):

PR #553, opened 2022-09-01
Archethic Node cron scheduler wrongly configured #540

### LINKED ISSUE #540: Restrict extended cron expression in smart contract.

See https://github.com/archethic-foundation/bug-bounty/issues/40

I propose to add a parameter in the `Utils.time_offset` to mark the usage of extended mode or not.

prix-uniris:
# Description
Before when supplied with old Cron Specification, the extended mode was triggering SC with per seconds which is not the intended effect which is to send per minute

Fixes #540 

## Type of change

# How Has This Been Tested?

Manually Tested

# Checklist:
Screenshot from 2022-09-01 20-05-59

### DIFF
--- a/lib/archethic/contracts/interpreter.ex
+++ b/lib/archethic/contracts/interpreter.ex
@@ -822,10 +822,22 @@ defmodule Archethic.Contracts.Interpreter do
         "interval" ->
           [{{:atom, "at"}, interval}] = opts
 
+          # Extended Mode Signifies whether use CRON Extended Mode or Not while running SC Intervals
+          extended_mode =
+            case length(String.split(interval, " ", trim: true)) do
+              5 -> false
+              _ -> true
+            end
+
           Map.update!(
             acc,
             :contract,
-            &Contract.add_trigger(&1, :interval, [at: interval], actions)
+            &Contract.add_trigger(
+              &1,
+              :interval,
+              [at: interval, extended_mode: extended_mode],
+              actions
+            )
           )
 
         "transaction" ->
--- a/lib/archethic/contracts/interpreter.ex
+++ b/lib/archethic/contracts/interpreter.ex
@@ -822,20 +822,13 @@ defmodule Archethic.Contracts.Interpreter do
         "interval" ->
           [{{:atom, "at"}, interval}] = opts
 
-          # Extended Mode Signifies whether use CRON Extended Mode or Not while running SC Intervals
-          extended_mode =
-            case length(String.split(interval, " ", trim: true)) do
-              5 -> false
-              _ -> true
-            end
-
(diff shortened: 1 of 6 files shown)


lib/archethic/contracts/interpreter.ex
```diff
@@ -822,10 +822,22 @@ defmodule Archethic.Contracts.Interpreter do
...
+            case length(String.split(interval, " ", trim: true)) do
+              5 -> false
+              _ -> true
+            end
+
           Map.update!(
             acc,
             :contract,
-            &Contract.add_trigger(&1, :interval, [at: interval], actions)
+            &Contract.add_trigger(
```
samuelmanzanera, line 835: We want to avoid SC to use extended as it's irrelevant in the context of blockchain networks. 
No tx would be able to be validated in 1s with a significant number of nodes + network latency.
So I think we should pass directly in the worker the flag as false

lib/archethic/contracts/worker.ex
```diff
@@ -240,8 +240,14 @@ defmodule Archethic.Contracts.Worker do
...
 
-  defp schedule_trigger(trigger = %Trigger{type: :interval, opts: [at: interval]}) do
-    Process.send_after(self(), trigger, Utils.time_offset(interval) * 1000)
+  defp schedule_trigger(
+         trigger = %Trigger{type: :interval, opts: [at: interval, extended_mode: extended_mode]}
+       ) do
+    Process.send_after(
+      self(),
+      trigger,
+      Utils.time_offset(interval, DateTime.utc_now(), extended_mode) * 1000
```
samuelmanzanera, line 249: We should pass it as false

samuelmanzanera (changes requested):

lib/archethic/contracts/interpreter.ex
prix-uniris: Oh okay, I also thought of the same in start but thought it is a added feature so implemented that way. Will make the changes to do simple no additional opts. 

Also there are not many checks on SC cron phrase, is it fine or will we add more restrict checks later

lib/archethic/contracts/worker.ex
prix-uniris: Yup will do it

lib/archethic/contracts/interpreter.ex
samuelmanzanera: There is check
https://github.com/archethic-foundation/archethic-node/blob/58491b69f566350dfae7b49690816f063faf28ca/lib/archethic/contracts/interpreter.ex#L995-L1003

samuelmanzanera (approved):

PR #559, opened 2022-09-06
Drop transaction in beacon chains for slot and summaries

### LINKED ISSUE #523: Drop transaction in beacon chains for slot and summaries

From now we are using transaction chains for the beacon chain itself.
However this brings few drawbacks and problems:
- TPM latency issue: too much origin signature to perform (ie 256 signature in the same time during beacon summary)
- Waste of data stored (useless transaction_data serialization)
- Too long to fetch the transaction chain summary
- Not global view of the slots (only one transaction passed)

We can leverage beacon summary directly into the cache:
- Aggregate multiple slots for a given subset and time
- Faster serialization and retrieval in the beacon chain explorer (direct hit of the summary cache)
- No need to have signatures for the beacon , as the summary is stored locally and not with consensus
- Need to add a new message to get the summary from the cache
- Write to disk to persist cache in case of failure
- Write to disk the summary in a new files/folders

### Additional context

_No response_

Neylix:
# Description

This change improve scalability for the beacon chain :
- Slots are not anymore a transaction, slots are sent directly to beacon nodes by a simple message
- All slots received are added in the summary cach (only 1 was added before)
- During summary, all slots are aggregated to make a real view of all node for a slot (only 1 view before)
- Beacon summaries are not anymore a transaction, summaries are directly written in DB

- To ensure persistent data in case of node crash, when a slot is received, it is written in a DB file
- When node restart, summary cache is restored from this file

- A new message is created to retrieve current slots for beacon explorer

Fixes #523 

## Type of change

# How Has This Been Tested?

Add tests in test files

Run multiple node, stop them, restart them -> Beacon summaries are correctly retrieved
Navigate in beacon explorer, all transactions are displayed

# Checklist:


lib/archethic/beacon_chain/update.ex
```diff
@@ -70,7 +70,7 @@ defmodule Archethic.BeaconChain.Update do
   end
 
   def handle_cast({:unsubscribe, node_public_key}, state) do
-    new_state = Map.delete(state, node_public_key)
+    new_state = if node_public_key, do: Map.delete(state, node_public_key), else: Map.new()
```
samuelmanzanera, line 73: Why do you want to clean if the public key is not the same ?
If some malicious actors send a P2P message to unsubscribe it could remove subscription for all others clients (nodes)

lib/archethic/beacon_chain/update.ex
```diff
@@ -26,8 +26,8 @@ defmodule Archethic.BeaconChain.Update do
   @doc """
   Unsubscribe for Beacon update to a node
   """
-  @spec unsubscribe(binary()) :: :ok
-  def unsubscribe(node_public_key) do
+  @spec unsubscribe(binary() | nil) :: :ok
```
samuelmanzanera, line 29: Why do you accept nil ?

lib/archethic/beacon_chain/subset/summary_cache.ex
```diff
@@ -41,6 +63,49 @@ defmodule Archethic.BeaconChain.Subset.SummaryCache do
...
+      deserialize(content, [])
+      |> Enum.each(fn {summary_time, subset, slot} ->
+        if summary_time == next_summary_time, do: true = :ets.insert(@table_name, {subset, slot})
+      end)
+    else
+      :ok
+    end
+  end
+
+  defp serialize(subset, slot = %Slot{slot_time: slot_time}) do
```
samuelmanzanera, line 93: I think the subset can be omitted as the slot contains the subset information

lib/archethic/beacon_chain.ex
```diff
@@ -202,17 +165,33 @@ defmodule Archethic.BeaconChain do
...
         {:error, :not_found}
     end
   end
 
+  @doc """
+  Get all slots of a subset from summary cache and return unique transaction summaries
+  """
+  @spec get_summary_slots(binary()) :: list(TransactionSummary.t())
+  def get_summary_slots(subset) when is_binary(subset) do
+    SummaryCache.get_current_slots(subset)
```
samuelmanzanera, line 183: If you want leverage completely streams, an idea would be to read as enumerable the ets table instead of lookup. Because lookup will load everything in memory.
For this, you can scan the ets table record and fill up the stream.
In that sense you would have a complete stream of the beacon slots

PR #570, opened 2022-09-14
Support UTXO & Tx Inputs streaming

### LINKED ISSUE #568: Unspent outputs fetch causes timeouts

During the transaction validation, there is two blocking RPC calls: get previous transaction and list unspent outputs.
While the implementation is simple and works well in good connectivity.

For the unspent outputs, we might faced timeouts due to a problem of connectivity or significant size of UTXOs to retrieve.
Because we are using a quorum read, we are waiting the timeout of the bad connected node to make a decision on the conflict resolution. This is causing task timeout.

Hence the design must be change during the transaction validation to be tolerant of the failure and latency in this aspect as well.

To resolve this, we can leverage either:
- async (non blocking) communication to make sure to have the right time to get the UTXOs.
The GetUnspentOutputs can be async, to notify the ack of the message while sending the UTXOs separately 
- batch: like the fetch of transaction chains we can get unspent outputs by streaming them sequentially

samuelmanzanera:
# Description

This PR provides a way to stream UTXOs and Transaction inputs to avoid timeouts in case of many data to download.
This would be used in the mining, UI and API.

Fixes #568 

## Type of change

# Checklist:


lib/archethic_web/graphql_schema/resolver.ex
```diff
@@ -122,13 +122,19 @@ defmodule ArchethicWeb.GraphQLSchema.Resolver do
...
-        {:ok, Enum.map(inputs, &TransactionInput.to_map/1)}
-
-      {:error, _} = e ->
-        e
+  def get_inputs(address, paging_offset \\ 0, limit \\ 0) do
+    inputs =
+      address
+      |> Archethic.get_transaction_inputs(paging_offset)
+      |> Enum.map(&TransactionInput.to_map/1)
+      |> Enum.sort_by(& &1.timestamp, {:desc, DateTime})
```
Neylix, line 130: Sorting by timestamp is a bit confusing there I think, or it should be the same sorting in the GettransationInputs message. Here the problem:
I request all transaction inputs and I get 
[1, 2 ,3 , 4, 5] (1 is the more recent)
Then I want to get inputs from offset 2, I expected to have [3, 4, 5]
but I get [1, 2, 3] because offset in the message is based on an asc sort and not desc sort

lib/archethic_web/graphql_schema/resolver.ex
```diff
@@ -122,13 +122,19 @@ defmodule ArchethicWeb.GraphQLSchema.Resolver do
...
+
+    case limit do
+      0 ->
+        {:ok, inputs}
+
+      limit ->
```
Neylix, line 136: Isn't it a better practice to put the limit also in the message itself, so the nodes returns us only the data we need and not all the data ?

Neylix (changes requested):

Neylix (approved):

PR #573, opened 2022-09-15
Add last transactions feed on the homepage #404

### LINKED ISSUE #404: Add last transactions feed on the homepage

We want to display the latest transaction in the network as a component, to have something like 10 last transactions.

A new live view component under the beacon chain stats should be added using the beacon live features.

### Additional context

_No response_

prix-uniris:
# Description

Fixes #404

## Type of change

# How Has This Been Tested?

Manual Tested

# Checklist:

### DIFF
--- a/lib/archethic_web/live/explorer/cache.ex
+++ b/lib/archethic_web/live/explorer/cache.ex
@@ -57,7 +57,7 @@ defmodule ArchethicWeb.ExplorerLive.LastTenTransactionCache do
         with result <- resolver.() do
           Logger.debug("Caching results for last 10 Transactions")
 
-          Enum.map(result, fn %TransactionSummary{address: address} = txn ->
+          Enum.each(result, fn %TransactionSummary{address: address} = txn ->
             :ets.insert(@table, {address, txn})
             txn
           end)
@@ -86,7 +86,7 @@ defmodule ArchethicWeb.ExplorerLive.LastTenTransactionCache do
         with result <- resolver.() do
           Logger.debug("Caching results for last 10 Transactions")
 
-          Enum.map(result, fn %TransactionSummary{address: address} = txn ->
+          Enum.each(result, fn %TransactionSummary{address: address} = txn ->
             :ets.insert(@table, {address, txn})
             txn
           end)
--- a/lib/archethic_web/live/explorer/cache.ex
+++ b/lib/archethic_web/live/explorer/cache.ex
@@ -1,0 +1,113 @@
+defmodule ArchethicWeb.ExplorerLive.LastTenTransactionCache do
+  alias Archethic.TransactionChain.TransactionSummary
+
+  @table :last_ten_transactions
+
+  @moduledoc false
+  use GenServer
+  require Logger
+
+  def start_link(opts) do
+    GenServer.start_link(__MODULE__, opts, name: __MODULE__)
+  end
+
+  def init(_args) do
+    :ets.new(@table, [:set, :public, :named_table])
+
+    {:ok, []}
+  end
+
+  @doc """
+  Retrieve a value from cache
+  """
+  def get(address) do
+    GenServer.call(__MODULE__, {:get, address})
+  end
+
+  def get_all() do
+    GenServer.call(__MODULE__, :get_all)
+  end
+
+  def handle_call(:get_all, _, state) do
+    result = :ets.match(@table, :"$1")
+    {:reply, result, state}
+  end
+
+  def handle_call({:get, address}, _, state) do
+    Logger.debug("Fetching from cache")
+
+    result =
+      case :ets.lookup(@table, address) do
+        [{^address, value}] ->
+          {value}
+
+        _ ->
+          nil
+      end
+
+    {:reply, result, state}
+  end
+
+  @doc """
+  Runs a piece of code if not already cached
+  """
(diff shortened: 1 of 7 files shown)


lib/archethic_web/live/last_ten_transactions.ex
```diff
@@ -0,0 +1,157 @@
+defmodule ArchethicWeb.LastTenTransactionsComponent do
```
samuelmanzanera, line 1: I think the name of module could be more generic like `TopTransactionsComponent`

lib/archethic_web/live/explorer/cache.ex
```diff
@@ -0,0 +1,113 @@
+defmodule ArchethicWeb.ExplorerLive.LastTenTransactionCache do
```
samuelmanzanera, line 1: I think the name could be more generic like `TopTransactionCache`

lib/archethic_web/live/last_ten_transactions.ex
```diff
@@ -0,0 +1,157 @@
...
+    previous_time = BeaconChain.previous_summary_time(time)
+    txns = txns ++ list_transactions_from_summary(previous_time)
+    fetch_previous_dates(previous_time, txns)
+  end
+
+  defp fetch_previous_dates(_time = %DateTime{}, txns) when length(txns) >= 10 do
+    txns
+  end
+
+  defp fetch_last_ten_transactions() do
```
samuelmanzanera, line 89: the same here about naming.
the module or the function could define arguments to determine the size of list instead.

lib/archethic_web/live/last_ten_transactions.ex
```diff
@@ -0,0 +1,157 @@
...
+    txns = list_transactions_from_current_slots()
+
+    if length(txns) < 10 do
+      fetch_previous_dates(DateTime.utc_now(), txns)
+    else
+      txns
+      |> Enum.sort_by(& &1.timestamp, {:desc, DateTime})
```
samuelmanzanera, line 96: the transactions are already sorted in `list_transaction_from_current_slots`

samuelmanzanera (changes requested):

lib/archethic_web/live/last_ten_transactions.ex
prix-uniris: Yes that would be better.

lib/archethic_web/live/last_ten_transactions.ex
prix-uniris: thanks, nice catch

samuelmanzanera:
You should remove as well TopTransactionLive.ex I think, as it have been replacedd by the new module you defined.

lib/archethic_web/live/top_transactions_component.ex
```diff
@@ -0,0 +1,156 @@
...
+    {:ok, socket}
+  end
+
+  def update(
+        %{transaction: transaction} = _assigns,
+        socket
+      )
+      when not is_nil(transaction) do
+    {:ok, transactions} =
+      TopTransactionsCache.resolve_put(transaction, fn ->
```
samuelmanzanera, line 33: Why are you fetching the list if you are receiving a new transaction, this part of fetching should be at the initialization or preloading of the component.

lib/archethic_web/live/explorer/cache.ex
```diff
@@ -0,0 +1,113 @@
...
+  Update the Local Cache Transaction and pop last txn
+  """
+  def resolve_put(%TransactionSummary{address: address} = txn, resolver)
+      when is_function(resolver, 0) do
+    case get_all() do
+      [] ->
+        with result <- resolver.() do
+          Logger.debug("Caching results for last 10 Transactions")
+
+          Enum.each(result, fn %TransactionSummary{address: address} = txn ->
```
samuelmanzanera, line 89: I don't get this part.
because you are naming the same way the parameter of the Enum.each fn and the parameter of the resolve_put fn, we lost track of which information will be inserted.

samuelmanzanera: Instead you could like cycling buffer:

```elixir
defmodule CyclingBuffer do
  use GenServer
  def start_link do
    GenServer.start_link(__MODULE__, [])
  end

  def init(_) do
    :ets.new(:buffer, [:ordered_set, :named_table, :public])
    :ets.insert(:buffer, {:size, 10, 0})
    {:ok, %{}}
  end

  def push( data) do
    [{_, size, _count}] = :ets.lookup(:buffer, :size)
    [new_index] = :ets.update_counter(:buffer, :size, [{3, 1, size, 1}])
    :ets.insert(:buffer, {{:data, new_index}, data})
    if new_index == 1 do
      :ets.delete(:buffer, {:data, new_index - size})
    end
  end

  def get do
    :ets.select(:buffer, [{{{:data, :_}, :"$1"}, [], [:"$1"]}])
  end
end
```

samuelmanzanera:
I still have some issue of loading the first time.It takes many times to display the page.

prix-uniris:
> I still have some issue of loading the first time.It takes many times to display the page.

Okay found the issue, its when we start the node, since it searched for 10txns it keeps running until we have it. Will fix this

prix-uniris:
> 

Fixed it, can you recheck

samuelmanzanera (approved):

PR #601, opened 2022-10-03
Fix paging address with selected fields

Neylix:
# Description

Fix a bug when requesting a chain to the DB with selected fields.
DB return transaction chain paginated by size of 10, if the chain has more than 10 transactions, DB return the paging address which is the transaction address of the last transaction read.
The problem is if we request specific fields without address, the transaction read have a nil address and so the paging address returned is nil. This beahavior leads to an infinite loop in some case.

## Type of change

# How Has This Been Tested?

Create a transaction chain with more than 10 transactions, request the chain with some specific fields without address (`TransactionChain.get(address, [data: [:content], validation_stamp: [:timestamp]])`) the return is `{chain, true, paging_address}`
with pagin address != nil

# Checklist:

### DIFF
--- a/lib/archethic/db/embedded_impl/chain_reader.ex
+++ b/lib/archethic/db/embedded_impl/chain_reader.ex
@@ -120,6 +120,9 @@ defmodule Archethic.DB.EmbeddedImpl.ChainReader do
   end
 
   defp do_process_get_chain(fd, address, fields) do
+    # Always return transaction address
+    fields = if Enum.empty?(fields), do: fields, else: Enum.uniq([:address | fields])
+
     column_names = fields_to_column_names(fields)
 
     # Read the transactions until the nb of transactions to fullfil the page (ie. 10 transactions)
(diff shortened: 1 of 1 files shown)


lib/archethic/db/embedded_impl/chain_reader.ex
```diff
@@ -120,6 +120,15 @@ defmodule Archethic.DB.EmbeddedImpl.ChainReader do
   end
 
   defp do_process_get_chain(fd, address, fields) do
+    # Always return transaction address
+    fields =
+      with false <- Enum.empty?(fields),
```
samuelmanzanera, line 125: If we want to avoid to check the list few times just to include a field you could leverage either:
- Enum.uniq -> `Enum.uniq([ :address | fields ])`
- MapSet -> `MapSet.new(fields) |> MapSet.union(MapSet.new([:address])`

samuelmanzanera (approved):

Issue #584, opened 2022-09-21
Certificate in origin_key api is not checked


Neylix:
### Describe the problem you discovered

Actually, when origin public key is not tpm family, we do not control the certificate.
But someone can put some data in the certificate attribute in the api and so create a transaction with some data inside.
As origin transaction are free and generated by nodes, people can create not tpm public key and store data in the transaction for free

### Describe the solution you'd like

In the API or in pending validation, if the origin public key is not tpm family, certificate should be empty

samuelmanzanera:
You should maybe leverage: `Crypto.verify_key_certificate?` to attest the public key with the certificate

PR #629, opened 2022-10-13
Certificate in origin_key api is not checked #584

apoorv-2204:
# Description

The endpoint `/api/origin_key` is used to add valid origin device for secure tx creation, But the certificate is not checked
for Origin :software || :on_chain_wallet , allowing end users to to send random data to certificate params, which directly gets
included to tx. Since origin txs are free, we must restrict the flaw to ensure only valid certificate contents are accepted in tx.

Though an empty certificate is allowed for Origin :software || :on_chain_wallet from end users, if we receive a valid certificate but with illicit intentions , we must verify the certificate with root CA public key, if corresponding valid root CA public key is not found the Certificate is declined or no new tx creation.

## Fixes 
- [x] Certificate in origin_key api is not checked #584)

## Type of change
- Bug fix 
  - Avoid invalid certs, avoid random data in certs
- Breaking change 
  - Crypto.ex
      - Certificate validation only occurs when a valid certificate contents not "", Valid root ca key not "", Valid origin public_key.
  
- [x] This change requires a documentation update

# How Has This Been Tested?

- [x] Unit test cases
- [x] Doc tests

# Checklist:


lib/archethic_web/controllers/api/origin_key_controller.ex
```diff
@@ -80,12 +94,18 @@ defmodule ArchethicWeb.API.OriginKeyController do
...
+      false ->
         {400, %{status: "error - invalid public key"}}
 
       {:exists?, true} ->
         {400, %{status: "error - public key exists"}}
 
+      :error ->
+        {400, %{status: "error - certificate error"}}
+
+      {:error, :invalid_certificate} ->
```
samuelmanzanera, line 106: How do you distinguish between invalid hex from public key and certificate ?
The easiest would be to leverage Ecto.Changeset with a schema to allow public key in hex while ensuring it's a valid public key and certificate which should be hex based as well.

lib/archethic/crypto.ex
```diff
@@ -1175,13 +1175,31 @@ defmodule Archethic.Crypto do
     end
   end
 
+  def get_root_ca_public_key(<<0::8, _::8, _::binary>>) do
```
samuelmanzanera, line 1178: I think you can use general case with 
```suggestion
  def get_root_ca_public_key(_), do: ""
```

apoorv-2204:
Manual Testing: 
Should Accept when, certificate is empty and valid origin public key

Screenshot_20221027_153557


should return already exists
Screenshot_20221027_153719

Should give invalid size
Screenshot_20221027_153805

Must be hexadecimal
Screenshot_20221027_153825

certificate must be hex
Screenshot_20221027_153930

must give invalid cert
Screenshot_20221027_154003

blank not allowed
Screenshot_20221027_154052


exceeds limit
Screenshot_20221027_154115

invalid origin
Screenshot_20221027_154201

lib/archethic/replication.ex
```diff
@@ -111,6 +106,12 @@ defmodule Archethic.Replication do
           TransactionChain.write_transaction(tx)
 
           :ok = ingest_transaction(tx)
+
```
apoorv-2204, line 109: ?

samuelmanzanera: The logging wasn't in the right place

samuelmanzanera:
tested with this conf: 
```
config :archethic, Archethic.Crypto,
  root_ca_public_keys: [
    usb: "3059301306072a8648ce3d020106082a8648ce3d03010703420004f0fe701a03ce375a6e57adbe0255808812036571c1424db2779c77e8b4a9ba80a15b118e8e7465ee2e94094e59c4b3f7177e99063af1b19bfcc4d7e1ac3f89dd"  |> Base.decode16!(case: :mixed)
  ]
```

PR #632, opened 2022-10-15
Add selfsigned certificate by default in prod config

Neylix:
# Description

By default there is no certificate for https in prod config. But without certificate, the https endpoint is not launched and so we cannot resolve request for aeweb redirection with dnslink.

To enable this functionnality, a self signed certificate is configured by default so the https endpoint is started and the node can serve aeweb website with certificate and https encryption

## Type of change

# Checklist:

### DIFF
--- a/lib/archethic_web/domain.ex
+++ b/lib/archethic_web/domain.ex
@@ -83,8 +83,20 @@ defmodule ArchethicWeb.Domain do
         keyfile = Keyword.fetch!(https_conf, :keyfile)
         certfile = Keyword.fetch!(https_conf, :certfile)
 
-        key = File.read!(keyfile) |> read_pem() |> hd()
-        cert = File.read!(certfile) |> read_pem() |> hd() |> elem(1)
+        key_content =
+          case File.read(keyfile) do
+            {:ok, content} -> content
+            {:error, _} -> File.read!(Application.app_dir(:archethic, keyfile))
+          end
+
+        cert_content =
+          case File.read(certfile) do
+            {:ok, content} -> content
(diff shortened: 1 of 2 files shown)


lib/archethic_web/domain.ex
```diff
@@ -80,8 +80,8 @@ defmodule ArchethicWeb.Domain do
           |> Application.get_env(ArchethicWeb.Endpoint)
           |> Keyword.fetch!(:https)
 
-        keyfile = Keyword.fetch!(https_conf, :keyfile)
-        certfile = Keyword.fetch!(https_conf, :certfile)
+        keyfile = Application.app_dir(:archethic, Keyword.fetch!(https_conf, :keyfile))
```
samuelmanzanera, line 83: I don't think you should do this change 
If someone wants to use a true certificate, for example cerbot/let's encrypt manages it in /etc folder. Hence we should let the file location as mentioned in the conf.

If you need to define the app_dir for self signed, we could do it in the conf itself.

samuelmanzanera (changes requested):

lib/archethic_web/domain.ex
Neylix: Yes right, the problem is that we cannot use `Application.app_dir(:archethic)` in config file as the application archethic do not exists yet.
And we need to use it because when we do a release, the application change it's folder depending of version number.
I don't find a way to create the complete path of the file during application compilation so I just added a case to read the file and if it's not found, add the app_dir and try again

samuelmanzanera (approved):

PR #657, opened 2022-10-26
Stop self repair if transaction's processing failed

samuelmanzanera:
# Description

This change ensures the self-repair will be halted if the transaction processing is failing

## Type of change

# Checklist:

### DIFF
--- a/lib/archethic/self_repair/sync.ex
+++ b/lib/archethic/self_repair/sync.ex
@@ -276,7 +276,7 @@ defmodule Archethic.SelfRepair.Sync do
     Logger.info("Need to synchronize #{Enum.count(transaction_summaries)} transactions")
     Logger.debug("Transaction to sync #{inspect(transaction_summaries)}")
 
-    Task.Supervisor.async_stream_nolink(
+    Task.Supervisor.async_stream(
       TaskSupervisor,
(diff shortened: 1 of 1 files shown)


Neylix (approved):

lib/archethic/self_repair/sync.ex
```diff
@@ -276,7 +276,7 @@ defmodule Archethic.SelfRepair.Sync do
     Logger.info("Need to synchronize #{Enum.count(transaction_summaries)} transactions")
     Logger.debug("Transaction to sync #{inspect(transaction_summaries)}")
 
-    Task.Supervisor.async_stream_nolink(
+    Task.Supervisor.async_stream(
```
apoorv-2204, line 279: It will crash the parent process, if downloading fails. process_transaction will only return error. Crash parent process to stop synchronisation of  transaction summary, and any  further, if downloading of anyone summary fails.

samuelmanzanera: The synchronisation will be stopped and the last sync date would not be updated as here: https://github.com/archethic-foundation/archethic-node/blob/master/lib/archethic/self_repair/scheduler.ex#L119

PR #662, opened 2022-11-02
Fix bootstraping bugs

Neylix:
# Description

Fixes multiple bugs: 
- Node transaction was not added in account mem table loader, causing error in replication when retrieving unspent outputs.
- Time condition to start self repair was not well handled, causing self repair to not start when it should.
- If a node has changed it's IP after some times, new nodes should always connect to the latest IP available even if they didn't synchronize all the node transactions
- Improve of synced? flag update
- Fix Ecto warning
- Fix error `DateTime.to_unix(nil, :second)` while loading transaction with code

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/lib/archethic/replication.ex
+++ b/lib/archethic/replication.ex
@@ -50,14 +50,18 @@ defmodule Archethic.Replication do
   It will download the transaction chain and unspents to validate the new transaction and store the new transaction chain
   and update the internal ledger and views
   """
-  @spec validate_and_store_transaction_chain(validated_tx :: Transaction.t()) ::
+  @spec validate_and_store_transaction_chain(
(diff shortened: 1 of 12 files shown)


lib/archethic/replication.ex
```diff
@@ -240,12 +246,12 @@ defmodule Archethic.Replication do
       transaction_type: tx.type
     )
 
-    inputs = fetch_inputs(tx)
+    inputs = if self_repair?, do: nil, else: fetch_inputs(tx)
```
samuelmanzanera, line 249: I think you should return empty list instead of nil

samuelmanzanera (approved):

Issue #645, opened 2022-10-19
Fix self-repair notifier


samuelmanzanera:
### Describe the problem you discovered

The current implementation of the self-repair notifier is not correct.
Some changes are required.

### Describe the solution you'd like

To correct if we should:
- [x] 1)  Exclude network transactions (as they are synchronized on all the nodes – no need of notifier for those)
- [x] 2) Read the transaction before to send it (for now, we are not sending the entire transaction, only the summary)
- [x] 3) Check if the transaction exists before to send it (to avoid sending significant amount of data)
- [x] 4) Add a flag in the transaction validator to not validate transaction movements in the `ReplicateTransaction` message
- [x] 5) Using a stream by chain to have the transactions in order

internet-zero:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Product-and-Design-Workspace-61efe71b9083a90018250d89/issues/archethic-foundation/archethic-node/645?planning-poker) with Zenhub @samuelmanzanera @Neylix @apoorv-2204

samuelmanzanera:
Closed by #671

Issue #566, opened 2022-09-12
Notify beacon summary in case of network topology change


samuelmanzanera:
### Is your feature request related to a problem?

From now, we are using the self-repair notifier to sync transactions when a shard topology change: node inside a shard is offline and we need to repair the shard to ensure availability.

However, in the case of beacon summary this is not currently implemented.

### Describe the solution you'd like

The notifier should also manage the beacon summary to support high availability

samuelmanzanera:
Closed by #671

PR #671, opened 2022-11-07
:jigsaw:  566 & 645, Aggregate Fix & Fix self repair notifier #645

apoorv-2204:
# Description

####  Self-Repair:
-Determine which shards plummeted. 
-Determine which transactions those shards were in charge of.
-Expand new shards for those transactions.

Proposed Solution :
- Use minimal information to inform the new shard.
   - `%ShardRepair{}` message with  two fields, the `genesis address` and the `last address`.
- The New Shard collects data on its own.
     - Via `fetch Transaction` and `Replicate validate transaction chain`
- Handle concurrent task requests using an FSM for each genesis address,to have bottleneck .                                                 
   - As a result multiple same job are not deployed concurrently.
- 
- [x] #645 
- [x] #566 

## Type of change

# How Has This Been Tested?

### Testing

- Unit tests

To test in real condition we need to update some conditions to limit the number of nodes we need to run.
Here is the patch to apply which change 
- number of storage node always return 3
- minimum geo patch to 1 and minimum avg_availability to 0.0
- remove burning fee address from IO storage node
- return static value for coingecko api (just to avoid spamming the API and having error)
- use a static geopatch when the node is a local one
[patch.txt](https://github.com/archethic-foundation/archethic-node/files/10112935/patch.txt)

Then you can run 5 nodes
- wait them to be authorized
- send a transaction
- look for which node stored it (normally 3 nodes)
- stop 1 or 2 of these nodes
After the next self repair when they goes unavailable, the notifier of each node should be triggered. The last storage node of the transaction should send to 1 or 2 new node the ShardRepair message, and these nodes should replicate it.
Other node should also replicate a SummaryAggregate if the disconnected node was a storage node for it

# Checklist:


apoorv-2204:
paste

lib/archethic/self_repair/notifier.ex
```diff
@@ -1,173 +1,72 @@
...
-    end
+  @doc """
+  Add a new address in the address list of the RepairWorker
+  """
+  @spec add_message(pid(), ShardRepair.t()) :: :ok
+  def add_message(pid, %ShardRepair{
+        storage_address: storage_address,
+        io_addresses: io_addresses
+      }) do
+    GenServer.cast(pid, {:add_address, storage_address, io_addresses})
```
samuelmanzanera, line 70: We should use abstraction with the module name instead of giving implementation details.

lib/archethic/self_repair.ex
```diff
@@ -83,4 +91,63 @@ defmodule Archethic.SelfRepair do
...
+    case diff_node do
+      [] ->
+        :ok
+
+      nodes ->
+        unavailable_nodes = Enum.map(nodes, & &1.first_public_key)
+
+        DynamicSupervisor.start_child(
+          NotifierSupervisor,
+          {Impl,
```
samuelmanzanera, line 110: I think you mean `Notifier` here.
Also it would be good to give some doc for the function.

test/archethic/self_repair/impl_test.exs
```diff
@@ -1,4 +1,4 @@
-defmodule Archethic.SelfRepair.Notifier.ImplTest do
+defmodule Archethic.SelfRepair.NotifierTest do
```
samuelmanzanera, line 1: You should also rename the file

Neylix:
Should we add the version number `@vsn Mix.Project.config()[:version]` in the GenServer ?

PR #673, opened 2022-11-07
Tx summary digest

samuelmanzanera:
# Description

Self-repair inconsistent existing transaction.
Those can occurred in case of network inconsistency or split brain.

## Type of change

- Enhancements

# Checklist:


samuelmanzanera:
> The update you made is to check if the transaction we already have in DB is the same as the one in the summary. If not we download it. But even if we download it, the replication will fail and return `transaction already exist` so we download the transaction for nothing. We should replace the existing transaction by the one downloaded in the self repair

Yes true indeed.
I have to find a way to do it properly then.
To avoid any issue in the chain stored.

PR #674, opened 2022-11-08
Ensure destination of elected message

### LINKED ISSUE #646: Ensure right destination of P2P messages

We discovered some ways of DDOS attacks as we are not checking whether an incoming message is rightly targeted to a node.
Hence, some adversary could spam the network to change some state in the network, or disturb it.

I propose few changes to resolve that:
- [x] Add check whether the node receiving the message is one of the elected for:
     - [x] ReplicateTransaction (io storage nodes)
     - [x] ReplicateTransactionChain (chain storage nodes)
     - [x] StartMining (validation nodes)
     - [x] NewBeaconSlot (beacon storage nodes)
 - [x] Include the sender public key in the `P2P.Message.process` function to leverage the source of sending for further verifications
 - [x] Provide some verification in the `ReplicationAttestation` processing to ensure the transaction summaries are correct. (like in the `BeaconChain.SlotValidation.valid_transaction_attestations?) -> to avoid memory message and memory overflow in the beacon chain subsets processes

samuelmanzanera:
# Description

Fixes #646 

## Type of change

- Enhancement

# Checklist:

### DIFF
--- a/lib/archethic/p2p/message.ex
+++ b/lib/archethic/p2p/message.ex
@@ -16,6 +16,8 @@ defmodule Archethic.P2P.Message do
 
   alias Archethic.Crypto
 
+  alias Archethic.Election
+
   alias Archethic.Mining
 
   alias Archethic.P2P
@@ -85,6 +87,8 @@ defmodule Archethic.P2P.Message do
   alias Archethic.TransactionChain.Transaction
   alias Archethic.TransactionChain.Transaction.CrossValidationStamp
   alias Archethic.TransactionChain.Transaction.ValidationStamp
+
+  alias Archethic.TransactionChain.Transaction.ValidationStamp.LedgerOperations
 
   alias Archethic.TransactionChain.Transaction.ValidationStamp.LedgerOperations.VersionedUnspentOutput
 
@@ -1280,7 +1284,8 @@ defmodule Archethic.P2P.Message do
   @doc """
   Handle a P2P message by processing it and return list of responses to be streamed back to the client
   """
-  def process(%GetBootstrappingNodes{patch: patch}) do
+  @spec process(request(), Crypto.key()) :: response()
+  def process(%GetBootstrappingNodes{patch: patch}, _) do
     top_nodes = P2P.authorized_and_available_nodes()
 
     closest_nodes =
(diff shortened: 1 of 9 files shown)


Neylix (commented):
What is the goal to add `sender_public_key` in message process, as it is never used ?

lib/archethic/p2p/message.ex
```diff
@@ -1388,26 +1396,37 @@ defmodule Archethic.P2P.Message do
...
+          welcome_node_public_key: welcome_node_public_key,
+          validation_node_public_keys: validation_nodes
+        },
+        _
+      ) do
     with {:election, true} <- {:election, Mining.valid_election?(tx, validation_nodes)},
+         {:elected, true} <-
+           {:elected, Enum.any?(validation_nodes, &(&1 == Crypto.last_node_public_key()))},
          {:mining, false} <- {:mining, Mining.processing?(tx.address)} do
       {:ok, _} = Mining.start(tx, welcome_node_public_key, validation_nodes)
```
Neylix, line 1411: You removed the `%OK{}` message response after starting Mining

lib/archethic/p2p/message.ex
```diff
@@ -1438,89 +1460,110 @@ defmodule Archethic.P2P.Message do
...
+      )
 
-            %AcknowledgeStorage{
-              address: tx.address,
-              signature:
-                Crypto.sign_with_first_node_key(TransactionSummary.serialize(tx_summary)),
-              node_public_key: Crypto.first_node_public_key()
-            }
+    # Replicate transaction chain only if the current node is one of the chain storage nodes
+    if Utils.key_in_node_list?(storage_nodes, Crypto.first_node_public_key()) do
```
Neylix, line 1477: You should not control the storage node pool if the transaction is a network type

samuelmanzanera:
> What is the goal to add `sender_public_key` in message process, as it is never used ?

The goal is to prepare in the case whether the message have been sent by the right peer.

lib/archethic/p2p/message.ex
samuelmanzanera: Why verify also the io storage nodes here ? For the IO it's checked on the %ReplicateTransaction{} message

Neylix: Because when we replicate a network transaction, the validations nodes send the message `ReplicateTransactionChain` to all storage and IO nodes

samuelmanzanera: For network transaction, ReplicateTransactionChain are sent to all the authorized nodes, and ReplicateTransaction is send to the all the rest of the nodes.

Neylix:
```
context
    |> ValidationContext.get_io_replication_nodes()
    |> P2P.broadcast_message(
      if Transaction.network_type?(validated_tx.type),
        do: %ReplicateTransactionChain{
          transaction: validated_tx
        },
        else: %ReplicateTransaction{transaction: validated_tx}
    )
```

samuelmanzanera: Forgotten this one.
I guess the chain storage nodes are only the authorized to provide more security during the tx validation, then the chain is replicated on all the nodes.

Neylix (approved):

PR #677, opened 2022-11-09
Fix validation node election

Neylix:
# Description

This PR fixes 2 bugs :
- The election for validation nodes is based on authorized nodes and after a filter is applied to target only available nodes. A problem can occurs if the selected validation nodes are all unavailable nodes, resulting in an empty list of validation nodes after the filter. To solve this, all election are done with the list of authorized and available nodes.
- During the self repair, the P2P view is updated after synchronization of the transactions. The problem is that the new available / unavailable nodes are not synchronized at the same time according to the time a node synchronize the transaction. So for some time the nodes may have a different P2P view resulting in error of election. To solve this, we can apply an availability date some time after the self repair as for the node authorization date

## Type of change

# How Has This Been Tested?

Integration test updated.

Many test by connecting / deconnecting nodes, doing self repair after many summaries

# Checklist:


lib/archethic/replication/transaction_validator.ex
```diff
@@ -193,12 +194,33 @@ defmodule Archethic.Replication.TransactionValidator do
...
              },
            cross_validation_stamps: cross_validation_stamps
          }
        ) do
-    authorized_nodes = P2P.authorized_and_available_nodes(tx_timestamp)
+    # TODO remove before mainnet launch
+    authorized_nodes =
+      if protocol_version <= 3 do
+        authorized_nodes =
+          Enum.filter(MemTable.authorized_nodes(), fn
```
samuelmanzanera, line 207: Why not using: `P2P.authorized_nodes()` ?

lib/archethic/p2p/mem_table.ex
```diff
@@ -694,15 +702,30 @@ defmodule Archethic.P2P.MemTable do
...
     Logger.info("Node globally available", node: Base.encode16(first_public_key))
 
-    if !:ets.member(@availability_lookup_table, first_public_key) do
-      true = :ets.insert(@availability_lookup_table, {first_public_key})
+    availability_pos = Keyword.fetch!(@discovery_index_position, :available?)
+
+    unless :ets.lookup_element(@discovery_table, first_public_key, availability_pos) do
+      availability_update_pos = Keyword.fetch!(@discovery_index_position, :availability_update)
+
+      :ets.update_element(@discovery_table, first_public_key, {availability_pos, true})
```
samuelmanzanera, line 721: You can combine the operations into a single update:
```elixir
:ets.update_element(table, key, [{pos1, val1}, {pos2, val2}]
```

config/prod.exs
```diff
@@ -204,7 +204,9 @@ config :archethic, Archethic.SharedSecrets.NodeRenewalScheduler,
 config :archethic, Archethic.SelfRepair.Scheduler,
   # Every day at 00:05:00
   # To give time for the beacon chain to produce summary
-  interval: System.get_env("ARCHETHIC_SELF_REPAIR_SCHEDULER_INTRERVAL", "0 5 0 * * * *")
+  interval: System.get_env("ARCHETHIC_SELF_REPAIR_SCHEDULER_INTRERVAL", "0 5 0 * * * *"),
+  # Availability application date 10 minutes after self repair
+  availability_application: 900
```
samuelmanzanera, line 209: I don't get those numbers
- 5s -> 10s
- 10m -> 900s

If you want to add 10m to the summary time, it should be 600s, no ?

lib/archethic/db.ex
```diff
@@ -62,12 +62,7 @@ defmodule Archethic.DB do
 
   @callback transaction_exists?(binary()) :: boolean()
 
-  @callback register_p2p_summary(
-              node_public_key :: Crypto.key(),
-              date :: DateTime.t(),
-              available? :: boolean(),
-              average_availability :: float()
-            ) :: :ok
+  @callback register_p2p_summary(list()) :: :ok
```
samuelmanzanera, line 65: You should include which kind of information is present in the list.
This way the implementations will know which data must be handled.

samuelmanzanera (approved):

PR #682, opened 2022-11-14
Check if transaction already exists during transaction validation

### LINKED ISSUE #678: Check transaction's address existence during validation

For now, the transaction's address existence is checked on the replication nodes
It might occur – due to network failure, sync failure or split brain – two transaction with the same address are replicated by the network, causing chain inconsistency and therefore invalid proof of integrity for the chain by the next transaction.

So to increase consistency in the transaction validation across multiple cross validation nodes, the existence should also be checked in the validation stamp.
We can add in the pending transaction validation a check to ensure the transaction is not already replicated and store this error result in the validation stamp.

### Additional context

_No response_

samuelmanzanera:
# Description

Trying to send the same transaction from the node should be prevented during transaction validation by checking the existence of the transaction

Fixes #678 

## Type of change

- Enhancement

# How Has This Been Tested?

If you send an existing transaction from the node an error should be triggered.

# Checklist:

### DIFF
--- a/lib/archethic.ex
+++ b/lib/archethic.ex
@@ -11,6 +11,7 @@ defmodule Archethic do
   alias __MODULE__.Mining
 
   alias __MODULE__.P2P
+  alias __MODULE__.P2P.Node
 
   alias __MODULE__.DB
 
@@ -19,7 +20,6 @@ defmodule Archethic do
   alias __MODULE__.P2P.Message.NewTransaction
   alias __MODULE__.P2P.Message.Ok
   alias __MODULE__.P2P.Message.StartMining
-  alias __MODULE__.P2P.Node
 
   alias __MODULE__.TransactionChain
   alias __MODULE__.TransactionChain.Transaction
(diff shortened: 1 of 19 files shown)


lib/archethic/mining/pending_transaction_validation.ex
```diff
@@ -79,6 +83,32 @@ defmodule Archethic.Mining.PendingTransactionValidation do
     end
   end
 
+  defp valid_not_exists(%Transaction{address: address}) do
+    storage_nodes = Election.chain_storage_nodes(address, P2P.authorized_and_available_nodes())
```
Neylix, line 87: As for other quorum_read, you should take only nearest nodes and locally available one.

But maybe this should be embedded directly in the quorum read function, as we always do it before. So we will avoid doing misstake on this selection and also don't write the same code many times

Neylix (changes requested):

lib/archethic/mining/pending_transaction_validation.ex
samuelmanzanera: :+1:

lib/archethic.ex
```diff
@@ -55,8 +48,6 @@ defmodule Archethic do
       do_send_transaction(tx)
     else
       P2P.authorized_and_available_nodes()
-      |> Enum.filter(&Node.locally_available?/1)
-      |> P2P.nearest_nodes()
```
Neylix, line 59: This selection should stay here I think as when we want to forward transaction, we should do it on local and near node to be fast

Issue #627, opened 2022-10-12
Determines whether transaction input is spent from the last transaction in the chain


samuelmanzanera:
### Is your feature request related to a problem?

Our UTXO model ensure the inputs are spent for each transaction (re-created).
After #621, if a token is not used, it's marked as spent in the previous transaction and unspent in the next one.
This is a strange behavior for the UTXO model.
We should be able to determine whether a transaction input is spent based on the last transaction on the chain.

### Describe the solution you'd like

After #624, the spent UTXO would not accessible in the MemTables.
Hence, to determine if the input is really spent (logically speaking within the entire chain), we should provide a way to determine this.

In the P2P request to fetch transaction inputs of a previous transaction, from the spent assets (stored on disk – after #624), we can request the last transaction of the chain to determine if the input is an UTXO.
Then we could flag the input as not unspent.

The explorer and the API will then be impacted in the way they retrieve information.

internet-zero:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Product-and-Design-Workspace-61efe71b9083a90018250d89/issues/archethic-foundation/archethic-node/627?planning-poker) with Zenhub @samuelmanzanera @apoorv-2204 @Neylix

PR #684, opened 2022-11-14
Move inputs from ETS to disk

### LINKED ISSUE #624: Purge Account UTXO table once an input is spent

All the inputs are currently store in memory.
In order to scale, you should evict the spent transaction outputs from the ETS tables and store them on disk for a later query.

After each transaction replication, you are notifying the previous shards for a new transaction in the chain.
Hence, the UTXO are flagged as spent.
We could then at this time delegates to a permanent storage on disk to reduce the footprint of the ETS tables.

### Additional context

_No response_

bchamagne:
# Description

Once a UTXO is spent, we move it to disk to reduce memory footprint.

Fixes #624

## Type of change

- New feature 

# How Has This Been Tested?

This is tested in the integration tests. 
I did not do any BEFORE/AFTER metrics, should I do some benchmark?

# Checklist:


bchamagne:
**IMPORTANT NOTE:** we loose millisecond precision due to serialization using `DateTime.to_unix/1`

lib/archethic/account/mem_tables/token_ledger.ex
```diff
@@ -142,22 +173,33 @@ defmodule Archethic.Account.MemTables.TokenLedger do
...
+                protocol_version: protocol_version
+              }
+              | acc
+            ]
+        end
+      end)
+      |> Enum.reverse()
+
+    spent = Inputs.get_inputs(:token, address)
+    Enum.concat(spent, unspent)
```
samuelmanzanera, line 203: You could pass the spent inputs as the reduce's accumulator. So, you could prepend to the list the unspent output in the reduce's function which will be more efficient than concatenating the two lists.

test/archethic/transaction_chain/transaction_input_test.exs
```diff
@@ -0,0 +1,28 @@
+defmodule Archethic.TransactionChain.TransactionInputTest do
```
samuelmanzanera, line 1: I think there is already a TransactionInputTest, located at `test/archethic/transaction_chain/transaction/input_test.exs`

lib/archethic/db/embedded_impl/inputs.ex
```diff
@@ -0,0 +1,73 @@
...
+  There will be many files
+  """
+
+  alias Archethic.DB.EmbeddedImpl
+  alias Archethic.TransactionChain.VersionedTransactionInput
+  alias Archethic.Utils
+
+  @type ledger :: :token | :UCO
+
+  @spec append_inputs(
```
samuelmanzanera, line 13: A good thing would be to leverage a behavior in the `db.ex` file, so external contexts particularly during testing could leverage implementations as mock to simplify testing without relying on the file system. Hence, the `embedded_impl.ex` could be the facing API to the inputs offloading.

samuelmanzanera (changes requested):
Thanks for the PR, I wrote you some comments to improve it in terms of performances and maintainability.

samuelmanzanera:
> **IMPORTANT NOTE:** we loose millisecond precision due to serialization using `DateTime.to_unix/1`

If we would like to preserve it, we can use `DateTime.to_unix(date, :millisecond)`

samuelmanzanera:
After restart of the node, the inputs are duplicated with different state

image

To reproduce it, start a node, do 2 faucet requests, stop the node, and restart it. Then go to the previous addresses generated, you will see the double inputs, one spent and one unspent, while they are the same.

samuelmanzanera (approved):

lib/archethic/account/mem_tables/token_ledger.ex
```diff
@@ -130,34 +135,59 @@ defmodule Archethic.Account.MemTables.TokenLedger do
...
+
+    DB.stop_inputs_writer(pid)
   end
 
   @doc """
   Retrieve the entire inputs for a given address (spent or unspent)
   """
   @spec get_inputs(binary()) :: list(VersionedTransactionInput.t())
   def get_inputs(address) when is_binary(address) do
+    spent = DB.get_inputs(:token, address)
```
Neylix, line 170: I think it's not needed to always read the DB and the ETS table as the inputs will be either in the ETS table, either in the DB but never both in the same time for an address.
So you may first read the ETS table to retrieve the inputs, and if there is nothing in then you can go throught DB.

lib/archethic/account/mem_tables/uco_ledger.ex
```diff
@@ -113,46 +118,72 @@ defmodule Archethic.Account.MemTables.UCOLedger do
...
+
+    DB.stop_inputs_writer(pid)
   end
 
   @doc """
   Retrieve the entire inputs for a given address (spent or unspent)
   """
   @spec get_inputs(binary()) :: list(VersionedTransactionInput.t())
   def get_inputs(address) when is_binary(address) do
+    spent = DB.get_inputs(:UCO, address)
```
Neylix, line 165: Same as token

lib/archethic/db/embedded_impl/inputs_reader.ex
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  defp deserialize_inputs_file(<<>>, acc), do: acc
+
+  defp deserialize_inputs_file(bitstring, acc) do
+    {input_bit_size, rest} = Utils.VarInt.get_value(bitstring)
+
+    # every serialization contains some padding to be a binary (multipe of 8bits)
+    {input_bitstring, rest} =
+      case rem(input_bit_size, 8) do
```
Neylix, line 39: Here you need to do some calculation because the `input_bit_size` is the size of the input itself without the wrapped bits.
Maybe in the inputs writer you should wrap the input before getting it's size, as it you don't need to use `rem` to determine the wrapped bits size. The function `VersionedTransactionInput.deserialize(input_bitstring)` may returns a non empty rest which will correspond to the wrapped bit that can be ignored

lib/archethic/db/embedded_impl/inputs_reader.ex
```diff
@@ -0,0 +1,54 @@
...
+  alias Archethic.DB.EmbeddedImpl.InputsWriter
+  alias Archethic.TransactionChain.VersionedTransactionInput
+  alias Archethic.Utils
+
+  @spec get_inputs(ledger :: InputsWriter.ledger(), address :: binary()) ::
+          list(VersionedTransactionInput.t())
+  def get_inputs(ledger, address) do
+    filename = InputsWriter.address_to_filename(ledger, address)
+
+    case File.open(filename, [:read, :binary]) do
```
Neylix, line 15: Here you can use directly use `File.read` instead of `File.open` and `IO.binread`

lib/archethic/account/mem_tables/token_ledger.ex
Neylix: I think on the explorer it's just a view, but when there is a new transaction, in `spend_all_unspent_outputs` you delete all unspent outputs from the ETS table and write them in DB.
In the explorer, we check if an output is spent or not based on the last unspent output of the chain. But in the ETS table there is only the last transaction unspent ouputs I think.

bchamagne: You are correct! and that's why I had to do the #627. I finally understand fully.

lib/archethic/db/embedded_impl/inputs_reader.ex
```diff
@@ -0,0 +1,34 @@
...
+      {:error, :enoent} ->
+        []
+
+      {:ok, bin} ->
+        bin
+        |> deserialize_inputs_file([])
+    end
+  end
+
+  defp deserialize_inputs_file(<<>>, acc), do: Enum.reverse(acc)
```
Neylix, line 25: I think the `Enum.reverse` is not necessary as the write is not written in a specific order.

Neylix (approved):

samuelmanzanera (approved):

Issue #641, opened 2022-10-18
Take into consideration sobelow warnings


Neylix:
### Is your feature request related to a problem?

Sobelow as been integrated into CI with PR #631, hence we should take into consideration the warnings returned by it.

### Describe the solution you'd like

Sobelow provide 3 confidence levels (low, medium, high).
We should fix the warning at least for medium and high confidence.
We should add an option to exit the process with an error if medium or high confidence warnings are returned to fail the CI test using option `--exit medium`.
Low confidence level may be corrected but should not be blocking.

internet-zero:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Product-and-Design-Workspace-61efe71b9083a90018250d89/issues/archethic-foundation/archethic-node/641?planning-poker) with Zenhub @apoorv-2204 @Neylix @samuelmanzanera

PR #702, opened 2022-11-17
Take into consideration sobelow warnings#641

apoorv-2204:
# Description

We must acknowledge static security analysis from sobelow lib.Its Important.

Fixes 

- [ ] #641 

## Type of change

# How Has This Been Tested?

# Checklist:


lib/archethic/utils.ex
```diff
@@ -811,4 +812,17 @@ defmodule Archethic.Utils do
...
+  @doc """
+  Uses String.to_existing_atom() if fails , then it creates a new atom.
+  Prevents String.to_atom DDOS attack.sobelow.
+  """
+  @spec convert_string_to_atom(string :: String.t()) :: atom()
+  def convert_string_to_atom(string) do
+    String.to_existing_atom(string)
+  rescue
+    _ ->
+      String.to_atom(string)
```
samuelmanzanera, line 826: The goal of the String.to_existing_atom is to avoid atom table exhaustion. 
By catching the throw you might exhaust the atom limits.

bchamagne (commented):
Hey Apoorv, 
just a few comments to get rid of that `Utils.convert_string_to_atom/1`

Cheers

lib/archethic/utils/websocket/handler.ex
```diff
@@ -195,7 +197,7 @@ defmodule Archethic.Utils.WebSocket.SocketHandler do
     {command, queries} = Map.pop(queries, msg_ref)
     state = Map.put(state, :queries, queries)
 
-    status = payload["status"] |> String.to_atom()
+    status = Utils.convert_string_to_atom(payload["status"])
```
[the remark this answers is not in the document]
tenmoves: https://www.erlang.org/doc/efficiency_guide/commoncaveats.html#list_to_atom-1

lib/archethic/utils.ex
apoorv-2204: > The goal of the String.to_existing_atom is to avoid atom table exhaustion. By catching the throw you might exhaust the atom limits.

hi this function , may be ,it was created to mitigate, that if atom does not already exist then  
`String.to_existing_atom ` will throw a error or raise, causing the Process to crash. To avoid crash
if a atom does not exists then create a atom,  `String.to_atom`. So the Process don't crash.
The first time execution of SC will require atom creation. 

`https://github.com/nccgroup/sobelow/blob/master/lib/sobelow/dos/string_to_atom.ex`

```  
@moduledoc """
  # Denial of Service via `String.to_atom`
  In Elixir, atoms are not garbage collected. As such, if user input
  is passed to the `String.to_atom` function, it may result in memory
  exhaustion. Prefer the `String.to_existing_atom` function for untrusted
  user input.
```

apoorv-2204: difficult to conclude , what is expected?

bchamagne: I believe this happens because the module where the atom is used is not loaded yet (modules are lazy loaded). The issue might be fixed by ensuring the module is loaded. See that example:

```
iex(27)> String.to_existing_atom("add_uco_transfer")
** (ArgumentError) errors were found at the given arguments:

  * 1st argument: not an already existing atom

    :erlang.binary_to_existing_atom("add_uco_transfer", :utf8)
    iex:27: (file)
```

```
iex(41)> Code.ensure_loaded(Archethic.Contracts.Interpreter.TransactionStatements)
{:module, Archethic.Contracts.Interpreter.TransactionStatements}
```

```
iex(45)> String.to_existing_atom("add_uco_transfer")
:add_uco_transfer
```

bchamagne (approved):

tenmoves (approved):

PR #705, opened 2022-11-21
AEWeb - Directory listing

### LINKED ISSUE #596: Listing directory for web hosting deployment without index.html

When you are uploading a folder for webhosting , if the folder doesn't contain an index.html, the page is not displayed as not found.
In any other webserver, it display by default listing directory.

We could create a basic HTML page to list all the files present in the reference transaction and create links to other transactions.

### Additional context

_No response_

bchamagne:
# Description

Display a directory listing when the user hits an endpoint that correspond to a directory and the directory does not contain a `index.html` file.
This should be performant because only 1 transaction is downloaded (the reference transaction). But, because of this we cannot determine the files' sizes. 

I used Firefox source as template.

Fixes #596

## Type of change

- New feature 

# How Has This Been Tested?

Tested with unit tests.
Tested with the `aewebcli` with the following structure [myaewebdir.zip](https://github.com/archethic-foundation/archethic-node/files/10056869/myaewebdir.zip)

Screenshot 2022-11-30 at 14-31-20 Index of _

# Checklist:

### DIFF
--- a/config/prod.exs
+++ b/config/prod.exs
@@ -237,7 +237,6 @@ config :archethic, ArchethicWeb.Endpoint,
           false ->
             System.get_env("ARCHETHIC_DOMAIN_NAME", "mainnet.archethic.net")
         end,
-      port: System.get_env("ARCHETHIC_HTTPS_PORT", "50000") |> String.to_integer(),
       path: "/explorer"
     }),
   http: [:inet6, port: System.get_env("ARCHETHIC_HTTP_PORT", "40000") |> String.to_integer()],
--- a/config/prod.exs
+++ b/config/prod.exs
@@ -227,21 +227,19 @@ config :archethic, ArchethicWeb.FaucetController,
 # before starting your production server.
 config :archethic, ArchethicWeb.Endpoint,
   explorer_url:
-    (case(System.get_env("ARCHETHIC_NETWORK_TYPE") == "testnet") do
-       true ->
-         Path.join([
-           "https://",
-           "#{System.get_env("ARCHETHIC_DOMAIN_NAME", "testnet.archethic.net")}:#{System.get_env("ARCHETHIC_HTTPS_PORT", "50000")}",
-           "explorer"
-         ])
+    URI.to_string(%URI{
+      scheme: "https",
+      host:
+        case(System.get_env("ARCHETHIC_NETWORK_TYPE") == "testnet") do
+          true ->
+            System.get_env("ARCHETHIC_DOMAIN_NAME", "testnet.archethic.net")
 
-       false ->
-         Path.join([
-           "https://",
-           "#{System.get_env("ARCHETHIC_DOMAIN_NAME", "mainnet.archethic.net")}:#{System.get_env("ARCHETHIC_HTTPS_PORT", "50000")}",
-           "explorer"
-         ])
-     end),
+          false ->
+            System.get_env("ARCHETHIC_DOMAIN_NAME", "mainnet.archethic.net")
+        end,
+      port: System.get_env("ARCHETHIC_HTTPS_PORT", "50000") |> String.to_integer(),
+      path: "/explorer"
+    }),
   http: [:inet6, port: System.get_env("ARCHETHIC_HTTP_PORT", "40000") |> String.to_integer()],
   url: [host: nil, port: System.get_env("ARCHETHIC_HTTP_PORT", "40000") |> String.to_integer()],
   cache_static_manifest: "priv/static/cache_manifest.json",
--- a/config/prod.exs
+++ b/config/prod.exs
@@ -227,11 +227,21 @@ config :archethic, ArchethicWeb.FaucetController,
 # before starting your production server.
 config :archethic, ArchethicWeb.Endpoint,
   explorer_url:
-    Path.join([
-      "https://",
-      "#{System.get_env("ARCHETHIC_DOMAIN_NAME", "mainnet.archethic.net")}:#{System.get_env("ARCHETHIC_HTTPS_PORT", "50000")}",
-      "explorer"
-    ]),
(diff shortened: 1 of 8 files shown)


bchamagne:
Please note there is a new config variable that should have different value in `MainNet` et `TestNet`

config/prod.exs
```diff
@@ -226,6 +226,12 @@ config :archethic, ArchethicWeb.FaucetController,
 # which you should run after static files are built and
 # before starting your production server.
 config :archethic, ArchethicWeb.Endpoint,
+  explorer_url:
```
samuelmanzanera, line 229: You should also take care of the testnet possibliity.
For this you can leverage `ARCHETHIC_NETWORK_TYPE` env variable.

So in short:
- if network type is testnet
   - take the `ARCHETHIC_DOMAIN_NAME`
   - if not defined, use `https://testnet.archethic.net`
- else
   - take the `ARCHETHIC_DOMAIN_NAME`
   - if not defined, use `https://mainnet.archethic.net`

samuelmanzanera (approved):

lib/archethic_web/controllers/api/web_hosting_controller.ex
```diff
@@ -57,41 +64,153 @@ defmodule ArchethicWeb.API.WebHostingController do
...
+          transaction :: Transaction.t(),
+          cached_headers :: list()
+        ) ::
+          {:ok, listing_html :: binary() | nil, encodage :: nil | binary(), mime_type :: binary(),
+           cached? :: boolean(), etag :: binary()}
+  def dir_listing(request_path, params, transaction, cache_headers) do
+    url_path = Map.get(params, "url_path", [])
+
+    with {:ok, json_content} <- Jason.decode(transaction.data.content),
+         {cached?, etag} <- get_cache(cache_headers, transaction.address, url_path) do
```
Neylix, line 127: Here is the file is already cached, you don't need to parse the content, you can directly return the result saying that the content is alreday cached as it is done for `get_file_content`

lib/archethic_web/controllers/api/web_hosting_controller.ex
```diff
@@ -57,41 +64,153 @@ defmodule ArchethicWeb.API.WebHostingController do
...
+          _ ->
+            {
+              Path.join(["/" | url_path]),
+              %{href: request_path |> Path.join("..") |> Path.expand()}
+            }
+        end
+
+      assigns =
+        Map.keys(json_content_subset)
+        |> Enum.map(fn key ->
```
Neylix, line 152: Here you can directly use `Enum.map` on the map and directly access to the value like
```
json_content_subset
|> Enum.map(fn
  {key, %{"address" => address}} -> {:file, key, address}
  {key, _} -> {:dir, key}
end)
```

This will simplify as we get the key and the value in the same instruction

lib/archethic_web/controllers/api/web_hosting_controller.ex
```diff
@@ -57,41 +64,153 @@ defmodule ArchethicWeb.API.WebHostingController do
            cached? :: boolean(), etag :: binary()}
           | {:error, :invalid_address}
           | {:error, :invalid_content}
-          | {:error, :not_found}
+          | {:error, :file_not_found}
+          | {:error, :is_a_directory}
           | {:error, :invalid_encodage}
           | {:error, any()}
   def get_website(params = %{"address" => address}, cache_headers) do
     url_path = Map.get(params, "url_path", [])
```
Neylix, line 72: I think here you can use destructuration for the needed transaction field
```
{:ok,
 transaction = %Transaction{
   address: last_address,
   data: %TransactionData{content: content}
 }} <- Archethic.get_last_transaction(address)
```

lib/archethic_web/controllers/api/web_hosting_controller.ex
```diff
@@ -57,41 +64,153 @@ defmodule ArchethicWeb.API.WebHostingController do
...
+
+  @spec dir_listing(
+          request_path :: String.t(),
+          params :: map(),
+          transaction :: Transaction.t(),
+          cached_headers :: list()
+        ) ::
+          {:ok, listing_html :: binary() | nil, encodage :: nil | binary(), mime_type :: binary(),
+           cached? :: boolean(), etag :: binary()}
+  def dir_listing(request_path, params, transaction, cache_headers) do
```
Neylix, line 123: Same here you can use destrucuration like 
```
def dir_listing(
  request_path,
  params,
  %Transaction{
    address: tx_address,
    data: %TransactionData{content: content},
    validation_stamp: %ValidationStamp{timestamp: tx_timestamp}
  },
  cache_headers
)
```

this will remove the need to create the `get_transaction_timestamp` function. By the way you can add a validation stamp in your unit test so the timestamp will not be `nil`

config/prod.exs
```diff
@@ -226,6 +226,20 @@ config :archethic, ArchethicWeb.FaucetController,
...
+      scheme: "https",
+      host:
+        case(System.get_env("ARCHETHIC_NETWORK_TYPE") == "testnet") do
+          true ->
+            System.get_env("ARCHETHIC_DOMAIN_NAME", "testnet.archethic.net")
+
+          false ->
+            System.get_env("ARCHETHIC_DOMAIN_NAME", "mainnet.archethic.net")
+        end,
+      port: System.get_env("ARCHETHIC_HTTPS_PORT", "50000") |> String.to_integer(),
```
Neylix, line 240: I think the port on mainnet.archethic.net and testnet.archethic.net would be the default ssl port 443. Port 50000 is used when querying directly the node with it's ip

Neylix (changes requested):

bchamagne:
I suggest to review commits 1 by 1 since @Neylix comments. (last one looks huge but it's just moving code around)

Neylix (approved):

PR #709, opened 2022-11-23
compute node reward based on scheduler time

### LINKED ISSUE #693: Calcul node reward based on scheduler time

Actually the `node_reward` transaction is sent based on a configurable interval. n production this interval is set to 1 transaction per day.
In the reward calculation we get the number of days of the current month and divide 50 by this number. This fit well for production configuration but if the `node_reward` transaction are sent in a different interval than once a day, the reward calculation is not accurate.

We should divide the monthly reward amount (currently 50$) by the number of interval in the current month

tenmoves:
# Description

Actually the node_reward transaction is sent based on a configurable interval. n production this interval is set to 1 transaction per day.
In the reward calculation we get the number of days of the current month and divide 50 by this number. This fit well for production configuration but if the node_reward transaction are sent in a different interval than once a day, the reward calculation is not accurate.

We should divide the monthly reward amount (currently 50$) by the number of interval in the current month

Note: 

I saw that sometimes it can take a little bit of time to compute, so I decided to do he computation in compile time to not slow down the execution.

Fixes #693 

## Type of change

# How Has This Been Tested?

The function that will provide the number of occurrences per month was tested with doctests, the reward function in itself is already tested I don't think we should test the calculus in itself but more the behaviour (which is already the case).

# Checklist:

### DIFF
--- a/lib/archethic/reward.ex
+++ b/lib/archethic/reward.ex
@@ -34,6 +34,7 @@ defmodule Archethic.Reward do
   require Logger
 
   @unit_uco 100_000_000
+  @number_of_reward_occurences_per_month Utils.number_of_reward_occurences_per_month()
 
   @doc """
   Get rewards amount for validation nodes
@@ -47,8 +48,12 @@ defmodule Archethic.Reward do
       |> OracleChain.get_uco_price()
       |> Keyword.get(:usd)
 
-    trunc(50 / uco_usd_price / Calendar.ISO.days_in_month(date.year, date.month) * @unit_uco)
+    number_of_reward_occurences_per_month = number_of_reward_occurences_per_month()
+
+    trunc(50 / uco_usd_price / number_of_reward_occurences_per_month * @unit_uco)
   end
+
+  defp number_of_reward_occurences_per_month(), do: @number_of_reward_occurences_per_month
 
   @doc """
   Create a transaction for minting new rewards
--- a/lib/archethic/utils.ex
+++ b/lib/archethic/utils.ex
@@ -848,13 +848,11 @@ defmodule Archethic.Utils do
       current_datetime
       |> Date.beginning_of_month()
       |> date_to_datetime_converter.(true)
-      |> IO.inspect(label: "start")
 
     end_of_month_datetime =
       current_datetime
       |> Date.end_of_month()
       |> date_to_datetime_converter.(false)
-      |> IO.inspect(label: "end")
 
     interval
     |> CronParser.parse!(true)
--- a/lib/archethic/utils.ex
+++ b/lib/archethic/utils.ex
@@ -12,6 +12,8 @@ defmodule Archethic.Utils do
 
   alias Archethic.P2P.Node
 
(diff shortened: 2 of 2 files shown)


lib/archethic/reward.ex
```diff
@@ -47,9 +48,13 @@ defmodule Archethic.Reward do
       |> OracleChain.get_uco_price()
       |> Keyword.get(:usd)
 
-    trunc(50 / uco_usd_price / Calendar.ISO.days_in_month(date.year, date.month) * @unit_uco)
+    number_of_reward_occurences_per_month = number_of_reward_occurences_per_month()
+
+    trunc(50 / uco_usd_price / number_of_reward_occurences_per_month * @unit_uco)
   end
 
+  defp number_of_reward_occurences_per_month(), do: @number_of_reward_occurences_per_month
```
samuelmanzanera, line 56: What is the purpose to use a private function while you could use directly the module attribute?

lib/archethic/utils.ex
```diff
@@ -811,4 +813,57 @@ defmodule Archethic.Utils do
...
+      ) do
+    time = fn
+      true -> " 00:00:00Z"
+      false -> " 23:59:59Z"
+    end
+
+    date_to_datetime_converter = fn date, start_of_month? ->
+      time = time.(start_of_month?)
+
+      date
```
samuelmanzanera, line 844: I think you can directly use string interpolation: 
```elixir
NaiveDateTime.from_iso8601!("#{date}#{time}")
```

lib/archethic/reward.ex
tenmoves: because reading module attribute directly will run the function at runtime, however if we wrap in a function it will be interpreted at compile time and only the value of the call will be written in the corresponding Beam file.
Since this function may take some time (more than 1s) and the cron job will not change in runtime I thought it's better this way.

the first paragraph of this article explains this a little better. https://medium.com/multiverse-tech/how-to-speed-up-your-elixir-compile-times-part-1-understanding-elixir-compilation-64d44a32ec6e

lib/archethic/utils.ex
tenmoves: thank you I will try it

lib/archethic/utils.ex
```diff
@@ -811,4 +813,57 @@ defmodule Archethic.Utils do
...
+      |> to_string()
+      |> Kernel.<>(time)
+      |> NaiveDateTime.from_iso8601!()
+    end
+
+    {start_of_month_datetime, end_of_month_datetime} =
+      current_datetime
+      |> then(
```
samuelmanzanera, line 852: This is a bit hard to read.
I don't think `then` brings value here.
Hard to tell what is going here without analyzing the anonymous functions.

tenmoves: I will rewrite it

samuelmanzanera (approved):

Neylix (approved):

PR #717, opened 2022-11-25
Update last transaction address only if timestamp is greater

### LINKED ISSUE #712: Update last transaction address only if the timestamp is greater

Actually there is 2 way to set the last address for a chain
- from the replication
- from the notification of a last address

In both way, if the current last address in the DB has a greater timestamp than the one received, the last address is replaced by the new one. Which is not correct

In the `ChainIndex` module, before setting the new last address, we should control that the new timestamp is greater than the one already stored

tenmoves:
# Description

Actually there is 2 way to set the last address for a chain

from the replication
from the notification of a last address
In both way, if the current last address in the DB has a greater timestamp than the one received, the last address is replaced by the new one. Which is not correct.

 before setting the new last address, we should control that the new timestamp is greater than the one already stored

Fixes #712 

## Type of change

# How Has This Been Tested?

I tested 2 cases: 

- we should not update last transaction only if timestamp is lesser
- we should update last transaction if timestamp is greater

# Checklist:

### DIFF
--- a/lib/archethic/db/embedded_impl/chain_index.ex
+++ b/lib/archethic/db/embedded_impl/chain_index.ex
@@ -378,30 +378,12 @@ defmodule Archethic.DB.EmbeddedImpl.ChainIndex do
         _ -> true
       end
 
-    write_last_chain_transaction(
-      write_last_chain_transaction?,
-      filename,
-      encoded_data,
-      genesis_address,
-      new_address,
-      unix_time
-    )
-  end
-
-  defp write_last_chain_transaction(false, _, _, _, _, _), do: :ok
-
-  defp write_last_chain_transaction(
-         true,
-         filename,
-         encoded_data,
-         genesis_address,
-         new_address,
-         unix_time
-       ) do
-    with :ok <- File.write!(filename, encoded_data, [:binary, :append]),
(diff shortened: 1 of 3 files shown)


lib/archethic/db/embedded_impl/chain_index.ex
```diff
@@ -371,14 +371,36 @@ defmodule Archethic.DB.EmbeddedImpl.ChainIndex do
...
+        [{_, ^new_address, _}] -> false
+        [{_, _, chain_unix_time}] when unix_time < chain_unix_time -> false
+        _ -> true
+      end
 
-      _ ->
-        :ok = File.write!(filename, encoded_data, [:binary, :append])
-        true = :ets.insert(:archethic_db_last_index, {genesis_address, new_address, unix_time})
-        :ok
+    write_last_chain_transaction(
```
samuelmanzanera, line 381: I think using the if statement would simplify the api of the `write_last_chain_transaction/6` fn

tenmoves: no problem I will rewrite it

samuelmanzanera (approved):

Neylix (approved):

PR #718, opened 2022-11-25
Skip replication if there is a (cross)validation error

### LINKED ISSUE #696: Do not send replication message if validation has error

Actually whenever the validation of a transaction has error or not, we send the replication message to the storage nodes.
But if there is an error in the validation, the storage nodes will not replicate the transaction. So we send a message for nothing because we already know that the replication will fail

In the distributed workflow, if there is an error in the validation stamp or in the cross validation stamp, then do not send the replication message to the storage node, and directly notify the error to the welcome node.

bchamagne:
# Description

In the distributed workflow, if there is an error in the validation stamp or in the cross validation stamp, then do not send the replication messages to the storage nodes, and directly notify the error to the welcome node.

`notify_error` used to be an internal event but we had to change this to a function to be able to use it from the `replication.state_enter` callback.

Fixes #696 

**IMPORTANT NOTE:** The scenario `:insufficient_funds` is not checked in the distributed_workflow. This means a transaction from an address with insufficient funds will not mark the stamps as invalid and the transaction will be replicated to the storage nodes where they'll be discarded.

- [x] check for insufficient funds

## Type of change

- Bug fix 

# How Has This Been Tested?

Tests added to the suite.
Scenario tested on the transaction builder (with 3 nodes): Creation of a token with an invalid JSON. You should see the usual error and you should see a "skipped replication" in the logs.

# Checklist:

### DIFF
--- a/lib/archethic/mining/validation_context.ex
+++ b/lib/archethic/mining/validation_context.ex
@@ -1261,4 +1261,25 @@ defmodule Archethic.Mining.ValidationContext do
     |> Enum.map(&Enum.at(storage_nodes, &1))
     |> Enum.reject(&Utils.key_in_node_list?(chain_storage_nodes, &1.first_public_key))
   end
+
+  @spec get_error_as_atom(t()) :: atom()
+  def get_error_as_atom(ctx = %__MODULE__{}) do
+    case ctx.validation_stamp.error do
+      nil ->
+        inconsistencies =
+          ctx.cross_validation_stamps
+          |> Enum.flat_map(& &1.inconsistencies)
+
+        case inconsistencies do
+          [] ->
+            nil
+
+          _ ->
+            :invalid_cross_validation_stamp
+        end
+
+      _ ->
+        :invalid_validation_stamp
+    end
+  end
 end
--- a/lib/archethic/mining/validation_context.ex
+++ b/lib/archethic/mining/validation_context.ex
@@ -1262,8 +1262,11 @@ defmodule Archethic.Mining.ValidationContext do
     |> Enum.reject(&Utils.key_in_node_list?(chain_storage_nodes, &1.first_public_key))
   end
 
-  @spec get_error_as_atom(t()) :: atom()
-  def get_error_as_atom(ctx = %__MODULE__{}) do
+  @doc """
+  Get the first available error or nil
+  """
+  @spec get_first_error(t()) :: atom()
+  def get_first_error(ctx = %__MODULE__{}) do
     case ctx.validation_stamp.error do
       nil ->
         inconsistencies =
--- a/lib/archethic/mining/validation_context.ex
+++ b/lib/archethic/mining/validation_context.ex
@@ -1274,12 +1274,12 @@ defmodule Archethic.Mining.ValidationContext do
           [] ->
             nil
 
-          _ ->
-            :invalid_cross_validation_stamp
+          [first_error | _] ->
+            first_error
         end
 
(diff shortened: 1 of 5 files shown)


samuelmanzanera:
I think the purpose is also to avoid replication if all the nodes are agreed on the transaction's invalidity.

lib/archethic/mining/validation_context.ex
```diff
@@ -1261,4 +1261,25 @@ defmodule Archethic.Mining.ValidationContext do
     |> Enum.map(&Enum.at(storage_nodes, &1))
     |> Enum.reject(&Utils.key_in_node_list?(chain_storage_nodes, &1.first_public_key))
   end
+
+  @spec get_error_as_atom(t()) :: atom()
+  def get_error_as_atom(ctx = %__MODULE__{}) do
```
samuelmanzanera, line 1266: I didn't get the name `as_atom `, as the error is already in the atom format.

samuelmanzanera: This suggest the error will be converted into an atom

test/archethic/mining/distributed_workflow_test.exs
```diff
@@ -968,7 +998,279 @@ defmodule Archethic.Mining.DistributedWorkflowTest do
...
+      Workflow.add_cross_validation_stamp(
+        coordinator_pid,
+        %CrossValidationStamp{
+          signature:
+            Crypto.sign(
+              [ValidationStamp.serialize(context.validation_stamp), <<1>>],
+              elem(@publickey1, 1)
+            ),
+          node_public_key: elem(@publickey1, 0),
+          inconsistencies: [:signature]
```
samuelmanzanera, line 1057: I think in this text it should not have the cross validation stamp errors but only the pending one.

bchamagne: good catch

lib/archethic/mining/validation_context.ex
```diff
@@ -1261,4 +1261,28 @@ defmodule Archethic.Mining.ValidationContext do
     |> Enum.map(&Enum.at(storage_nodes, &1))
     |> Enum.reject(&Utils.key_in_node_list?(chain_storage_nodes, &1.first_public_key))
   end
+
+  @doc """
+  Get the first available error or nil
+  """
+  @spec get_first_error(t()) :: atom()
+  def get_first_error(ctx = %__MODULE__{}) do
```
Neylix, line 1269: An other way to write this function using pattern matching is : 
```elixir
def get_first_error(%__MODULE__{
      validation_stamp: %ValidationStamp{error: nil},
      cross_validation_stamps: cross_validation_stamps
    }) do
  cross_validation_stamps
  |> Enum.reduce_while(nil, fn
    %CrossValidationStamp{inconsistencies: []}, nil -> {:cont, nil}
    %CrossValidationStamp{inconsistencies: [first_error | _rest]}, nil -> {:halt, first_error}
  end)
end

def get_first_error(%__MODULE__{validation_stamp: %ValidationStamp{error: error}}), do: error
```

It may be more consistent with the rest of the project as a lot of function use pattern matching in priority.
As you prefer

Neylix (commented):
Working well :+1: 
Just a remark to be consistent with the coding style of the project

Neylix (approved):

samuelmanzanera (approved):

Neylix (approved):

PR #721, opened 2022-11-28
Connection's reconnect is asynchronous

### LINKED ISSUE #710: Use asynchronous connection

In the `Connection` module, when a host is not reachable we try to reconnect every 500ms. But if the host ip do not exists, the `gen_tcp.connect` will fall in timeout after 4 seconds. This instruction is synchronous and block all new messages in the Genserver, this can perturb other functionnalities.

In the `Connection` module, the function to try to reconnect `transport.handle_connect` should be asynchronous

bchamagne:
# Description

Reconnect workflow used to be synchronous, which meant all incoming messages were queued to be processed later, and the callers had to wait until eventually they receive a :timeout. 

This PR changed the reconnect workflow to be asynchronous, so the callers can get a :closed error message without delay.

**NOTE**: the first connection is still asynchronous to be able to keep existing code.

Fixes #710

## Type of change

- Bug fix

# How Has This Been Tested?

Case added to the unit tests

# Checklist:

### DIFF
--- a/lib/archethic/p2p/client/connection.ex
+++ b/lib/archethic/p2p/client/connection.ex
@@ -86,10 +86,8 @@ defmodule Archethic.P2P.Client.Connection do
         {:ok, {:connected, socket}, new_data}
 
       {:error, _} ->
-        {_, new_data, actions} =
-          handle_event(:enter, {:connected, make_ref()}, :disconnected, data)
-
-        {:ok, :disconnected, new_data, actions}
+        actions = [{{:timeout, :reconnect}, 500, nil}]
+        {:ok, :disconnected, data, actions}
     end
   end
 
@@ -204,12 +202,10 @@ defmodule Archethic.P2P.Client.Connection do
           transport: transport
         }
       ) do
-    me = self()
-
     # try to connect asynchronously so it does not block the messages coming in
-    Task.start_link(fn ->
-      # handle_connect might take a long time
-      send(me, {:connect_result, transport.handle_connect(ip, port)})
+    # Task.async/1 will send a {:info, {ref, result}} message to the connection process
+    Task.async(fn ->
+      transport.handle_connect(ip, port)
     end)
 
     :keep_state_and_data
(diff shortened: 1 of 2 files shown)


lib/archethic/p2p/client/connection.ex
```diff
@@ -77,8 +77,20 @@ defmodule Archethic.P2P.Client.Connection do
...
+    # We call ourselves the `state_enter` because `GenStateMachine` will not do it from init
+    # `handle_connect` might take a long time
+    case transport.handle_connect(ip, port) do
+      {:ok, socket} ->
+        {_, new_data} = handle_event(:enter, :disconnected, {:connected, socket}, data)
+        {:ok, {:connected, socket}, new_data}
+
+      {:error, _} ->
+        {_, new_data, actions} =
+          handle_event(:enter, {:connected, make_ref()}, :disconnected, data)
```
Neylix, line 90: I think here we should not call this handle event as we where not previously in `:connected` state. We should just return
```elixir
actions = [{{:timeout, :reconnect}, 500, nil}]
{:ok, :disconnected, data, actions}
```

lib/archethic/p2p/client/connection.ex
```diff
@@ -186,20 +198,21 @@ defmodule Archethic.P2P.Client.Connection do
...
+    me = self()
 
-      {:error, _} ->
-        actions = [{{:timeout, :reconnect}, 500, nil}]
-        {:keep_state_and_data, actions}
-    end
+    # try to connect asynchronously so it does not block the messages coming in
+    Task.start_link(fn ->
+      # handle_connect might take a long time
+      send(me, {:connect_result, transport.handle_connect(ip, port)})
```
Neylix, line 212: Why sending a new message as the resolution of the task will already send a message with the result of function which can be catched with 
```elixir 
handle_event(:info, {ref, function_return}, :disconnected, data)
```

test/archethic/p2p/client/connection_test.exs
```diff
@@ -45,6 +45,36 @@ defmodule Archethic.P2P.Client.ConnectionTest do
...
     end
 
+    test "should get an error, :closed when trying to reach an unreachable node" do
+      defmodule MockTransportUnreachable do
+        alias Archethic.P2P.Client.Transport
+
+        @behaviour Transport
+
+        def handle_connect(_ip, _port) do
+          {:error, :timeout}
```
Neylix, line 55: Here you should add `Process.sleep` to simulate a long connection time.
Actually the `assert {:error, :closed}` may return the result after the connection is done event with `Process.sleep`
The test should test if we receive a return `{:error, :closed}` before the connect function is finished

Neylix (changes requested):

lib/archethic/p2p/client/connection.ex
```diff
@@ -77,8 +77,18 @@ defmodule Archethic.P2P.Client.Connection do
...
     }
 
-    actions = [{:next_event, :internal, :connect}]
-    {:ok, :disconnected, data, actions}
+    # First connection attempt is synchronous (to deal with the `start_link` followed by a `send_message`)
+    # We call ourselves the `state_enter` because `GenStateMachine` will not do it from init
+    # `handle_connect` might take a long time
+    case transport.handle_connect(ip, port) do
+      {:ok, socket} ->
+        {_, new_data} = handle_event(:enter, :disconnected, {:connected, socket}, data)
```
samuelmanzanera, line 85: Using `handle_event` is not the most elegant, to avoid duplicate code, you could wrap the code in the `enter, :disconnected, {:connected, socket}, data` event in a private function

bchamagne: As discussed with Sam, we may use the `:postpone` atom in gen_statem to deal with this in an elegant manner.

samuelmanzanera (approved):

Neylix (approved):

PR #722, opened 2022-11-28
Ensure there is never more validation nodes than storage nodes 

### LINKED ISSUE #695: Number of validation nodes should not be more than replica number

In the consensus we have an election for storage nodes and for validation nodes. In the consensus we should not have more validation nodes than storage nodes.
In some case it is actually possible according to the amount of a transaction

In the election of validation nodes, we should also verify that the number of validation nodes is <= of the storage nodes number.

In the same time, we can remove the sort by authorized nodes in the storage nodes election as this sort is overwritten by the rotating key sort just after

bchamagne:
# Description

On the validation nodes' election, we have to make sure we never have more validation nodes than storage nodes. 

Fixes #695 

## Type of change

- Bug fix 

# How Has This Been Tested?

Test added to the suite.

# Checklist:


test/archethic/election_test.exs
```diff
@@ -131,6 +134,94 @@ defmodule Archethic.ElectionTest do
       assert Enum.map(first_election, & &1.last_public_key) !=
                Enum.map(second_election, & &1.last_public_key)
     end
+
+    @tag :toto
```
Neylix, line 138: It's the story of toto ... :laughing:

Neylix (approved):

PR #742, opened 2022-12-02
[Aeweb] Return content of reference Txn 

apoorv-2204:
# Description

## Type of change

Please delete options that are not relevant.

- New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Screenshot_20221202_222550

# Checklist:

- My code dont follows the style guidelines of this project
- I have not performed a self-review of my own code
- I have not commented my code, particularly in hard-to-understand areas
- I have not made corresponding changes to the documentation
- My changes generate  new warnings
- I have added tests that prove my fix is effective or that my feature doesnt works
- New and existing unit tests  doesnt pass locally with my changes
- Any dependent changes have been merged and published in downstream modules

PR #744, opened 2022-12-06
AEWeb cache (files and transactions)

### LINKED ISSUE #644: AEWeb: improve files response by caching their content

Actually, each time a file is requested for a website, the API download the last reference chain, find the transaction address of the file content, download this transaction and return the file content.
If the browser request 50 files which are stored in the same transaction, the API will download 50 times the same transaction. This is not performant.

To fix this issue we can create a cache for AEWeb, which will store the most request transaction content.
The cache can have a configurable size, i.e. for a node with 16 Gb of ram, AEWeb cache could be 2 Gb.
This cache should store the transaction address, it's content and the last date this transaction has been requested.
If we reach the maximum size and want to add new content, the oldest requested file should be delete to be replaced by the new one.
The behavior of web_hosting will be the following: 
- A request is made for a genesis address and a file
- Request the last transaction address of the genesis address chain
- Check if this last transaction is in the cache
  - If not, download the transaction and store address, content and current time in the cache
  - If yes, get it's content from it and update last requested time with current time
- Find the address of the requested file
- Check if the transaction address is in the cache
  - If not, download the transaction and store address, content and current time in the cache
  - If yes, get it's content from it and update last requested time with current time
- Return the file content

### Additional context

_No response_

bchamagne:
# Description

Adds `LRU` and `LRUDisk` modules to cache data on memory and on disk respectively. The algorithm used for eviction is the [Least Recent Used](https://en.wikipedia.org/wiki/Cache_replacement_policies#LRU), meaning that the oldest files read are evicted when space is required. The max size is in bytes and in config so users can control how much cache they  want to support.
Theses caches are used in the AEWeb controller to cache the reference transaction (queried on every request) and to cache the files' content.

There is a few things to discuss:
- the hash of the cache file name (currently base64(cache_key))
- where to put the modules (currently in archethic_cache)
- ~do a cast instead of a call for the put fn to not wait for I/O (will not be able to return succes boolean)~

There will be a conflict when #742 is merged.

Fixes #644

## Type of change

# How Has This Been Tested?

Tests added to the suite.

# Checklist:

### DIFF
--- a/lib/archethic_cache/lru.ex
+++ b/lib/archethic_cache/lru.ex
@@ -1,0 +1,149 @@
+defmodule ArchethicCache.LRU do
+  @moduledoc """
+  A cache that stores the values in an ETS table.
+  There are hooks available to be able to add effects (ex: write to disk).
+
+  It keeps track of the order and bytes in the genserver state.
+  The `bytes_used` are tracked in here because if we just monitor ETS table size, we would not be able to have a disk cache.
+  The `keys` are used to determine the Least Recent Used (first is the most recent used, last is the least recent used).
+  """
+
+  @spec start_link(GenServer.name(), non_neg_integer(), keyword()) ::
+          {:ok, binary()} | {:error, term()}
+  def start_link(name, max_bytes, opts \\ []) do
+    GenServer.start_link(__MODULE__, [name, max_bytes, opts], name: name)
+  end
+
+  @spec put(GenServer.server(), term(), term()) :: boolean()
+  def put(pid, key, value) do
+    GenServer.call(pid, {:put, key, value})
+  end
+
+  @spec get(GenServer.server(), term()) :: nil | term()
+  def get(pid, key) do
+    GenServer.call(pid, {:get, key})
+  end
+
+  @spec purge(GenServer.server()) :: :ok
+  def purge(pid) do
+    GenServer.call(pid, :purge)
+  end
+
+  def init([name, max_bytes, opts]) do
+    table = :ets.new(:"aecache_#{name}", [:set, {:read_concurrency, true}])
+
+    {:ok,
+     %{
+       table: table,
+       bytes_max: max_bytes,
+       bytes_used: 0,
+       keys: [],
+       put_fn:
+         Keyword.get(opts, :put_fn, fn _key, value ->
+           {value, :erlang.external_size(value)}
+         end),
+       get_fn:
+         Keyword.get(opts, :get_fn, fn _key, value ->
+           value
+         end),
+       evict_fn:
+         Keyword.get(opts, :evict_fn, fn _key, value ->
+           :erlang.external_size(value)
+         end)
+     }}
+  end
+
+  def handle_call({:get, key}, _from, state) do
+    {reply, new_state} =
+      case :ets.lookup(state.table, key) do
+        [{^key, value}] ->
+          {
+            state.get_fn.(key, value),
+            %{state | keys: state.keys |> move_front(key)}
+          }
+
+        [] ->
+          {nil, state}
+      end
+
+    {:reply, reply, new_state}
+  end
+
+  def handle_call({:put, key, value}, _from, state) do
+    size = :erlang.external_size(value)
+
+    if size > state.bytes_max do
+      {:reply, false, state}
+    else
+      # maybe evict some keys to make space
+      state =
+        evict_until(state, fn state ->
+          state.bytes_used + size <= state.bytes_max
+        end)
+
+      case :ets.lookup(state.table, key) do
+        [] ->
+          {value_to_store, size_to_store} = state.put_fn.(key, value)
+
+          :ets.insert(state.table, {key, value_to_store})
+
+          new_state = %{
+            state
+            | keys: [key | state.keys],
+              bytes_used: state.bytes_used + size_to_store
+          }
+
+          {:reply, true, new_state}
+
+        [{^key, old_value}] ->
+          # this is a replacement, we need to evict to update the bytes_used
+          bytes_evicted = state.evict_fn.(key, old_value)
+          {value_to_store, size_to_store} = state.put_fn.(key, value)
+
+          :ets.insert(state.table, {key, value_to_store})
+
+          new_state = %{
+            state
+            | keys: state.keys |> move_front(key),
+              bytes_used: state.bytes_used + size_to_store - bytes_evicted
+          }
+
+          {:reply, true, new_state}
+      end
+    end
+  end
+
+  def handle_call(:purge, _from, state) do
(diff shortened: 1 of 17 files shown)


bchamagne:
@Neylix reviewed this with me today. He proposed a refactoring to make the LRUDisk module simpler. In a few words, storing the size as well as the value in the ETS map will make everything easier.

lib/archethic_cache/lru_disk.ex
```diff
@@ -0,0 +1,69 @@
...
+  @spec start_link(GenServer.name(), non_neg_integer(), binary()) ::
+          {:ok, binary()} | {:error, term()}
+  def start_link(name, max_bytes, cache_dir) do
+    cache_dir = Path.join(cache_dir, "#{name}")
+    :ok = reset_directory(cache_dir)
+
+    LRU.start_link(name, max_bytes,
+      put_fn: fn key, value ->
+        # retry to create the dir everytime in case someones delete it
+        File.mkdir_p!(cache_dir)
```
samuelmanzanera, line 19: I didn't get the `someones delete it`.
If the folder doesn't exist, the LRU would crash and restart from fresh state.

I'm afraid of the overhead trying to recreate a folder each time an entry is added

lib/archethic_cache/lru.ex
```diff
@@ -0,0 +1,155 @@
...
+  end
+
+  defp evict_until(
+         state = %{table: table, keys: keys, evict_fn: evict_fn, bytes_used: bytes_used},
+         predicate
+       ) do
+    if predicate.(state) do
+      state
+    else
+      case List.last(keys) do
```
samuelmanzanera, line 132: I think you can optimize this one because: `List.last` and `List.delete` are O(n) time complexity both. You could do it in a once.

```elixir
case Enum.reverse(keys) do
  [] ->
     state
  [oldest_key | rest] ->
     ....
     evict_until(
            %{
              state
              | bytes_used: bytes_used - size,
                keys: rest
            },
            predicate
          )
end
```

lib/archethic_cache/lru.ex
```diff
@@ -0,0 +1,155 @@
...
+        [] ->
+          {nil, state}
+      end
+
+    {:reply, reply, new_state}
+  end
+
+  def handle_call(:purge, _from, state = %{table: table, evict_fn: evict_fn}) do
+    # we call the evict_fn to be able to clean effects (ex: file written to disk)
+    :ets.tab2list(table)
```
samuelmanzanera, line 69: You might use `:ets.foldr` to avoid twice list traversals.
This will save you the overhead of constructing a list for all rows in the table first.

lib/archethic_cache/lru_disk.ex
```diff
@@ -0,0 +1,66 @@
...
+
+  defp reset_directory(dir) do
+    File.rm_rf!(dir)
+    File.mkdir_p!(dir)
+  end
+
+  defp key_to_path(cache_dir, key) do
+    Path.join(
+      cache_dir,
+      Base.encode64(:erlang.term_to_binary(key))
```
Neylix, line 63: The file name is depending on the size of the requested url, so the file name can be very long. Maybe we should hash the key before encoding it in base 64, so the file name will have the same size no matter their url.
for example:
```elixir
addr = :crypto.strong_rand_bytes(34) |> Base.encode16
key = {addr, ["path", "to", "content"]}
Base.encode64(:erlang.term_to_binary(key))
# "g2gCbQAAAEQ4RkI4QkQ5RTkyRjVGRTZEODU0OTY5NDU1RTdGOEVBMDNCODQ0MEI2ODJEQkNFNUI2RkU0RjQ4MEFDRUQwMzc5NUNBQWwAAAADbQAAAARwYXRobQAAAAJ0b20AAAAHY29udGVudGo="

# Using a hash before
:crypto.hash(:sha256, :erlang.term_to_binary(key)) |> Base.encode64
# "mKj3eScg2Xh/949NXi59NjdHeNrhI4Zm3GuUv5Ehpis="
```

Neylix: Also we should use `Base.url_encode64` as the classic `Base.encode64` may add "/" in the file name which is a forbidden character

Neylix (changes requested):

lib/archethic_web/controllers/api/web_hosting_controller.ex
```diff
@@ -383,5 +405,24 @@ defmodule ArchethicWeb.API.WebHostingController do
...
+  defp do_get_file_content(_, _), do: :file_error
+
+  @spec search_transaction(binary()) :: {:ok, Transaction.t()} | {:error, atom()}
+  defp search_transaction(address) do
+    # :cache_tx is started by the ArchethicWeb.Supervisor
+    case LRU.get(:cache_tx, address) do
+      nil ->
+        case Archethic.search_transaction(address) do
+          {:ok, tx} ->
+            LRU.put(:cache_tx, address, tx)
```
Neylix, line 417: Actually the web hosting controller require only 3 fields from a transaction: address, data.content, validation_stamp.timestamp
Maybe we can reduce the data stored in the LRU by creating a Transaction object with only theses fieds and remove all other unecessary fields

lib/archethic_cache/lru_disk.ex
samuelmanzanera: What is the initial reason of the base64 ? why not hex ?

lib/archethic_web/controllers/api/web_hosting_controller.ex
samuelmanzanera: Using a tuple should do it.

lib/archethic_cache/lru_disk.ex
bchamagne: > What is the initial reason of the base64 ? why not hex ?

no reason, 64 will return a shorter string than 16, should I switch to 16?

bchamagne:
**This PR is blocked by #747 and we will resume development only when it's merged.** 
Reason: the web_hosting_controller.ex has been refactored in mentionned ticket.

bchamagne:
This is now ready and tested!

Neylix (approved):

lib/archethic_web/reference_transaction.ex
```diff
@@ -0,0 +1,71 @@
+defmodule ArchethicWeb.ReferenceTransaction do
```
samuelmanzanera, line 1: To remove confusion about the naming of this module, we could specify that is about web hosting. Otherwise it means a global reference transaction - which is not the case. It's only for AEWeb. 

Maybe `ArchethicWeb.WebHostingController. ReferenceTransaction`

samuelmanzanera (approved):

PR #746, opened 2022-12-06
add new Api to get beacon chain aggregate

### LINKED ISSUE #639: Add new API to get beacon aggregate

Since the PR #559 beacon and beacon summaries are not anymore transactions. Before we could get these transaction using graphql API `networkTransaction`, but now we don't have any way to get the summary data.

We should create a new schema in graphql like `beaconSummary` with a date in parameter which returns the beacon aggregate for this date.
If the date is the last summary date, API should use beacon summary and not beacon aggregate as it is done in explorer.
The date in parameter may be in this format "AAAA-MM-JJ".

### Additional context

I think we should't create an API that returns mutiple summaries as they can contains a lot of data.
Maybe we can create a new type for the transaction summaries which is paginated or have a size limit like
```
query {
  beaconSummary(date: "2022-10-18") {
    transactionSummaries(limit: 100) {
      address
    }
  }
}
```

tenmoves:
# Description

Since the PR https://github.com/archethic-foundation/archethic-node/pull/559 beacon and beacon summaries are not anymore transactions. Before we could get these transaction using graphql API networkTransaction, but now we don't have any way to get the summary data.

We create a new schema in graphql like beaconSummary with a timestamp in parameter which returns the beacon aggregate for this date.
If the date is the last summary date, API should use beacon summary and not beacon aggregate as it is done in explorer.

example graphql query

```
{
  beaconChainSummary(timestamp: 1670369760) {
    version
    p2pAvailabilities
    summaryTime
    availabilityAddingTime
    transactionSummaries(limit: 1, pagingOffset: 2) {
      fee
      type
      timestamp
      address
      movementsAddresses
    }
  }
}
```

Fixes #639 

## Type of change

# How Has This Been Tested?

- Tested locally in graphiQl
- Unit tests 

# Checklist:


lib/archethic/beacon_chain.ex
```diff
@@ -414,4 +416,55 @@ defmodule Archethic.BeaconChain do
         e
     end
   end
+
+  # Slots which are already has been added
+  # Real time transaction can be get from pubsub
+  def list_transactions_from_current_slots(date = %DateTime{} \\ DateTime.utc_now()) do
```
samuelmanzanera, line 422: If you can add typespecs this would be great !

lib/archethic_web/graphql_schema/resolver.ex
```diff
@@ -212,6 +214,37 @@ defmodule ArchethicWeb.GraphQLSchema.Resolver do
...
   end
 
+  def beacon_chain_summary(datetime) do
+    current_datetime = DateTime.utc_now()
+
+    next_datetime_summary_time =
+      datetime
+      |> BeaconChain.next_summary_date()
+
+    previous_current_date_summary_time =
```
samuelmanzanera, line 224: I think there variable name is confusing between `previous` and `next_summary_date`

lib/archethic_web/live/chains/beacon_live.ex
```diff
@@ -246,54 +242,6 @@ defmodule ArchethicWeb.BeaconChainLive do
...
-
-        _ ->
-          []
-      end
-    end)
-    |> Stream.filter(&match?({:ok, _}, &1))
-    |> Stream.flat_map(&elem(&1, 1))
-    |> Enum.to_list()
-  end
+  defdelegate list_transactions_from_current_slots(date \\ DateTime.utc_now()),
```
samuelmanzanera, line 245: Do you think it's necessary to use delegate?
Wouldn't it be easier to just call BeaconChain.list_transaction_from_current_slots?

samuelmanzanera:
I think there is a problem with the summary time

image

The summary should not be `DateTime.utc_now` but the next summary time.
This is causing the no display of the last transactions recorded.

lib/archethic_web/graphql_schema/resolver.ex
```diff
@@ -212,6 +214,37 @@ defmodule ArchethicWeb.GraphQLSchema.Resolver do
...
+      current_datetime
+      |> BeaconChain.next_summary_date()
+
+    authorized_nodes = P2P.authorized_and_available_nodes()
+
+    cond do
```
Neylix, line 230: Ithink it would be better to use a `case` statement instead of a `cond`, so the `DateTime.compare` will be done only once and not for each cond if the previous is false

lib/archethic_web/graphql_schema/resolver.ex
```diff
@@ -212,6 +214,37 @@ defmodule ArchethicWeb.GraphQLSchema.Resolver do
...
+      DateTime.compare(next_datetime_summary_time, previous_current_date_summary_time) == :gt ->
+        {
+          :error,
+          "No data found at this date"
+        }
+
+      DateTime.compare(next_datetime_summary_time, previous_current_date_summary_time) == :eq ->
+        {
+          :ok,
+          BeaconChain.fetch_and_aggregate_summaries(datetime, authorized_nodes)
```
Neylix, line 240: Also there you should pass the `next_datetime_summary_time` value instead of `datetime` because if datetime is not an exact summary time it will return an error as the summary is not found

lib/archethic_web/graphql_schema/resolver.ex
```diff
@@ -212,6 +214,37 @@ defmodule ArchethicWeb.GraphQLSchema.Resolver do
...
+        }
+
+      DateTime.compare(next_datetime_summary_time, previous_current_date_summary_time) == :lt ->
+        BeaconChain.fetch_summaries_aggregate(datetime, authorized_nodes)
```
Neylix, line 244: Same here for the `datetime`

Neylix (changes requested):

lib/archethic_web/graphql_schema/beacon_chain_summary_type.ex
```diff
@@ -0,0 +1,87 @@
...
+  """
+  object :transaction_summary do
+    field(:timestamp, :timestamp)
+    field(:address, :address)
+    field(:movements_addresses, list_of(:address))
+    field(:type, :string)
+    field(:fee, :integer)
+  end
+
+  scalar :p2p_availabilities do
```
Neylix, line 53: For the p2p availabilities, I would suggest to change the format of the returned data.
Actually the p2p availabilities are understanble by the nodes, but for a developer or a user, having the subset and a list of 0/1 without knowing what is the node behind is not very meaningfull.
Maybe format the availabilitie to something like this ?
```jsonc
{
  "p2pAvailabilities": [
    {
      "public_key": "0123456798ABCDEF",
      "available": true, // false
      "avairage_availability": 0.89,
      "endOfNodeSynchronization": false // true
    },
    {
      "public_key": "0123456798ABCDEF",
      "available": true, // false
      "avairage_availability": 0.89,
      "endOfNodeSynchronization": false // true
    }
  ]
}
```

lib/archethic/beacon_chain.ex
tenmoves: yes

lib/archethic_web/graphql_schema/resolver.ex
tenmoves: typo from the previous version

lib/archethic_web/live/chains/beacon_live.ex
tenmoves: no it's not necessary, will do a simple call if you find it more readable

lib/archethic_web/graphql_schema/resolver.ex
tenmoves: yep I had different conditions in version 1, will fix it.

lib/archethic_web/graphql_schema/resolver.ex
tenmoves: in this case they are equal but I will change it for clarity.

lib/archethic_web/graphql_schema/resolver.ex
tenmoves: yes

bchamagne:
I think there's something wrong with the dates (I might be wrong)

Screenshot 2023-01-05 at 12-08-41 GraphiQL Workspace
No timestamp given: I can see my transfer

Screenshot 2023-01-05 at 12-09-29 GraphiQL Workspace
I added the timestamp that is the date of the transfer: there is nothing

Screenshot 2023-01-05 at 12-10-45 GraphiQL Workspace
Same timestamp as above but a minute later (1 slot later): the date changed to slot - 1

tenmoves:
@bchamagne the behaviour you have detected is because you only have a single node running.

Neylix (approved):

PR #747, opened 2022-12-07
Incorporates new Aeweb developments

apoorv-2204:
# Description
https://github.com/archethic-foundation/aeweb-cli/pull/88

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- Test A
- Test B

# Checklist:


bchamagne:
- [x] Tests 
- [x] Directory listing 
- [x] Web hosting 

Good work mate!

bchamagne (approved):

config/dev.exs
```diff
@@ -1,6 +1,6 @@
 import Config
 
-config :logger, level: System.get_env("ARCHETHIC_LOGGER_LEVEL", "debug") |> String.to_atom()
+config :logger, level: System.get_env("ARCHETHIC_LOGGER_LEVEL", "warning") |> String.to_atom()
```
samuelmanzanera, line 3: I don't think this change is necessary, and it will disturb habits for the local development

apoorv-2204: reverted 👍 +1

lib/archethic_web/controllers/api/web_hosting_controller/resources.ex
```diff
@@ -0,0 +1,175 @@
...
+          true
+
+        _ ->
+          false
+      end
+
+    {cached?, etag}
+  end
+
+  defp is_a_directory(_metadata, ""), do: true
```
Neylix, line 162: I think there is no case where the file_path is empty as it is the result of `Enum.join(url_path, @path_seperator)` with a non empty url_path list

lib/archethic_web/controllers/api/web_hosting_controller/resources.ex
```diff
@@ -0,0 +1,175 @@
...
+
+  defp is_a_directory(metadata, file_path) do
```
Neylix, line 164: As this function returns a boolean, you should name it `is_a_directory?`

lib/archethic_web/controllers/api/web_hosting_controller/resources.ex
```diff
@@ -0,0 +1,175 @@
...
+
+  def get_file_content(_, true, _),
+    do: {:ok, nil, nil}
+
+  def get_file_content(
+        file_metadata = %{@addresses_key => address_list},
+        _cached? = false,
+        resource_path
+      ) do
+    resource_path = if resource_path == "", do: "index.html", else: resource_path
```
Neylix, line 96: I think the resource_path would never be an empty string as it is return by the function `get_file` which return either `Enum.join(url_path, @path_seperator)` or `"index.html"`

lib/archethic_web/controllers/api/web_hosting_controller/resources.ex
```diff
@@ -0,0 +1,175 @@
...
+
+    try do
+      file_content =
+        Enum.reduce(address_list, "", fn address, acc_map ->
```
Neylix, line 100: You may let the accumulator name as `acc` because `acc_map` sounds like the accumlator is a map but in this case it is a string

lib/archethic_web/controllers/api/web_hosting_controller/resources.ex
```diff
@@ -0,0 +1,175 @@
+defmodule ArchethicWeb.API.WebHostingController.Resources do
+  @moduledoc false
+
+  use ArchethicWeb, :controller
```
Neylix, line 4: Same as directory listing, you can remove
```elixir
  use ArchethicWeb, :controller

  alias Archethic

File: lib/archethic_web/controllers/api/web_hosting_controller/resources.ex [hunk 7]
```diff
@@ -0,0 +1,175 @@
...
+
+  alias Archethic
+
+  alias Archethic.TransactionChain.Transaction
+  alias Archethic.TransactionChain.TransactionData
+
+  use Pathex
+
+  require Logger
+  @addresses_key "addresses"
```
Neylix, line 14: Same as directory listing, global variables are not usefull I think since they are used only once

lib/archethic_web/controllers/api/web_hosting_controller/resources.ex
```diff
@@ -0,0 +1,175 @@
...
+
+          {:ok, decoded_content} = Jason.decode(txn_content)
+
+          {:ok, res_content} = access(decoded_content, resource_path)
+
+          acc_map <> res_content
+        end)
+
+      {:ok, file_content} = Base.url_decode64(file_content, padding: false)
+      {:ok, encoding} = access(file_metadata, "encoding", nil)
```
Neylix, line 114: You don't need to use the `access` function here as you can just pattern match the `encoding` value from the parameter

lib/archethic_web/controllers/api/web_hosting_controller/resources.ex
```diff
@@ -0,0 +1,175 @@
...
+      {:ok, file_content, encoding}
+    rescue
+      MatchError ->
+        {:error, :invalid_encoding}
```
Neylix, line 118: What is the difference between `encoding_error` and `invalid_encoding` ?

Neylix (changes requested):

Neylix:
Some remarks done, but it's working well. Good jobs !

lib/archethic_web/controllers/api/web_hosting_controller/resources.ex
apoorv-2204: For better error reporting , test get_file_content/3 should return raw file content "Invalid file encoding"

lib/archethic/transaction_chain/mem_tables_loader.ex
```diff
@@ -54,6 +54,8 @@ defmodule Archethic.TransactionChain.MemTablesLoader do
     :ok = handle_pending_transaction(tx)
     :ok = handle_transaction_recipients(tx)
 
+    # TO DO: fix bug here
+    # tx_address == nil ||
```
Neylix, line 58: Do you have a example case where the transaction address is nil ? It normally should never be the case

apoorv-2204: During node reboot

lib/archethic_web/controllers/api/web_hosting_controller/resources.ex
```diff
@@ -0,0 +1,180 @@
...
+      {:ok, file_content, encoding, mime_type, cached?, etag}
+    else
+      {:error, %Jason.DecodeError{}} ->
+        {:error, :invalid_content}
+
+      {:error, :file_not_found} ->
+        {:error, :file_not_found}
+
+      {:error, :get_metadata} ->
+        {:error, "Error: Cant access metadata and aewebversion, Reftx: #{last_address}"}
```
Neylix, line 39: Here you should use `Base.encode16` on the `last_address` otherwhise the response will be non readable

Neylix (commented):
Also there is some function that could be defined in private, mainnly in resource.ex and directory_listing.ex

Neylix (approved):

lib/archethic/transaction_chain/mem_tables_loader.ex
Neylix: I don't get this error, is it when the node restart juste after crashing or after some time ? It appears with multuple nodes or only one ?

Neylix: If you still have this error, you can remove the comment and create an issue for it please

apoorv-2204: -Sure sir  I will remove it

Neylix (approved):

PR #769, opened 2022-12-13
Pr 727 create oracle services data agregator

### LINKED ISSUE #272: Transaction already exists
### What happened?

After few minutes on running the docker-compose I got some errors regarding oracle transactions.
Some transactions are been submitted with already existing addresses.
So the systems catches them and prevent to be written.

image

### What is expected ?

We should not have the same address for existing transactions, as it may be seen as a problem of sync or key derivation

### LINKED ISSUE #727: Create an oracle data aggregator

Actually the source of an oracle is unique. But this source is not accessible anymore or sending wrong data.
In this case we need to fetch the information from multiple sources.

For each services, we should have a data aggregator which receive the data from multiple sources and aggregate them to have the most accurate value.
For the UCO Price we should have an aggregator doing the median of the received data :
[10, 12, 13, 14, 60] -> 13

### Additional context

For the implementation we may have a behavior module which delegate the call to the good service aggregator.
This aggregator may know the number of sources it should receive, and when it receive all these data then aggregate them and return the value.

netboz:
# Description

- Add value aggregator in front of services providers

Fixes #727

## Type of change

- This change requires a documentation update ( configuration file changed )

# How Has This Been Tested?

- From 3 mocked services, asserts the median is computed from the received values
- From 4 mocked services ( even number ) asserts the median from the 2 mean values is returned

# Checklist:


tenmoves:
this pr fixes https://github.com/archethic-foundation/archethic-node/issues/727 and not 272 (typo)

lib/archethic/oracle_chain/services/uco_price.ex
```diff
@@ -13,20 +13,59 @@ defmodule Archethic.OracleChain.Services.UCOPrice do
...
+            result
+
+          {:error, reason} ->
+            Logger.warning("Service : #{inspect(__MODULE__)} : Cannot fetch values from
+                provider: #{inspect(provider)} with reason : #{inspect(reason)}.")
+
+            {false, provider}
+        end
+      end)
+      |> Enum.map(fn
```
samuelmanzanera, line 32: I think you have to handle the timeout and not crash the tasks if it's happening.
You can use the option: `on_timeout: :kill_task`

samuelmanzanera:
I did try the PR and I got this error:
image

samuelmanzanera (approved):

bchamagne:
I tried to block 2 of the 3 providers and I get this result: 

```
2023-01-05 13:29:05.633 [warning] Service : Archethic.OracleChain.Services.UCOPrice : Cannot fetch values from
                provider: Archethic.OracleChain.Services.UCOPrice.Providers.CoinMarketCap with reason : [failed_connect: [{:to_address, {'coinmarketcap.com', 80}}, {:inet, [:inet], :econnrefused}], failed_connect: [{:to_address, {'coinmarketcap.com', 80}}, {:inet, [:inet], :econnrefused}]].
2023-01-05 13:29:05.703 [warning] Service : Archethic.OracleChain.Services.UCOPrice : Unexpected answer while querying provider : {:ok, {false, Archethic.OracleChain.Services.UCOPrice.Providers.CoinMarketCap}}
{:ok, %{"eur" => 0.04285300000000001, "usd" => 0.04542245773793001}}
```

The price is really far from what's on the 3rd provider (coinpaprika is at ~0.8)

bchamagne (approved):

lib/archethic/oracle_chain/services/uco_price/providers/coin_marketcap.ex
```diff
@@ -0,0 +1,72 @@
...
+  alias Archethic.OracleChain.Services.UCOPrice.Providers.Impl
+
+  @behaviour Impl
+
+  require Logger
+
+  @impl Impl
+  @spec fetch(list(binary())) :: {:ok, %{required(String.t()) => any()}} | {:error, any()}
+  def fetch(pairs) when is_list(pairs) do
+    query = 'http://coinmrketcap.com/currencies/uniris/markets/'
```
Neylix, line 13: The link is not the good one, there is a typo error. Also you should use https request instead of http. In the httpc.request function you should use the ssl verification as it is done for coingecko provider

lib/archethic/oracle_chain/services/uco_price/providers/coin_marketcap.ex
```diff
@@ -0,0 +1,72 @@
...
+          {pair, [price]}
+        else
+          :error ->
+            {:error, "invalid content"}
+
+          {:error, _} = e ->
+            e
+        end
+      end)
+      |> Enum.map(fn {:ok, val} -> val end)
```
Neylix, line 52: Here you can refactor the code to be smaller: 
```elixir
# Async task should return {:ok, {pair, [price]}}
res = 
...
|> Stream.filter(&match?({:ok, {:ok, _}}, &1))
|> Stream.map(fn {:ok, {:ok, val}} -> val end)
|> Enum.into(%{})

{:ok, res}
```

lib/archethic/oracle_chain/services/uco_price/providers/coin_paprika.ex
```diff
@@ -0,0 +1,62 @@
...
+           :httpc.request(:get, {query, []}, httpc_options, []),
+         {:ok, payload} <- Jason.decode(body) do
+      quotes =
+        payload
+        |> Enum.map(fn %{"quotes" => quotes} -> quotes end)
+
+      prices =
+        pairs
+        |> Enum.map(fn pair ->
+          {pair, Enum.map(quotes, &get_in(&1, [String.upcase(pair), "price"]))}
```
Neylix, line 43: Here there might be a problem. The `get_in` function returns `nil` if no value is found. But we cannot have a nil value in the list of prices, because aggregator cannot handle it. We should ensure there is not nil value in list of prices.

Neylix (changes requested):

lib/archethic/oracle_chain/services/uco_price/providers/coin_paprika.ex
tenmoves: what should be the default value in this case ?

lib/archethic/oracle_chain/services/uco_price/providers/coin_marketcap.ex
netboz: Hi Neylix. Indeed, sorry, yesterday while testing I introduced this typo to test resilience of agregator when service crash. I will revert to correct address.

lib/archethic/oracle_chain/services/uco_price/providers/coin_paprika.ex
Neylix: Yes, if there is a nil value, just remove it from the list and if there is only nil value, return an error or an empty list

lib/archethic/oracle_chain/services/uco_price.ex
```diff
@@ -13,20 +13,69 @@ defmodule Archethic.OracleChain.Services.UCOPrice do
...
+          end
+        end,
+        on_timeout: :kill_task
+      )
+      |> Stream.filter(fn
+        {:ok, {:ok, _result}} ->
+          true
+
+        other ->
+          Logger.warning(
```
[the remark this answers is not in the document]
netboz: Hi Neylix. If the service doing the fech comes to crash, then an `{:exit, reason}` is returned by `async_stream_nolink`. This clause then catches it. The purpose of this close is actually to catch all unexpected returns from `async_stream_nolink`.

lib/archethic/oracle_chain/services/uco_price.ex
```diff
@@ -13,20 +13,69 @@ defmodule Archethic.OracleChain.Services.UCOPrice do
...
+            "Service : #{inspect(__MODULE__)} : Unexpected answer while querying provider : #{inspect(other)}"
+          )
+
+          false
+      end)
+      |> Enum.map(fn
+        {_, {_, result = %{}}} ->
+          result
+
+        other_service_answer_format ->
```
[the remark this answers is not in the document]
netboz: I agree, but if it comes a service returns badly formatted values, then this will crash the scheduler, I wouldn't take the risk to rely on having always valid format from queried services.

lib/archethic/oracle_chain/services/uco_price.ex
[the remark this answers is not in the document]
Neylix: The function `match?` will return false if the clause does not match, `match?({:ok, {:ok, _}}, {:exit, reason})` return false then.

lib/archethic/oracle_chain/services/uco_price.ex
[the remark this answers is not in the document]
Neylix: I think the provider cannot return badly formatted value as it is defined in the `fetch` function spec of the provider, so if a provider can return a non formatted value, dialyzer will raise a warning

Neylix (approved):

Issue #729, opened 2022-11-30
Fetch UCO price from CoinMarketCap


Neylix:
### Is your feature request related to a problem?

As #727 will permit to have multiple data source for an oracle value, we should get the uco price from other source.

### Describe the solution you'd like

CoinMarketCap do not explose a free API to get currency price. But we can parse the HTML page containing the UCO price.
https://coinmarketcap.com/currencies/uniris/markets/
On this page there is multiple price information, one global and one per exchange. We could get all there price information and add them into the aggregator

### Additional context

For HTML parsing we can use Floki : https://hexdocs.pm/floki/Floki.html

PR #772, opened 2022-12-13
Fetch UCO price from coinPaprika and coinMarketCap

### LINKED ISSUE #728: Fetch UCO price from coinPaprika

As #727 will permit to have multiple data source for an oracle value, we should get the uco price from other source.

Coin paprika expose a free API to fetch the price of a currency from multiple platform:

https://api.coinpaprika.com/v1/coins/uco-uniris/markets?quotes=USD,EUR

For each platform returned by the API, we can add it's price into the aggregator

### Additional context

_No response_

tenmoves:
# Description

Fixes #728 #729

## Type of change

# How Has This Been Tested?

-  Unit tests

# Checklist:


lib/archethic/oracle_chain/services/uco_price/providers/coin_marketcap.ex
```diff
@@ -0,0 +1,71 @@
...
+  def fetch(pairs) when is_list(pairs) do
+    query = 'http://coinmarketcap.com/currencies/uniris/markets/'
+
+    httpc_options = [
+      connect_timeout: 1000,
+      timeout: 2000
+    ]
+
+    returned_prices =
+      Enum.map(pairs, fn pair ->
```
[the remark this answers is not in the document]
tenmoves: yes thank you

lib/archethic/oracle_chain/services/uco_price/providers/coin_gecko.ex
```diff
@@ -34,7 +34,8 @@ defmodule Archethic.OracleChain.Services.UCOPrice.Providers.Coingecko do
            :httpc.request(:get, {query, []}, httpc_options, []),
          {:ok, payload} <- Jason.decode(body),
          {:ok, prices} <- Map.fetch(payload, "archethic") do
-      {:ok, prices}
+      formatted_prices = Enum.map(prices, fn {pair, price} -> {pair, [price]} end)
+      {:ok, formatted_prices}
```
samuelmanzanera, line 38: What is the reason of this formatting?

tenmoves: because one service returns a list of prices from different sources, so I thought it's better to aggregate the prices at the top level and not on each provider

samuelmanzanera: OK I get it.
In terms of aggregation we will have more granularity and precision.
However, we have to think it's the weight of the incorrect values will impact too much the final aggregation

tenmoves: we have a trust_level provided by the api, the values are high low and not_provided

tenmoves: on one provider, but the best would be to discard any value that has a variance greater than let's say 10%.
WDYT ?

tenmoves: but for that we need to trust one source blindly

samuelmanzanera: Finally the median of all the values will remove the outlier and not blind them

PR #775, opened 2022-12-13
use plug attack to add a throttle for api to handle ddos attacks

### LINKED ISSUE #768: Add anti DDOS for abusive requests on the API
We can use lib like this to protect endpoints for abusive queries

https://hex.pm/packages/plug_attack

tenmoves:
# Description

Fixes #768 

## Type of change

# How Has This Been Tested?

- We query an api 10 time then we make sure on the 11 time that we get a forbidden http code

# Checklist:


tenmoves:
@bchamagne issue handled in new version

bchamagne:
I retested to upload the archethic-website and I sometimes get a FORBIDDEN when estimating the fees. 
I guess there is still some number tweaking to do. 

Screenshot_2022-12-19_17-01-32
In the screenshot above, you can see on the right side, 3 attempts to deploy, 1 success (yellow text) and 2 failures

config/config.exs
```diff
@@ -64,6 +64,16 @@ config :archethic, :transaction_data_content_max_size, 3_145_728
...
 
+config :archethic, :throttle,
+  by_ip: [
+    period: 1000,
+    limit: 150
+  ],
+  by_ip_and_path: [
+    period: 1000,
+    limit: 20
+  ]
```
tenmoves, line 75: @bchamagne can you give me the numbers for the test you did to compare with the configuration ?

tenmoves: or how to reproduce it locally

tenmoves: this is normal, the website contains 266 files but the limit is at 150 no it's normal for it to crash

bchamagne: I tried 
```
config :archethic, :throttle,
  by_ip: [
    period: 1000,
    limit: 300
  ],
  by_ip_and_path: [
    period: 1000,
    limit: 300
  ]
```
and I get the forbidden. People will not be able to upload to AEWeb if we don't relax the numbers

lib/archethic_web/explorer_router.ex
```diff
@@ -94,6 +99,11 @@ defmodule ArchethicWeb.ExplorerRouter do
     )
   end
 
+  scope "/api" do
```
Neylix, line 102: This scope should be before the other api scope, because the last `forward("/"...)` will catch all path after api including web_hosting. So `/api/web_hosting` is not accessible

Neylix (approved):

Neylix (approved):

Issue #760, opened 2022-12-09
Forward the welcome node when forwarding tx


samuelmanzanera:
If we send a transaction to a non-authorized node, the transaction will be forwarded to the authorized nodes to send the transaction to the validation nodes.
However, the welcome node is not forwarded, so the client gets a timeout.
We need to keep track of the original welcome node.
For this purpose, we can use the sender public key from the %NewTransaction{} message.

PR #783, opened 2022-12-16
 Forward the welcome node when forwarding tx#760

apoorv-2204:
# Description

Resolves the 760 by using sender public key, and with default params as Crypto.last_node_key , if not given,

```
 # Since welcome node is not anymore constant, as we want to unauthorised
  # nodes to do some labor. Following bootstrapping, the txn of a new node
  # is sent, with the welcome node being the same new node whose information
  # does not exist in the network. Thus solitary and distributed workflows
  # are more susceptible to failures.
```
- [x] #760 

## Type of change

# How Has This Been Tested?

- manual

    - [x] Standalone workflow testing
    - [x] distributed workflow testing
    - [x] faucet testing on unauth nodes
    - [x]  aeweb cli on unauth nodes

# Checklist:


lib/archethic/bootstrap/transaction_handler.ex
```diff
@@ -99,7 +99,7 @@ defmodule Archethic.Bootstrap.TransactionHandler do
         condition inherit: [
           # We need to ensure the type stays consistent
           type: node,
-        
+
```
samuelmanzanera, line 102: You should discard this change, as the inherit constraints would not be valid

lib/archethic_web/controllers/faucet_controller.ex
```diff
@@ -78,55 +78,52 @@ defmodule ArchethicWeb.FaucetController do
...
-            uco: %UCOLedger{
-              transfers: [
-                %UCOLedger.Transfer{
-                  to: recipient_address,
-                  amount: 10_000_000_000
-                }
-              ]
+  @unit_uco 100_000_000
+  @max_uco 30
+  @uco_limit @max_uco * @unit_uco
```
samuelmanzanera, line 101: why 30 ?

lib/archethic_web/controllers/faucet_controller.ex
```diff
@@ -78,55 +78,52 @@ defmodule ArchethicWeb.FaucetController do
...
-
-    with {:ok, last_address} <-
-           Archethic.get_last_transaction_address(pool_gen_address),
+    with {:ok, last_address} <- Archethic.get_last_transaction_address(pool_gen_address),
          {:ok, last_index} <- Archethic.get_transaction_chain_length(last_address) do
-      create_transaction(last_index, curve, recipient_address)
+      {:ok, Transaction.new(:transfer, get_txn_data(recipient_address), @pool_seed, last_index)}
     else
-      {:error, _} = e ->
+      e ->
```
samuelmanzanera, line 94: If you refactor this function, the else statement does not bring value here. You could let it return any value which will be handled in the above fn.

lib/archethic_web/controllers/faucet_controller.ex
```diff
@@ -78,55 +78,52 @@ defmodule ArchethicWeb.FaucetController do
...
+          ]
+        }
+      }
+    }
+  end
 
-    tx_address = tx.address
-    TransactionSubscriber.register(tx_address, System.monotonic_time())
+  defp send_txn(txn, recipient_address) do
+    txn_addr = txn.address
```
samuelmanzanera, line 118: You should use pattern matching in the function argument. This will be more elegant ;)

lib/archethic/bootstrap.ex
```diff
@@ -307,7 +307,13 @@ defmodule Archethic.Bootstrap do
     end
 
     tx =
-      TransactionHandler.create_node_transaction(ip, port, http_port, transport, reward_address)
+      TransactionHandler.create_node_transaction(
```
samuelmanzanera, line 310: Why do you change the formatting ?
We are all using `mix format` but this was not handled by the standard formatter.
To avoid confusion and longer PR review, please use the standard one provided by Mix.

lib/archethic_web/controllers/faucet_controller.ex
apoorv-2204: sry

lib/archethic/mining/standalone_workflow.ex
```diff
@@ -42,11 +42,15 @@ defmodule Archethic.Mining.StandaloneWorkflow do
 
   def init(arg) do
     tx = Keyword.get(arg, :transaction)
+    welcome_node = Keyword.get(arg, :welcome_node)
```
bchamagne, line 45: I recommend `Keyword.fetch!/2` since it's not optional at all.

lib/archethic_web/controllers/faucet_controller.ex
```diff
@@ -78,47 +78,44 @@ defmodule ArchethicWeb.FaucetController do
     end
   end
 
-  defp transfer(
-         recipient_address,
-         curve \\ Crypto.default_curve()
-       )
-       when is_bitstring(recipient_address) do
-    {gen_pub, _} = Crypto.derive_keypair(@pool_seed, 0, curve)
+  @curve Crypto.default_curve()
```
bchamagne, line 81: FYI, when you do this, you add a _compilation time dependency_  `Crypto -> FaucetController`, meaning everytime you compile `Crypto`, you will have to compile `FaucetController`. 

I suggest to avoid this, and directly use it in the function content so it is a _runtime dependency_.

lib/archethic/mining/standalone_workflow.ex
```diff
@@ -199,7 +203,7 @@ defmodule Archethic.Mining.StandaloneWorkflow do
       transaction_address: Base.encode16(tx_address)
     )
 
-    Logger.debug("Notify error back to the welcome node",
+    Logger.error("Notify error back to the welcome node",
```
Neylix, line 206: This log should stay as debug I think, it is only an information that the node is notifying the welcome node, that is not an error. The error log while validating the transaction is already done before

Neylix (commented):
Appart the Logger level, it looks good to me.
I don't understand why you refactor the faucet controller as there is no link with the issue but anyway

lib/archethic/mining/standalone_workflow.ex
apoorv-2204: done 👍

lib/archethic_web/controllers/faucet_controller.ex
apoorv-2204: reverted to 100

lib/archethic_web/controllers/faucet_controller.ex
apoorv-2204: done

lib/archethic_web/controllers/faucet_controller.ex
apoorv-2204: 👍

lib/archethic/bootstrap.ex
apoorv-2204: my bad it was a change due to debug lines I put.
anyway I removed it

lib/archethic/mining/standalone_workflow.ex
apoorv-2204: 👍

lib/archethic_web/controllers/faucet_controller.ex
apoorv-2204: 👍

apoorv-2204:
> Appart the Logger level, it looks good to me. I don't understand why you refactor the faucet controller as there is no link with the issue but anyway


> level, it looks good to me.

Reverted 👍

samuelmanzanera (approved):

PR #784, opened 2022-12-16
Explorer optimizations (read DB before asking network when possible)

### LINKED ISSUE #730: For explorer, fetch data from DB if they are available

Explorer will be consulted by a lot of people, and so the node will request a lot of informations on the network.
To reduce the network charge and enhance the explorer speed, we should first check if the information needed is accessible from the node DB before requesting it on the network.

Actually the explorer uses 2 modules to fetch informations `Archethic` and `BeaconChain`.
We could remove the direct request on `BeaconChain` and only ask informations on `Archethic` module.

As it, in the `Archethic` module, we can try to fetch the information from DB and return it if it exists, if not then we request the network using quorum.

### Additional context

_No response_

bchamagne:
# Description

Fetch explorer's data from DB if it is available on current node. This avoid plenty of network requests.
**ALSO**: Refactored a bit the pagination to be consistent across all pages

Fixes #730

## Type of change

# How Has This Been Tested?

Tested the explorer 

# Checklist:

### DIFF
--- a/lib/archethic.ex
+++ b/lib/archethic.ex
@@ -332,12 +332,7 @@ defmodule Archethic do
   def fetch_summaries_aggregate(date) do
     nodes = P2P.authorized_and_available_nodes()
 
-    is_a_beacon_storage_node? =
-      date
-      |> Crypto.derive_beacon_aggregate_address()
-      |> Election.beacon_storage_node?(date, Crypto.first_node_public_key(), nodes)
-
-    if is_a_beacon_storage_node? do
+    if is_a_beacon_storage_node?(date, nodes) do
       BeaconChain.get_summaries_aggregate(date)
     else
       BeaconChain.fetch_summaries_aggregate(date, nodes)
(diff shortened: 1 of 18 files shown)


bchamagne:
As discussed with @Neylix , the same optimisation can be used on many functions from the `Archethic` module.
Plus, I should not run an election to determine if the node has the local data, because elections change over time (they are based on the currently available nodes).

bchamagne:
and of course I messed up with the rebase

lib/archethic.ex
```diff
@@ -321,4 +335,132 @@ defmodule Archethic do
...
+    end)
+    |> Stream.uniq_by(& &1.address)
+    |> Enum.sort_by(& &1.timestamp, {:desc, DateTime})
+  end
+
+  @doc """
+  Check if a transaction exists at address
+  """
+  @spec transaction_exists?(binary()) :: boolean()
+  def transaction_exists?(address) do
```
Neylix, line 420: This function can check the DB before asking network, using `TransactionChain.transaction_exists?` if it return false, then we ask the network. This function can check both in db storage or io storage

Neylix (changes requested):

Neylix (approved):

PR #786, opened 2022-12-16
Improve atomic commitment replication

### LINKED ISSUE #779: Improve atomic commitment for replication
For now, the atomic commitment is managed on the validation nodes side (aka before the replication requests).
Hence, the clients had to determine the level of confidence for the transaction's validity.

It might happen when the transaction is considered valid for the validation nodes but no, anymore for the storage nodes (for example due to the address's change)

The idea is to use some kind of [2PC](https://en.wikipedia.org/wiki/Two-phase_commit_protocol) approach.

Where each validation nodes responsible for a sub-shard (branch of the replication tree) will request a validation (`vote`) from the replication nodes. 
Once the sub-graph produces an atomic commitment of the transaction's validation, this validation node will notify other validator nodes about the success of its branch.
Once all the validation nodes reach the final atomic commitment of the entire replication tree, they can request the replication (`commit`) of the transaction. 
After timeout, both storage and validation should abort and stop the transaction's validation.

By this way when the clients will receive confirmations they will receive confirmations for a complete atomic commitment of the transaction.

This requires to hold on memory a kind of transaction memory pool awaiting the commit message.

samuelmanzanera:
# Description

Fixes #779 

## Type of change

# Checklist:

### DIFF
--- a/lib/archethic/mining/distributed_workflow.ex
+++ b/lib/archethic/mining/distributed_workflow.ex
@@ -30,17 +30,26 @@ defmodule Archethic.Mining.DistributedWorkflow do
   alias Archethic.P2P.Message.AddMiningContext
   alias Archethic.P2P.Message.CrossValidate
   alias Archethic.P2P.Message.CrossValidationDone
+  alias Archethic.P2P.Message.Ok
   alias Archethic.P2P.Message.NotifyPreviousChain
+  alias Archethic.P2P.Message.NotifyReplicationValidation
   alias Archethic.P2P.Message.ReplicateTransactionChain
+  alias Archethic.P2P.Message.ReplicatePendingTransactionChain
   alias Archethic.P2P.Message.ReplicateTransaction
+  alias Archethic.P2P.Message.ReplicationError
+  alias Archethic.P2P.Message.ValidateTransaction
   alias Archethic.P2P.Message.ValidationError
   alias Archethic.P2P.Node
+
+  alias Archethic.TaskSupervisor
 
   alias Archethic.TransactionChain
   alias Archethic.TransactionChain.Transaction
   alias Archethic.TransactionChain.Transaction.CrossValidationStamp
   alias Archethic.TransactionChain.Transaction.ValidationStamp
   alias Archethic.TransactionChain.TransactionSummary
+
+  alias Archethic.Utils
 
   require Logger
 
@@ -122,6 +131,14 @@ defmodule Archethic.Mining.DistributedWorkflow do
   @spec add_cross_validation_stamp(worker_pid :: pid(), stamp :: CrossValidationStamp.t()) :: :ok
   def add_cross_validation_stamp(pid, stamp = %CrossValidationStamp{}) do
     GenStateMachine.cast(pid, {:add_cross_validation_stamp, stamp})
+  end
+
+  @doc """
+  Add a replication validation in the mining process
+  """
+  @spec add_replication_validation(worker_pid :: pid(), node_public_key :: Crypto.key()) :: :ok
+  def add_replication_validation(pid, node_public_key) do
+    GenStateMachine.cast(pid, {:add_replication_validation, node_public_key})
   end
 
   defp get_context_timeout(:hosting), do: Message.get_max_timeout()
@@ -635,6 +652,7 @@ defmodule Archethic.Mining.DistributedWorkflow do
         from_state,
         :replication,
         data = %{
+          node_public_key: node_public_key,
           context:
             context = %ValidationContext{
               transaction: %Transaction{address: tx_address, type: type}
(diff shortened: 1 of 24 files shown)


lib/archethic/p2p/message.ex
```diff
@@ -1938,6 +1968,48 @@ defmodule Archethic.P2P.Message do
...
+
+      {:error, :transaction_already_exists} ->
+        %ReplicationError{address: tx.address, reason: :transaction_already_exists}
+
+      {:error, reason} ->
+        %ReplicationError{address: tx.address, reason: reason}
+    end
+  end
+
+  def process(%ReplicatePendingTransactionChain{address: address}, sender_public_key) do
```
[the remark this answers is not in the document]
samuelmanzanera: The idea is also to reduce bandwidth..

lib/archethic/mining/standalone_workflow.ex
```diff
@@ -161,15 +166,49 @@ defmodule Archethic.Mining.StandaloneWorkflow do
...
-    })
+    message = %ValidateTransaction{
+      transaction: validated_tx
+    }
+
+    results =
+      Task.Supervisor.async_stream_nolink(
+        Archethic.TaskSupervisor,
+        replication_nodes,
+        &P2P.send_message(&1, message),
```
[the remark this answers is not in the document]
samuelmanzanera: The shortcut is done internally

lib/archethic/p2p/message.ex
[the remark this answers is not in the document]
bchamagne: oh right, of course!

bchamagne (approved):
I tested with 3 nodes and the replication is doing fine. 
I also checked that TransactionPool is getting cleaned up.

lib/archethic/replication/transaction_pool.ex
```diff
@@ -0,0 +1,76 @@
...
+       clean_ref: clean_ref,
+       transactions: %{}
+     }}
+  end
+
+  def handle_cast(
+        {:add_transaction, tx = %Transaction{address: address, type: type}},
+        state = %{timeout_duration: timeout_duration}
+      ) do
+    timeout = DateTime.add(DateTime.utc_now(), timeout_duration, :millisecond)
```
bchamagne, line 40: naming: expiration instead of timeout

lib/archethic/mining/distributed_workflow.ex
```diff
@@ -939,6 +978,86 @@ defmodule Archethic.Mining.DistributedWorkflow do
...
+        Archethic.TaskSupervisor,
+        storage_nodes,
+        fn node ->
+          {P2P.send_message(node, message), node}
+        end,
+        ordered: false,
+        on_timeout: :kill_task,
+        timeout: Message.get_timeout(message) + 2000
+      )
+      |> Stream.filter(&match?({:ok, _}, &1))
```
Neylix, line 1010: Here this filter may cause problem.
If a node do not respond to the message, this filter removes it from the list. Then we add the remaining nodes public key in the validation context. But as there is one node never added to the validation context, the function `ValidationContext.enough_replication_validations?` will never return true as a node is missing.
So if a node do not respond to the `ValidateTransaction` message, we already now that the transaction replication will never happen.
There is 2 way to think with this problem,
- We want all storage node to respond this message, so if one do not respond we can directly enter in replication_error.
- We want to continue even if a storage node do not respond this message, so instead of sending the public of the responding storage node, the validation node can inform the other validation nodes that it's sub shard is ok for replication. And then we start the replication if all the validation nodes told their sub shard is ok.

lib/archethic/mining/distributed_workflow.ex
```diff
@@ -939,6 +978,86 @@ defmodule Archethic.Mining.DistributedWorkflow do
...
+                                                                }} ->
+        P2P.broadcast_message(validation_nodes, %NotifyReplicationValidation{
+          address: validated_tx.address,
+          node_public_key: first_public_key
+        })
+      end)
+      |> Stream.run()
+
+      new_context
+    else
```
Neylix, line 1046: If there is a replication error returned by one of the storage node, the validation node enter in replication_error state. But the other validations nodes are waiting for this node to send the NotifyReplicationValidation message. So I think if there is a replication error, we should also notify other validation nodes that the sub shard is invalid and so other validation nodes can also enter in replication_error state. Actually they will just wait until the mining timeout of 1min is reached

Neylix (changes requested):

lib/archethic/mining/distributed_workflow.ex
samuelmanzanera: Yes, we can use the`ReplicationError` message, which is already handled by the process

lib/archethic/mining/distributed_workflow.ex
samuelmanzanera: Nicely seen..
Indeed this is one pitfall of this strict approach. 
Indeed in the current implementation the shard would be invalidated if one replication node which was supposed to be available does not respond. Causing the other nodes to await the timeout. While this seems acceptable in term of transaction validation point of view, we might to inform the user about the transaction failure earlier. 

We have to support fault tolerance in all cases. The shard size can be viewed as an overbooking against the locally available nodes. 
So the 2nd approach might be more suited for the case, as the unavailability of the node is an hazard, considered as benignly fault and not intentional/malicious behavior.

Issue #698, opened 2022-11-17
Upnp port forwarding fail on local ip change


Neylix:
### Describe the problem you discovered

Actually we use miniupnpc for port forwarding, it tells to the router to redirect incoming request on a specific port to the node local ip.
But if the local ip of the node changes, the new upnp request will fail because the router already has a redirection on this port

### Describe the solution you'd like

Before using upnp to forward the port, we should check if there is already a redirection on this port.
If there is no redirection -> continue and forward port
If there is a redirection on the same local ip -> continue and forward port (to update time limit)
If there is a redirection on a different local ip -> remove the existing redirection then continue and forward port

roychowdhuryrohit-dev:
Why not use mac binding to disable local IP change?

PR #793, opened 2022-12-29
UPNP Port Forwarding Fail On Local IP Change#698

apoorv-2204:
# Description

Attempt to resolve  upnp_port_forwarding_fail_on_local_ip_change#698.
- If there is no redirection -> continue and forward port -> `Archethic.Networking.IPLookup.NATDiscovery.MiniUPNP.open(port)`
- If there is a redirection on the same local ip -> continue and forward port (to update time limit)
   miniupnpc duration set to 0 as for mapping for a lifetime
- If there is a redirection on a different local ip -> remove the existing redirection then continue and forward port
   Archethic.Networking.IPLookup.NATDiscovery.MiniUPNP, we expect this to fail and capture conflict mapping entry
error .revoke that port and assigns to the node

Fixes 

- [ ] #698 

## Type of change

# How Has This Been Tested?
- When the port forwarding fails , we capture the error, and remove port rule , and redo the rule.
 Screenshot shows working of code, trying to imitate that condition ,we expect same when node fails to port forward.

proof

# Checklist:

- My code not follows the style guidelines of this project
- I have not performed a self-review of my own code
- I have commented my code, particularly in easy-to-understand areas
- My changes generate 20 new warnings


lib/archethic/networking/ip_lookup/nat_discovery/miniupnp.ex
```diff
@@ -43,8 +43,31 @@ defmodule Archethic.Networking.IPLookup.NATDiscovery.MiniUPNP do
     end
   end
 
+  @spec do_open_port(:inet.ip_address(), non_neg_integer()) :: :ok | :error
   defp do_open_port(local_ip, port) do
-    opts = [
+    with {reason, status} when status != 0 <- System.cmd(@upnpc, map_query(local_ip, port)),
```
bchamagne, line 48: I find this pattern very hard to read

samuelmanzanera, line 48: I think using the happy path firstly, we would make more sense.
The error branching should be done on the else, I think.

bchamagne (approved):
Just read the code, I did not test it.

lib/archethic/networking/ip_lookup/nat_discovery/miniupnp.ex
samuelmanzanera: We could use something like this:
```elixir
case System.cmd(@upnpc, map_query(local_ip, port)) do
  {_, 0} ->
    :ok 
  {reason, status} ->
    reason
    |> parse_reason()
    |> handle_error()
 end
```

lib/archethic/networking/ip_lookup/nat_discovery/miniupnp.ex
```diff
@@ -54,18 +77,44 @@ defmodule Archethic.Networking.IPLookup.NATDiscovery.MiniUPNP do
...
-    case System.cmd(@upnpc, opts) do
-      {_, 0} ->
-        :ok
+  @spec parse_reason(String.t()) ::
+          {:error, :conflict_in_mapping_entry | :unknown}
+  defp parse_reason(reason) do
+    # upon more condtions , refactor with cond do end
+    if Regex.scan(~r/ConflictInMappingEntry/, reason, capture: :all) != [] do
+      Logger.warning("Port is employed to another host.")
+      {:error, :conflict_in_mapping_entry}
```
samuelmanzanera, line 92: The error tuple is more used to help to branch between success and failure.
Here you just want to format error's reason.
I think you could return simplify the error's reason, so you could easily use this function in a pipeline.

lib/archethic/networking/ip_lookup/nat_discovery/miniupnp.ex
```diff
@@ -54,18 +77,44 @@ defmodule Archethic.Networking.IPLookup.NATDiscovery.MiniUPNP do
...
+  end
 
-      {reason, _status} ->
-        Logger.debug(reason)
-        :error
+  @spec handle_error(:conflict_in_mapping_entry | :unknown, :inet.ip_address(), non_neg_integer()) ::
+          :ok | :error
+  defp handle_error(:conflict_in_mapping_entry, _, port) do
+    case System.cmd(@upnpc, revoke_query(port)) do
+      {_, 0} -> :ok
```
samuelmanzanera, line 102: I think you could leverage the reopening in the error's handling.

samuelmanzanera (changes requested):

apoorv-2204:
image

lib/archethic/networking/ip_lookup/nat_discovery/miniupnp.ex
```diff
@@ -43,8 +43,30 @@ defmodule Archethic.Networking.IPLookup.NATDiscovery.MiniUPNP do
     end
   end
 
-  defp do_open_port(local_ip, port) do
-    opts = [
+  @spec do_open_port(:inet.ip_address(), non_neg_integer()) :: :ok | :error
+  def do_open_port(local_ip, port, retires \\ 2)
```
samuelmanzanera, line 47: You might keep it private

lib/archethic/networking/ip_lookup/nat_discovery/miniupnp.ex
```diff
@@ -43,8 +43,30 @@ defmodule Archethic.Networking.IPLookup.NATDiscovery.MiniUPNP do
...
+        do_open_port(local_ip, port, retires - 1)
+    end
+  end
+
+  @protocol "tcp"
+  @spec map_query(:inet.ip_address(), non_neg_integer()) :: [String.t()]
+  defp map_query(local_ip, port) do
+    [
+      # description
+      "-e",
```
samuelmanzanera, line 68: Last time we got some issue with this parameter. Need to check with @Neylix

samuelmanzanera (changes requested):

lib/archethic/networking/ip_lookup/nat_discovery/miniupnp.ex
```diff
@@ -43,8 +43,27 @@ defmodule Archethic.Networking.IPLookup.NATDiscovery.MiniUPNP do
...
   end
 
-  defp do_open_port(local_ip, port) do
-    opts = [
+  @spec do_open_port(:inet.ip_address(), non_neg_integer()) :: :ok | :error
+  defp do_open_port(local_ip, port, retires \\ 2)
+
+  defp do_open_port(_local_ip, _port, 0), do: :error
+
+  defp do_open_port(local_ip, port, retires) do
```
bchamagne, line 51: typo: `retries`

ps: very good idea

lib/archethic/networking/ip_lookup/nat_discovery/miniupnp.ex
```diff
@@ -43,8 +43,27 @@ defmodule Archethic.Networking.IPLookup.NATDiscovery.MiniUPNP do
...
+    case System.cmd(@upnpc, map_query(local_ip, port)) do
+      {_, 0} ->
+        :ok
+
+      {reason, status} when status != 0 ->
```
samuelmanzanera, line 56: I think you can omit the status check

samuelmanzanera (commented):
Code seems ok

Neylix (approved):

PR #794, opened 2022-12-30
Fix bugs

Neylix:
# Description

Fix some bugs
- Uncatched error in Replication while fetching previous transaction causing replication crash
- When writing a transaction chain in DB, remove ALL transaction of this chain from KOLedger
- To calculate node availabilty we use a wheighted average to increase recent slot weight, but the slot list was not sorted so the wheight was not applied to the expected slot
- Ecto crash when having error somewhere (i.e content oversized) and the transaction contains ownerships
- Since the `GetTransactionChain` message can return in asc or desc order, the conflict resolver was not accurate : 

let's asume a chain TX1 -> TX2 -> TX3 -> TX4 -> TX5 -> TX6 -> TX7 and the nb of tx returned is 5 by request in desc order
if a node didn't replicate the last transaction we could have a result like 
node1 : 
```elixir
%{
  transactions: [TX6, TX5, TX4, TX3, TX2],
  more?: true,
  paging_state: TX2
}
```
node2 : 
```elixir
%{
  transactions: [TX7, TX6, TX5, TX4, TX3],
  more?: true,
  paging_state: TX3
}
```
node3: same as node2

The good return is node2 and node3 result, but as node1 and other have more? at true and length of transactions list is the same, so the previous conflict resolver take the first element of the returns list which can be node1 with the wrong result.
So to fix this I added a sort on the timestamp of the first transaction of the list (regarding the order requested)

## Type of change

# Checklist:


lib/archethic/transaction_chain.ex
```diff
@@ -210,7 +210,7 @@ defmodule Archethic.TransactionChain do
     } = Enum.at(sorted_chain, 0)
 
     DB.write_transaction_chain(sorted_chain)
-    KOLedger.remove_transaction(tx_address)
+    Enum.each(sorted_chain, &KOLedger.remove_transaction(&1.address))
```
tenmoves, line 213: why not remove transactions in parallel

lib/archethic/transaction_chain.ex
```diff
@@ -793,16 +793,35 @@ defmodule Archethic.TransactionChain do
...
+                ) == :lt
+
+              :desc ->
+                DateTime.compare(
+                  List.first(&1.transactions).validation_stamp.timestamp,
+                  List.first(&2.transactions).validation_stamp.timestamp
+                ) == :gt
+            end
+          end
+        end
```
tenmoves, line 820: maybe we can extract this comparator for more readability

lib/archethic_web/controllers/api/transaction_payload.ex
```diff
@@ -49,6 +49,8 @@ defmodule ArchethicWeb.API.TransactionPayload do
 
   def to_map(%{changes: changes}, acc) do
     Enum.reduce(changes, acc, fn {key, value}, acc ->
+      value = if key == :authorizedKeys, do: format_authorized_keys(value), else: value
```
tenmoves, line 52: since you are already doing format_authorized_keys, maybe it can take both key and value and pattern match on the specific key you need. I think it will be more readable.

tenmoves (approved):

lib/archethic/transaction_chain.ex
```diff
@@ -793,16 +793,39 @@ defmodule Archethic.TransactionChain do
...
+        # Prioritize more? at true
+        # then length of transaction list
+        # then regarding order, the oldest or newest transaction timestamp
+        # of the first element of the list
+        &with false <- &1.more? and !&2.more?,
+              false <- length(&1.transactions) > length(&2.transactions) do
+          if Enum.empty?(&1.transactions) do
+            false
+          else
+            case order do
```
samuelmanzanera, line 810: We could factorize to reduce boilerplate. 
And use the case to determine the right comparison operator.

lib/archethic/transaction_chain.ex
```diff
@@ -793,16 +793,39 @@ defmodule Archethic.TransactionChain do
...
+
     conflict_resolver = fn results ->
       results
       |> Enum.sort(
-        &((&1.more? and !&2.more?) or length(&1.transactions) > length(&2.transactions))
+        # Prioritize more? at true
+        # then length of transaction list
+        # then regarding order, the oldest or newest transaction timestamp
+        # of the first element of the list
+        &with false <- &1.more? and !&2.more?,
```
samuelmanzanera, line 805: What happens with the `else` clause ? Any catch ?

Neylix: The else clause will return true and will be catched by the sort function and apply the sort

samuelmanzanera: Indeed the result of the with is the opposite of the clauses, so true

PR #802, opened 2023-01-04
Pr 57 validate smart contract calls

### LINKED ISSUE #57: Validate smart contract calls
## Rationale

When a transaction is targeting a smart contract through the `recipients` field in the transaction data, we should provide a way to check if the transaction would be valid before to send.
Hence, clients can call this endpoint to be sure the transaction will pass.

## Why not doing it during transaction validation ?
Because if the transaction calls are invalid, we don't want a user to spam the network making DDOS attack for nothing and do not pay the fee as the transaction would be not valid.
So by letting the transaction be validated and refused on the smart contract side, we are using the fee as a security mechanism for malicious attacks towards smart contracts.

## How to do it

We should provide an API to check the validity of the smart contract call. 
This API should:
- [ ]  Download transaction's code of the smart contract targeted
- [ ]  Interpret the transaction conditions with the pending transaction as input
- [ ] Return errors or success

netboz:
# Description

Provide an endpoint to pre-validate smart contract execution

Fixes #57

## Type of change

# How Has This Been Tested?

- Test A
- Test B

# Checklist:


tenmoves (approved):

lib/archethic/contracts.ex
```diff
@@ -165,6 +167,37 @@ defmodule Archethic.Contracts do
     end
   end
 
+  def is_valid_contract?(
```
samuelmanzanera, line 170: I think we might improve the naming of this function, as it suggest the validation of the contract itself. (It's a but confusing).
But finally, the purpose is to validate the call or the execution of a contract based on some input (incoming transaction)

lib/archethic/contracts.ex
```diff
@@ -165,6 +167,37 @@ defmodule Archethic.Contracts do
...
+        prev_tx = %Transaction{data: %TransactionData{code: code}},
+        next_tx = %Transaction{},
+        date = %DateTime{}
+      ) do
+    case Interpreter.parse(code) do
+      {:ok,
+       %Contract{
+         triggers: triggers,
+         conditions: %{inherit: inherit_conditions}
```
samuelmanzanera, line 179: also you should check the transaction's condition instead of the inherit's condition.

samuelmanzanera (changes requested):

lib/archethic/contracts.ex
```diff
@@ -165,6 +167,41 @@ defmodule Archethic.Contracts do
...
+        }
+
+        with :ok <- validate_conditions(transaction_conditions, constants),
+             :ok <- validate_triggers(triggers, next_tx, date),
+             %Transaction{} <-
+               ActionInterpreter.execute(Map.fetch!(triggers, :transaction), constants) do
+          %{"valid" => true, "address" => prev_tx}
+        else
+          {:error, reason} ->
+            %{"valid" => false, "address" => prev_tx, "reason" => reason}
```
bchamagne, line 197: May I suggest to not use the `%{"field" => value}` format?  (and use `%{field: value}` instead)
I find it really error prone and we should only deal with that format when we parse json user input.

netboz: Ok 👍fixed.

bchamagne:
When I go to localhost:4000/api/transaction/contract/simulator I get this error: 

Screenshot 2023-01-12 at 15-20-17 Phoenix ActionClauseError at GET _api_transaction_contract_simulator

test/archethic_web/controllers/api/transaction_controller_test.exs
```diff
@@ -99,4 +105,272 @@ defmodule ArchethicWeb.API.TransactionControllerTest do
...
+      conn = post(conn, "/api/transaction/contract/simulator", new_tx)
+
+      assert(match?([%{"valid" => false}], json_response(conn, 200)))
+    end
+
+    test "should indicate when body fails changeset validation", %{conn: conn} do
+      code = """
+      condition inherit: [
+        type: transfer,
+        content: \"hello\",
```
bchamagne, line 268: I think there's no need for antislashes here

lib/archethic_web/controllers/api/transaction_controller.ex
```diff
@@ -143,4 +143,69 @@ defmodule ArchethicWeb.API.TransactionController do
...
+    end
+  end
+
+  defp fetch_recipient_tx_and_simulate(recipient, tx) do
+    case Archethic.search_transaction(recipient) do
+      {:ok, prev_tx} ->
+        prev_tx =
+          prev_tx
+          |> TransactionPayload.to_map()
+          |> Transaction.cast()
```
bchamagne, line 203: Isn't `prev_tx` already a `%Transaction{}`

lib/archethic/contracts.ex
```diff
@@ -165,6 +168,43 @@ defmodule Archethic.Contracts do
...
+       }} ->
+        constants = %{
+          "previous" => Constants.from_transaction(prev_tx),
+          "next" => Constants.from_transaction(next_tx)
+        }
+
+        with :ok <- validate_conditions(transaction_conditions, constants),
+             :ok <- validate_triggers(triggers, next_tx, date),
+             %Transaction{} <-
+               ActionInterpreter.execute(Map.fetch!(triggers, :transaction), constants) do
```
bchamagne, line 196: `ActionInterpreter.execute` is not necessary returning a transaction. It might return nil and still be a valid contract (for example, when there is nothing to do)

bchamagne (changes requested):

test/archethic_web/controllers/api/transaction_controller_test.exs
apoorv-2204: +1

lib/archethic/contracts.ex
netboz: Hi Bastien, looking at the returns from `ActionInterpreter.execute` ( https://github.com/archethic-foundation/archethic-node/blob/develop/lib/archethic/contracts/interpreter/action.ex#L364 )  it seems a transaction is always returned ?

netboz: Ah I see now I guess `next_transaction`, returned by `execute/2` might be `nil`. If this is asserted, I will update the specs for `execute/2`

netboz: If `Nil` can on;y be returned when executing (  `ActionInterpreter.execute/2` ) with empty SC code. Then I think this might be ok because empty contracts are already checked at higher level in the code.

My issue is that it produce : 

``` %{"valid" => false, "address" => "00009e059e8171643b959284fe542909f3b32198b8fc25b3e50447589b84341c1d67", "reason" => "Transaction doesn't hold any contract"} ```

And I think we should rather consider empty contracts as valid instead ?

Comment (netboz) [reply lib/archethic/contracts.ex:196]: Julien confirmed empty contracts should produce a valid result. I will change the code.

File: lib/archethic/contracts.ex [hunk 7]
```diff
@@ -165,6 +168,44 @@ defmodule Archethic.Contracts do
...
+          ActionInterpreter.execute(Map.fetch!(triggers, :transaction), constants)
+          %{valid: true, address: prev_tx.address}
+        else
+          {:error, reason} ->
+            %{valid: false, address: prev_tx.address, reason: reason}
+        end
+
+      {:error, reason} ->
+        %{valid: false, address: prev_tx.address, reason: reason}
+    end
```
bchamagne, line 206: I think this low level function should return a `{:ok, _} | {:error, _}` or even a `:valid | {:invalid, reason}`. This `%{valid: bool(), ...}` struct should only live in the controller. What do you think?

bchamagne (changes requested):
Here's a few comments (some very biased) based only on reading the code. I'm going to need your help for a real word test.

lib/archethic.ex
```diff
@@ -290,6 +293,32 @@ defmodule Archethic do
...
+      when code != "" do
+    Contracts.simulate_contract_execution!(prev_tx, incoming_tx, DateTime.utc_now())
+  end
+
+  def simulate_contract_execution!(
+        prev_tx,
+        _next_tx
+      ) do
+    %{
+      valid: true,
```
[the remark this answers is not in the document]
netboz: I thought empty contracts where supposed to be valid. But indeed I tried to simulate an empty contract and get a successful answer but didn't managed.

lib/archethic_web/controllers/api/transaction_controller.ex
```diff
@@ -143,4 +143,80 @@ defmodule ArchethicWeb.API.TransactionController do
...
+        case Archethic.simulate_contract_execution!(prev_tx, tx) do
+          :ok ->
+            %{:valid => true, :address => Base.encode64(tx.address)}
+
+          {:error, reason} ->
+            %{"valid" => false, :address => Base.encode64(tx.address), reason: reason}
+        end
+
+      {:error, reason} ->
+        %{"valid" => false, :address => Base.encode64(tx.address), reason: reason}
```
bchamagne, line 219: We usually use hexadecimal to represent `address` in a json (encode16).
Use only strings for the keys in the `=>` map notation

I did not manage to test it via the endpoint. All I get is `invalid_inherit_constraints`.
I am using this contract: 
```
condition inherit: [
	content: true
]

condition transaction: [
	uco_transfers: size() > 0
]

	set_content "hello"
end
```

bchamagne: I think I found the issue. It's because you send the "incoming transaction" in `simulate_contract_execution`. But `simulate_contract_execution` wants the "next transaction" (you get that one after executing the contract (or nil)).

bchamagne (approved):
Works great! 

Tested with the transaction builder & curl. 
I tested 3 scenarios:

valid contract: 
`[{"address":"00002223BBD4EC3D64AE597696C7D7ADE1CEE65C639D885450AD2D7B75592AC76AFA","valid":true}]`

invalid contract:
`[{"address":"00002223BBD4EC3D64AE597696C7D7ADE1CEE65C639D885450AD2D7B75592AC76AFA","reason":"invalid_inherit_conditions","valid":false}]`

exception throwing contract: 
`[{"reason":"A contract exited while simulating","valid":false}]`

I found an issue with the transaction condition, but this is not the goal of this PR, I'll create an issue for that.

_ps: maybe add the address to the exit  message for consistency?_

bchamagne:
We are not going to merge that one yet, we will integrate it as soon as interpreter v1 is merged. We want to check the conditions in the Interpreter.exectute functions directly. Then we can use Interpreter.execute directly in the validate_contract function and in the worker as well.

lib/archethic/contracts.ex
```diff
@@ -165,6 +167,137 @@ defmodule Archethic.Contracts do
...
+  defp do_simulate_contract(
+         version,
+         trigger_code,
+         trigger_type,
+         conditions,
+         prev_tx,
+         incoming_tx,
+         date
+       ) do
+    case valid_from_trigger?(trigger_type, incoming_tx, date) do
```
samuelmanzanera, line 215: I think a `if` would do the job instead of a case ;)

samuelmanzanera: you could even use a `with`

bchamagne:
I'm going to do one last thing on this PR: 

- move the `Contracts.simulate_contract_execution` into `Interpreter.execute`. And then I'm going to use from both the simulate endpoint and from the worker (the only difference is the worker will actually persist the next transaction).

lib/archethic/contracts/interpreter.ex
```diff
@@ -136,6 +197,179 @@ defmodule Archethic.Contracts.Interpreter do
...
+      version,
+      trigger_code,
+      contract,
+      nil,
+      conditions.inherit,
+      opts
+    )
+  end
+
+  defp execute_trigger_and_validate_inherit_condition(
```
samuelmanzanera, line 273: Maybe this function can be split into two: one for trigger execution and one for the inherit acceptance.

lib/archethic/contracts/worker.ex
```diff
@@ -482,4 +294,48 @@ defmodule Archethic.Contracts.Worker do
         {:error, :not_enough_funds}
     end
   end
+
+  # no one use the return value
+  defp do_execute_trigger_and_handle_result(
```
samuelmanzanera, line 299: Maybe this one also can be split into two: trigger execution and validation of the flow like this one.

lib/archethic_web/controllers/api/transaction_controller.ex
```diff
@@ -143,4 +143,139 @@ defmodule ArchethicWeb.API.TransactionController do
...
+      ) do
+    case TransactionPayload.changeset(params) do
+      changeset = %{valid?: true} ->
+        tx =
+          %Transaction{data: %TransactionData{recipients: recipients}} =
+          changeset
+          |> TransactionPayload.to_map()
+          |> Transaction.cast()
+
+        {:ok, sup} = Task.Supervisor.start_link(strategy: :one_for_one)
```
samuelmanzanera, line 163: Why do you start a task supervisor for each request?

lib/archethic.ex
```diff
@@ -302,6 +308,29 @@ defmodule Archethic do
...
+          Transaction.t(),
+          nil | Transaction.t()
+        ) ::
+          {:ok, nil | Transaction.t()}
+          | {:error,
+             :invalid_triggers_execution
+             | :invalid_transaction_constraints
+             | :invalid_oracle_constraints
+             | :invalid_inherit_constraints}
+  def parse_and_execute_contract_at(
```
samuelmanzanera, line 326: Maybe we should expect the contract's instance as a parameter instead of a contract's transaction. Hence the parsing validation can be done upfront in the controller layer without blowing the process.

samuelmanzanera (approved):
Code is nicer ;)

lib/archethic_web/controllers/api/transaction_controller.ex
```diff
@@ -143,4 +143,144 @@ defmodule ArchethicWeb.API.TransactionController do
...
+          Map.merge(error_details, %{"valid" => false, "reason" => "Query validation failled"})
+
+        conn
+        |> put_status(:ok)
+        |> json([json_body])
+    end
+  end
+
+  defp fetch_recipient_tx_and_simulate(recipient_address, tx) do
+    with {:ok, contract_tx} <- Archethic.search_transaction(recipient_address),
```
Neylix, line 220: Here you should fetch the last transaction of the recipient chain before searching the transaction. In some case the recipient address can be the genesis address of the contract chain, and so the API return there is no tranaction for the recipient address

PR #808, opened 2023-01-05
Add getGenesisAddress query to graphql

aime-risson:
# Description

This Pr aims to add a `getGenesisAddress` query to graphql to make it easier to find the genesis address of a transaction.

## Type of change

# How Has This Been Tested?

Start a node, open the explorer, find a transaction and query the graphql endpoint to retrieve the genesis address.

# Checklist:

### DIFF
--- a/lib/archethic_web/graphql_schema.ex
+++ b/lib/archethic_web/graphql_schema.ex
@@ -41,6 +41,17 @@ defmodule ArchethicWeb.GraphQLSchema do
     end
 
     @desc """
+    Query the network to find the genesis address of a transaction
+    """
+    field :get_genesis_address, :genesis_address do
+      arg(:address, non_null(:address))
+
+      resolve(fn %{address: address}, _ ->
+        Resolver.get_genesis_address(address)
+      end)
+    end
+
+    @desc """
     Query the network to find the last transaction from an address
     """
     field :last_transaction, :transaction do
--- a/lib/archethic_web/graphql_schema.ex
+++ b/lib/archethic_web/graphql_schema.ex
@@ -43,7 +43,7 @@ defmodule ArchethicWeb.GraphQLSchema do
     @desc """
     Query the network to find the genesis address of a transaction
     """
-    field :get_genesis_address, :genesis_address do
+    field :genesis_address, :genesis_address do
       arg(:address, non_null(:address))
 
       resolve(fn %{address: address}, _ ->
--- a/lib/archethic_web/graphql_schema.ex
+++ b/lib/archethic_web/graphql_schema.ex
@@ -43,7 +43,7 @@ defmodule ArchethicWeb.GraphQLSchema do
     @desc """
     Query the network to find the genesis address of a transaction
     """
-    field :genesis_address, :genesis_address do
+    field :genesis_address, :address do
       arg(:address, non_null(:address))
 
       resolve(fn %{address: address}, _ ->
--- a/lib/archethic_web/graphql_schema/resolver.ex
+++ b/lib/archethic_web/graphql_schema/resolver.ex
@@ -26,7 +26,6 @@ defmodule ArchethicWeb.GraphQLSchema.Resolver do
       _ ->
         {:ok, %{genesis: address}}
     end
-
   end
 
   def get_balance(address) do
--- a/lib/archethic_web/graphql_schema/resolver.ex
+++ b/lib/archethic_web/graphql_schema/resolver.ex
@@ -18,6 +18,30 @@ defmodule ArchethicWeb.GraphQLSchema.Resolver do
 
   @limit_page 10
 
+  def get_genesis_address(address) do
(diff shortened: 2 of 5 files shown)


lib/archethic_web/graphql_schema/resolver.ex
```diff
@@ -18,6 +18,30 @@ defmodule ArchethicWeb.GraphQLSchema.Resolver do
 
   @limit_page 10
 
+  def get_genesis_address(address) do
+    t1 = Task.async(fn -> TransactionChain.fetch_genesis_address_remotely(address) end)
```
tenmoves, line 22: Hello, ArchethicWeb should be agnostic to the underlying architecture of the archethic project ( it should only interact with the public api in the Archethic module).
I would suggest to change this call to `Archethic.fetch_genesis_address_remotely`

lib/archethic_web/graphql_schema.ex
```diff
@@ -40,6 +40,17 @@ defmodule ArchethicWeb.GraphQLSchema do
...
+    @desc """
+    Query the network to find the genesis address of a transaction
+    """
+    field :get_genesis_address, :genesis_address do
+      arg(:address, non_null(:address))
+
+      resolve(fn %{address: address}, _ ->
+        Resolver.get_genesis_address(address)
+      end)
+    end
```
tenmoves, line 52: this part need to be tested, you can have a look at test/archethic_web/graphql_schema_test.exs

lib/archethic_web/graphql_schema/resolver.ex
tenmoves: you probably don't need a Task here, I would call directly the function in this case.

aime-risson: > Hello, ArchethicWeb should be agnostic to the underlying architecture of the archethic project ( it should only interact with the public API in the Archethic module).

Ok thanks, I'll fix that! Could you explain to me why it should be agnostic to the underlying architecture of the archethic project, I don't really understand?

tenmoves: ArchethicWeb represents the frontend and Archethic represents the backend logic, in here they are in the same repository but this can change someday in the future, also we try to follow the hexagonal architecture (some call it clean architecture).

Finally if the web only depends on the public api, it means that if we refactor the whole backend tomorrow we will not need to change the web part and separate people / teams can work independently on different parts of the project. 

Feel free to ping me if it isn't clear.

samuelmanzanera:
Also, I think you could rename it as `genesisAddress` , because the entire graphQL API is readonly. Hence, the get could be omitted I guess.

lib/archethic_web/graphql_schema/resolver.ex
aime-risson: Very clear, thanks for the explanation!

I tried using the call `Archethic.fetch_genesis_address_remotely` as you suggested however it seems to always return the same address as entered.

here is the code:
```
def get_genesis_address(address) do
    case Archethic.fetch_genesis_address_remotely(address) do
      {:ok, genesis_address} ->
        {:ok, %{genesis: genesis_address}}

      _ ->
        {:ok, %{genesis: address}}
    end
end
```
Correcting the code by using TransactionChain.fetch_genesis_address_remotely(address) fixes the issue.
Do you have any idea why or what I might be doing wrong?

tenmoves: you can look at the definition here: https://github.com/archethic-foundation/archethic-node/blob/develop/lib/archethic.ex#L246

aime-risson (commented):
This is a fix to Archethic.fetch_genesis_address_remotly which does not return a correct map when  `genesis_address` is matched

lib/archethic_web/graphql_schema.ex
```diff
@@ -40,6 +40,17 @@ defmodule ArchethicWeb.GraphQLSchema do
       end)
     end
 
+    @desc """
+    Query the network to find the genesis address of a transaction
+    """
+    field :genesis_address, :genesis_address do
```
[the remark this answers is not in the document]
aime-risson: probably not needed indeed, will update it.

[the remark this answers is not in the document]
aime-risson: > Why create another type for the genesis address ?

Actually, I think it is needed. This query only returns the genesis address so I have to create a new type to only return the map `genesis: address`. Or am I getting this wrong ?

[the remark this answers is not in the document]
aime-risson: Thanks to @Neylix I was able to understand what was meant to do. The genesisAddress type was removed.

samuelmanzanera (approved):

bchamagne (approved):
neat!

tenmoves (approved):

Issue #700, opened 2022-11-17
Forwarding random port with upnpc


Neylix:
### Describe the problem you discovered

In previous implementation, while forwarding to port 0, the router was redirecting a random port. But since we use miniupnpc, we cannot anymore forward the port 0, so we cannot open a random port automatically.

### Describe the solution you'd like

We have 2 options : 
- Consider using an other tool for port forwarding like https://github.com/benoitc/erlang-nat if this implementation is working well it would be better than having a upnpc programm compiled in C and opening a random port using 0 works in this lib
- Determining a list of port possibility and taking a random number until the port opening is working

apoorv-2204:
> ## Update 

'https://github.com/benoitc/erlang-nat'
With the upnp capability enabled or disabled, my local network is inoperable with this nat lib operations, and I am unable to connect to other nodes.
### Proposed solution:
1) Continue with option nb 2,
2) Build a cli to ask you for input for env variables. else they can move with default

![image](https://user-images.githubusercontent.com/90304197/210516649-a4c46078-b855-4b1a-9848-4812f987aee9.png)

PR #810, opened 2023-01-05
Forwarding random port with upnpc #700

apoorv-2204:
# Description

We chose a random port , from a list of unassigned  ports as per IANA.

Fixes 

- [ ] #700 

## Type of change

# How Has This Been Tested?

Manual
image

# Checklist:


lib/archethic/networking/port_forwarding.ex
```diff
@@ -70,4 +75,27 @@ defmodule Archethic.Networking.PortForwarding do
   defp ip_lookup_provider do
     Application.get_env(:archethic, IPLookup)
   end
+
+  @iana_registry_path Path.join([
```
samuelmanzanera, line 79: You should put non mutable (by code) resources in the priv folder .

lib/archethic/networking/port_forwarding.ex
```diff
@@ -70,4 +75,27 @@ defmodule Archethic.Networking.PortForwarding do
...
+                        File.cwd() |> elem(1),
+                        "lib",
+                        "archethic",
+                        "networking",
+                        "registry",
+                        "iana_unassigned_ports"
+                      ])
+
+  # see https://github.com/apoorv-2204/iana_port_registry
+  defp list_iana_unassigned_ports() do
```
samuelmanzanera, line 89: I think you could simplify without the need to list the port from IANA. Between 49152 through 65535 should make it. 
The randomness could be recursive until one port is free.

samuelmanzanera (changes requested):

samuelmanzanera (commented):
Code seems ok for me

samuelmanzanera (approved):

bchamagne:
I see some functions that use this code are completely ignoring the random port:

ex:  Archethic.Networking.Scheduler

```
  defp open_ports do
    p2p_port = Application.get_env(:archethic, P2PListener) |> Keyword.fetch!(:port)
    web_port = Application.get_env(:archethic, WebEndpoint) |> get_in([:http, :port])

    with {:ok, _} <- PortForwarding.try_open_port(p2p_port, false),
         {:ok, _} <- PortForwarding.try_open_port(web_port, false) do
      {:ok, p2p_port, web_port}
    end
  end
```

bchamagne:
what about HTTP(S) port, you don't want to open random port for them (ex: ArchethicWeb.Supervisor.init)

bchamagne:
> what about HTTP(S) port, you don't want to open random port for them (ex: ArchethicWeb.Supervisor.init)

Not an issue, I didn't notice the `force = false` flag.

samuelmanzanera (approved):

samuelmanzanera:
@apoorv-2204 Could you check why the CI is not passing ?

apoorv-2204:
> @apoorv-2204 Could you check why the CI is not passing ?

Thanks for pointing it out

Neylix (commented):
On line 21 of PortForwarding module, if the function `do_try_open_port` return an error, the fallback is never called so the random port is never published

apoorv-2204:
> On line 21 of PortForwarding module, if the function `do_try_open_port` return an error, the fallback is never called so the random port is never published

```  {:ok, port} <- do_try_open_port(port) do
      {:ok, port}```

Neylix (approved):

PR #814, opened 2023-01-06
refactor message module

### LINKED ISSUE #799: Refactor Message module

Actually all serialization and deserialization are done in the message module, but as the number of message growth, the reaadability of this module and it's testing begin to be complex.

We should refactor this module to improve test and readability.
The serialization and deserialization could be done in the module of the message itself, an actual example is the `ShardRepair` message which use
```elixir
def encode(msg = %ShardRepair{}) do
  <<230::8, ShardRepair.serialize(msg)::bitstring>>
end
```
The testing part of serialization and deserialization can be done in doctest of this message

We can also put in the message itself the process, for example: 
```elixir
def process(msg = %ShardRepair{}) do
  ShardRepair.process(msg)
end
```
To go further we don't need to patern match de message type, but directly call the structure, so there would be only 1 function process in the Message module:
```elixir
def process(msg) do
  msg.__struct__.process(msg)
end
```

To encode and decode a message we need to know the id number of it. We could create a correspondence table like 
```elixir
id_map = %{
  ShardRepair => <<230>>,
  ... => ...
}
```
So we could have 1 function for encoding and decoding :
```elixir
def encode(msg) do
  msg_id = Map.get(id_map, msg.__struct__)
  serialized = msg.__struct__.serialize(msg)
  <<msg_id::8, serialized::binary>>
end

def decode(<<msg_id::8, rest::binary>>) do
  {msg_type, _} = Enum.find(id_map, fn {key, value} -> value == msg_id end)
  msg_type.deserialize(rest)
end
```

### Additional context

_No response_

tenmoves:
# Description

Fixes #799 

## Type of change

-  Refactoring

# Checklist:


bchamagne (commented):
Suggestion:

The encode function is not doing much, it is only doing work when there is no serialize/1 function. 

Is it possible to keep the encode/decode functions in the message module as one liners per clause? This way we could keep all the prefix in the same spot?

```
@spec encode(t()) :: bitstring()
def encode(msg = %GetBoostrappingNodes{}), do: <<0::8, msg.__struct__.serialize(msg)
def encode(msg = %GetStorageNonce{}), do: <<1::8, msg.__struct__.serialize(msg)
...

@spec decode(bitstring) :: t()
def decode(<<0::8, rest::bitstring), do: GetBootstrappingNodes.deserialize(rest)
def decode(<<1::8, rest::bitstring), do: GetStorageNonce.deserialize(rest)
...
```

lib/archethic/transaction_chain/transaction.ex
```diff
@@ -900,4 +900,9 @@ defmodule Archethic.TransactionChain.Transaction do
         |> Enum.map(&CrossValidationStamp.cast/1)
     }
   end
+
+  @spec encode(t()) :: bitstring()
```
bchamagne, line 904: can be removed?

lib/archethic/p2p/message.ex
```diff
@@ -237,1731 +208,26 @@ defmodule Archethic.P2P.Message do
...
-  end
-
-  def decode(<<254::8, rest::bitstring>>) do
-    {%Ok{}, rest}
+  def decode(<<message_id::8, rest::bitstring>>) do
+    msg_module = MessageId.id_to_module(message_id)
+    msg_module.deserialize(rest)
   end
 
   def decode(<<255::8>>), do: raise("255 message type is reserved for stream EOF")
```
bchamagne, line 226: This clause should be before the precedent:

```
iex(1)> Archethic.P2P.Message.decode(<<255::8>>)
** (FunctionClauseError) no function clause matching in Archethic.P2P.MessageId.id_to_module/1   
```

bchamagne (approved):
minor comments, looks great!

lib/archethic/p2p/message.ex
tenmoves: nice catch

Neylix:
For the conflict, there is actually 2 changes:
- GetFirstAddress and FirstAddress message has been renamed (including file name)
- 3 new messages is added

tenmoves (commented):
excellent the metaprogramming part

Neylix (changes requested):

Neylix (approved):

PR #816, opened 2023-01-09
Refactor DB plus some optimization on explorer

Neylix:
# Description

This PR provide some refactor in the DB, to remove and enhance functions.
Also permit to retrieve a chain and a chain size from the DB based on a genesis address.
Also remove the limit address while retrieving a chain from DB, now it always return 10 transactions or transaction until the end of the file

## Type of change

- Improvement 

# How Has This Been Tested?

Updated some unit tests

# Checklist:

### DIFF
--- a/lib/archethic.ex
+++ b/lib/archethic.ex
@@ -245,45 +245,55 @@ defmodule Archethic do
   """
   @spec get_transaction_chain_by_paging_address(binary(), binary() | nil, :asc | :desc) ::
           {:ok, list(Transaction.t())} | {:error, :network_issue}
-  def get_transaction_chain_by_paging_address(address, paging_address, order)
-      when is_binary(address) and order in [:asc, :desc] do
-    nodes = Election.chain_storage_nodes(address, P2P.authorized_and_available_nodes())
-
-    try do
-      transaction_exists? =
-        if paging_address == nil,
-          do: true,
-          else: TransactionChain.transaction_exists?(paging_address)
-
-      {local_chain, paging_address} =
-        with true <- transaction_exists?,
-             last_address when last_address != nil <-
-               TransactionChain.get_last_local_address(address),
-             true <- last_address != paging_address do
-          {transactions, _, _} =
-            TransactionChain.get(last_address, [], paging_state: paging_address, order: order)
-
-          {transactions, last_address}
+  def get_transaction_chain_by_paging_address(address, paging_address, :asc)
+      when is_binary(address) do
+    case get_last_transaction_address(address) do
+      {:ok, last_address} ->
+        with {local_chain, false, _} <-
+               TransactionChain.get(address, [], paging_state: paging_address, order: :asc),
+             %Transaction{address: ^last_address} <- List.last(local_chain) do
+          {:ok, local_chain}
         else
-          _ -> {[], paging_address}
+          {local_chain, true, _} ->
+            # Local chain already contains 10 transactions
+            {:ok, local_chain}
+
+          _ ->
+            last_address
+            |> Election.chain_storage_nodes(P2P.authorized_and_available_nodes())
+            |> TransactionChain.fetch_transaction_chain(last_address, paging_address, order: :asc)
         end
 
-      remote_chain =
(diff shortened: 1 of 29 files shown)


lib/archethic/db/embedded_impl/chain_reader.ex
```diff
@@ -128,33 +128,74 @@ defmodule Archethic.DB.EmbeddedImpl.ChainReader do
...
+
+              {transactions, next_transactions}
+
+            {transactions, false, _} ->
+              {transactions, :eof}
+
+            :eof ->
+              {:halt, nil}
+          end,
+          fn _ -> :file.close(fd) end
```
tenmoves, line 194: you used File.open from elixir and here :file from erlang, maybe it's better to use the same (probably FIle is just a wrapper around :file but it's better to use the same for consistency).

lib/archethic/db/embedded_impl/chain_reader.ex
```diff
@@ -266,43 +307,64 @@ defmodule Archethic.DB.EmbeddedImpl.ChainReader do
...
-
           if paging_state_idx < @page_size do
-            {limit_address, nil, false, ""}
+            {paging_state_idx, nil, false, nil}
           else
             idx = paging_state_idx - 1 - @page_size
 
-            {limit_address, all_addresses_asc |> Enum.at(idx), true,
+            {@page_size, all_addresses_asc |> Enum.at(idx), true,
              all_addresses_asc |> Enum.at(idx + 1)}
```
tenmoves, line 335: for clarity, I would suggest to do the transformations before returning the tuple

lib/archethic.ex
```diff
@@ -239,93 +239,53 @@ defmodule Archethic do
...
-             true <- TransactionChain.transaction_exists?(paging_address),
-             last_address when last_address != nil <-
-               TransactionChain.get_last_local_address(address),
-             true <- last_address != paging_address do
-          {TransactionChain.get_locally(last_address, paging_address, order), last_address}
+  def get_transaction_chain_by_paging_address(address, paging_address, :asc)
+      when is_binary(address) do
+    case get_last_transaction_address(address) do
+      {:ok, last_address} ->
+        with {local_chain, false, _} <-
```
bchamagne, line 252: Am I the only one to have trouble with this kind of `with`? It takes me 2 minutes to read it everytime.

bchamagne (approved):
well done!

lib/archethic.ex
Neylix: It is like a case, but with multiple values. If we rewrite this with case statement it would be like this
```elixir
case TransactionChain.get(address, [], paging_state: paging_address, order: :asc) do
  {local_chain, true, _} ->
    {:ok, local_chain}

  {local_chain, _, _} ->
    case List.last(local_chain) do
      %Transaction{address: ^last_address} -> 
        {:ok, local_chain}

      _ ->
        last_address
            |> Election.chain_storage_nodes(P2P.authorized_and_available_nodes())
            |> TransactionChain.fetch_transaction_chain(last_address, paging_address, order: :asc)
    end
end
```

lib/archethic/db/embedded_impl.ex
```diff
@@ -282,9 +282,9 @@ defmodule Archethic.DB.EmbeddedImpl do
   @doc """
   Return the first address of given chain's address
   """
-  @spec get_first_chain_address(address :: binary()) :: binary()
-  def get_first_chain_address(address) when is_binary(address) do
```
samuelmanzanera, line 286: If we remove this one, the PR about `get_first_address` might not be possible

Neylix: It's just renaming the function as it is actually getting the genesis address and not the first address

samuelmanzanera: Ok

samuelmanzanera: We might rename the message itself then , as it might create confusion

Neylix: Done !

samuelmanzanera (approved):

PR #821, opened 2023-01-10
Fix replication chain ingestion

Neylix:
# Description

There is some case where a node should replicate a transaction but it do not due to some latency or network issue. If there is a new transaction in this chain and the node is again storage node of this transaction, it will store the previous transaction without ingesting data as it should have been done.
This PR fix this problem by ingesting if needed the previous transaction of a chain

## Type of change

# Checklist:

### DIFF
--- a/lib/archethic/replication.ex
+++ b/lib/archethic/replication.ex
@@ -39,6 +39,7 @@ defmodule Archethic.Replication do
   alias Archethic.TransactionChain
   alias Archethic.TransactionChain.Transaction
   alias Archethic.TransactionChain.Transaction.ValidationStamp
+  alias Archethic.TransactionChain.Transaction.ValidationStamp.LedgerOperations
 
   alias Archethic.Utils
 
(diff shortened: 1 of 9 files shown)


lib/archethic/replication.ex
```diff
@@ -106,7 +106,19 @@ defmodule Archethic.Replication do
...
-          |> Stream.each(&TransactionChain.write/1)
+          |> Stream.flat_map(& &1)
+          |> Stream.each(fn tx = %Transaction{address: address} ->
+            TransactionChain.write_transaction(tx)
+
+            # There is some case where a transaction is not replicated while it should
+            # because of some latency or network issue. So when we replicate a past chain
+            # we also ingest the transaction if we are storage node of it
+
+            if address
```
samuelmanzanera: I think you can extract into a variable to make it easier to read
Long instructions in a condition statement are harder to understand.

lib/archethic/replication.ex
```diff
@@ -107,18 +108,39 @@ defmodule Archethic.Replication do
...
+               } ->
+              TransactionChain.write_transaction(tx)
+
+              # There is some case where a transaction is not replicated while it should
+              # because of some latency or network issue. So when we replicate a past chain
+              # we also ingest the transaction if we are storage node of it
+
+              io_addresses = Enum.map(transaction_movements, & &1.to)
+
+              cond do
```
samuelmanzanera, line 128: Same remark in this commit ;)

samuelmanzanera (approved):

PR #822, opened 2023-01-12
SC: add get_calls/0  action

### LINKED ISSUE #512: Add the possibility to get the inputs in the smart contract

In order to enhance the scalabily of the smart contract, we can use some batching process.
This can work like this: create a smart contract triggered by interval, at the trigger time the smart contract will take all the inputs it got since the last trigger time and doing some actions.

So to do this batching process we have to provide the way to get the inputs and all data associated (content, uco transfer ...) in the smart contract.

We can imagine a fonction like `get_inputs` that return a list with all the inputs of the smart contract. These inputs would have the same attribute as a transaction.

### Additional context

_No response_

bchamagne:
# Description

This PR aims at adding a `get_calls/1` function to the smart contracts' actions. 
This function can be used to create aggregator contracts that do not do anything until some threshold.
I found few issues along the way that I will create and fix outside of this PR. 

- [x] Create a get_calls/0 that will automatically add the contract.address 
- [x] Update smart contract documentation

Should the get_calls/1 be available outside the action block?

Fixes #512

## Type of change

# How Has This Been Tested?

I have tested with this contract: 

```
condition inherit: [
	type: true,
	content: true,
	uco_transfers: true
]

condition transaction: [
	uco_transfers: size() > 0
]

	transactions = get_calls(contract.address)
	set_content size(transactions)
	set_type transfer
	add_uco_transfer to: "000030831178cd6a49fe446778455a7a980729a293bfa16b0a1d2743935db210da76", amount: 1337
end
```

When I trigger this contract, a new transaction in the contract chain have the content updated to "1". 

# Checklist:

### DIFF
--- a/lib/archethic/contracts/interpreter/library.ex
+++ b/lib/archethic/contracts/interpreter/library.ex
@@ -191,16 +191,14 @@ defmodule Archethic.Contracts.Interpreter.Library do
   """
   @spec get_calls(binary()) :: list(map())
   def get_calls(contract_address) do
-    contract_address = Utils.maybe_decode_hex(contract_address)
-
-    # TODO: parallelize?
-    TransactionLookup.list_contract_transactions(contract_address)
-    |> Enum.map(fn {address, _, _} -> address end)
-    |> Enum.map(fn address ->
+    contract_address
+    |> Utils.maybe_decode_hex()
+    |> TransactionLookup.list_contract_transactions()
+    |> Enum.map(fn {address, _, _} ->
+      # TODO: parallelize
       {:ok, tx} = TransactionChain.get_transaction(address, [], :io)
-      tx
+      ContractConstants.from_transaction(tx)
     end)
-    |> Enum.map(&ContractConstants.from_transaction(&1))
   end
 
   @doc """
--- a/lib/archethic/contracts/interpreter/library.ex
+++ b/lib/archethic/contracts/interpreter/library.ex
@@ -8,7 +8,8 @@ defmodule Archethic.Contracts.Interpreter.Library do
     P2P.Message.FirstPublicKey,
     TransactionChain,
     Contracts.ContractConstants,
-    Contracts.TransactionLookup
+    Contracts.TransactionLookup,
+    Contracts.Interpreter.Utils
   }
 
   @doc """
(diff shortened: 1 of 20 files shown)


lib/archethic/contracts/interpreter/library.ex
```diff
@@ -151,29 +156,47 @@ defmodule Archethic.Contracts.Interpreter.Library do
...
+
+  This is useful for contracts that want to throttle their calls
+  """
+  @spec get_calls(binary()) :: list(map())
+  def get_calls(contract_address) do
+    contract_address = Utils.maybe_decode_hex(contract_address)
+
+    # TODO: parallelize?
+    TransactionLookup.list_contract_transactions(contract_address)
+    |> Enum.map(fn {address, _, _} -> address end)
```
samuelmanzanera, line 186: I think you can reduce the iterations, using a single Enum.map

lib/archethic/contracts/interpreter/library.ex
```diff
@@ -151,29 +156,47 @@ defmodule Archethic.Contracts.Interpreter.Library do
...
+  Get the inputs(type= :call) of the given transaction
+
+  This is useful for contracts that want to throttle their calls
+  """
+  @spec get_calls(binary()) :: list(map())
+  def get_calls(contract_address) do
+    contract_address = Utils.maybe_decode_hex(contract_address)
+
+    # TODO: parallelize?
+    TransactionLookup.list_contract_transactions(contract_address)
```
samuelmanzanera, line 185: Also you can optimize the reading by piping the contract_address

samuelmanzanera:
> Should the get_calls/1 be available outside the action block?

I think for now, we can keep it in the actions block

bchamagne:
As discussed with @Neylix , we should remove the get_calls/1 for 2 reasons: 

- 1: no use case yet
- 2: actual implementation look at calls in local node, which is fine for contract's address, but not for other contract's address. We would have to download the transactions from network. Downloading hundreds of them might not be efficient.

What do you think @samuelmanzanera ?

lib/archethic/db/embedded_impl/chain_reader.ex
```diff
@@ -201,6 +201,21 @@ defmodule Archethic.DB.EmbeddedImpl.ChainReader do
     end
   end
 
+  @doc """
+  Read a transaction from io storage
+  """
+  @spec get_io_transaction(binary(), fields :: list(), db_path :: String.t()) ::
+          Transaction.t() | nil
```
Neylix, line 208: As the function `read_transaction` this function should return a tuple `{:ok, tx}` or `{:error, reason}`

Neylix (changes requested):

Neylix (approved):

Neylix (approved):

PR #830, opened 2023-01-13
Feat/add transaction error subscription to playbook

cluster:
# Description

The goal of this PR is to add a graphql subscription to `transactionError` in the playbook, so we know if the transaction has failed or succeeded.

## Type of change

# How Has This Been Tested?

Tested locally


mix.exs
```diff
@@ -107,7 +107,7 @@ defmodule Archethic.MixProject do
       {:ex_json_schema, "~> 0.9", override: true},
       {:pathex, "~> 2.4"},
       {:easy_ssl, "~> 1.3"},
-      {:castore, "~> 0.1"},
+      {:castore, "~> 1.0", override: true},
```
tenmoves, line 110: why is the override: true needed ?

cluster: It's because otherwise we get the following error:
```
mix deps.get
Resolving Hex dependencies...

Failed to use "castore" because
  elixir_make (version 0.7.3) requires ~> 0.1
  mint (version 1.4.2) requires ~> 0.1.0
  mix.lock specifies 1.0.0

** (Mix) Hex dependency resolution failed, change the version requirements of your dependencies or unlock them (by using mix deps.update or mix deps.unlock). If you are unable to resolve the conflicts you can try overriding with {:dependency, "~> 1.0", override: true}
```
🤔

Neylix (approved):

PR #853, opened 2023-01-19
Reward month occurence should be dynamic

### LINKED ISSUE #827: Reward month occurence is static

To calculate the reward for a node reward transaction, we need to know how many node reward transaction there will be in a month. To calculate this number, we have a function that provide the number of occurence for a month based on the reward scheduler interval.
The problem is that this number is calculated on compile time, but it should evolve depending of the current month.
It may cause error in reward amount, and if 2 nodes didn't compile the code on the same month, they will have different value for the reward amount causing node reward transaction to fail.

Each month, the number should be updated with the current month

tenmoves:
# Description

Fixes #827 

## Type of change

# How Has This Been Tested?

- doctests
- manual

# Checklist:

### DIFF
--- a/lib/archethic/reward.ex
+++ b/lib/archethic/reward.ex
@@ -34,7 +34,7 @@ defmodule Archethic.Reward do
   require Logger
 
   @unit_uco 100_000_000
-  @number_of_reward_occurences_per_month Utils.number_of_reward_occurences_per_month()
+  @number_of_occurences_per_month_for_a_year Utils.number_of_possible_reward_occurences_per_month_for_a_year()
 
   @doc """
   Get rewards amount for validation nodes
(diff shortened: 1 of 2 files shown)


lib/archethic/reward.ex
```diff
@@ -53,7 +53,17 @@ defmodule Archethic.Reward do
     trunc(50 / uco_usd_price / number_of_reward_occurences_per_month * @unit_uco)
   end
 
-  defp number_of_reward_occurences_per_month(), do: @number_of_reward_occurences_per_month
+  defp number_of_reward_occurences_per_month() do
```
bchamagne, line 56: If I understand correctly, the function `Utils.number_of_reward_occurences_per_month()` is working well and the issue is that it was called only at compile time.

Isn't it possible to just move it to runtime like this ?

```
defp number_of_reward_occurences_per_month() do 
    datetime = NaiveDateTime.utc_now()
    interval = Application.get_env(:archethic, RewardScheduler)[:interval]
    Utils.number_of_reward_occurences_per_month(interval, datetime)
end
```

tenmoves: no because in some cases depending on the interval  it can take more than 1s to execute which is not acceptable

Neylix (approved):

PR #854, opened 2023-01-20
Pr 824 integrate hydrating cache oracle chain

### LINKED ISSUE #824: Enhance Oracle services data fetching with self hydrating cache

- We might need a per-service refresh frequency
- Currently if a service is not responding the service time-out is added to oracle transaction processing time.
- Later, as more services will be added ( meteo, ... ), we will need a solution that can scale better
- This will bring also a TTL to the services values

This can be achieved by integrating a ["self hydrating"](https://gitlab.com/yan.guiborat1/cache) cache between the services and the scheduler.

The cache will be started at node startup.
The cache, at initialisation will retrieve the services list and the hydrating ( update ) function from node configuration.
If the scheduler request a value for which the hydrating function is running, then the cache will wait for function execution termination before returning the value.

### Additional context

_No response_

### LINKED ISSUE #836: Oracle tx crashes if the provider cannot fetch eur or usd price

In local I got this error : 

image

The uco price verification method should not crash if eur or usd price is missing

netboz:
# Description

This will address ticket #824 and 
Fixes #824 
Fixes #836 

## Type of change

There is still some refinement needed.

# How Has This Been Tested?

# Checklist:


config/dev.exs
```diff
@@ -75,6 +75,26 @@ config :archethic,
...
+     3000}
+  ],
+  coco_service: [
+    {:coin_paprika, Archethic.OracleChain.Services.UCOPrice.Providers.CoinPaprika, :fetch,
+     [["usd", "eur"]], 3000},
+    {:coin_gecko, Archethic.OracleChain.Services.UCOPrice.Providers.Coingecko, :fetch,
+     [["usd", "eur"]], 3000},
+    {:coin_market_cap, Archethic.OracleChain.Services.UCOPrice.Providers.CoinMarketCap, :fetch,
+     [["usd", "eur"]], 3000}
+  ]
```
tenmoves, line 96: coco_service ?

lib/archethic/oracle_chain/services/uco_price.ex
```diff
@@ -18,45 +20,42 @@ defmodule Archethic.OracleChain.Services.UCOPrice do
...
+          false
+      end)
+      |> Enum.map(fn
+        {_, result = %{}} ->
           result
       end)
+
       ## Here stream looks like : [%{"eur"=>[0.44], "usd"=[0.32]}, ..., %{"eur"=>[0.42, 0.43], "usd"=[0.35]}]
       |> Enum.reduce(%{}, &agregate_providers_data/2)
+      |> IO.inspect()
```
tenmoves, line 54: inspect

lib/archethic/utils/hydrating_cache/cache_entry.ex
```diff
@@ -0,0 +1,229 @@
...
+
+  defstruct running_func_task: :undefined,
+            hydrating_func: :undefined,
+            getters: [],
+            ttl: :undefined,
+            refresh_interval: :undefined,
+            key: :undefined,
+            value: :"$$undefined",
+            timer_func: :undefined,
+            timer_discard: :undefined
```
tenmoves, line 14: do we need to specifity undefined in the struct definition ?

lib/archethic/utils/hydrating_cache/cache_entry.ex
```diff
@@ -0,0 +1,229 @@
...
+       refresh_interval: refresh_interval
+     }}
+  end
+
+  @impl :gen_statem
+  def callback_mode, do: [:handle_event_function, :state_enter]
+
+  @impl :gen_statem
+  def handle_event({:call, from}, {:get, _from}, :idle, data) do
+    IO.puts("==> Get in idle state")
```
tenmoves, line 57: maybe we can use logger to be able to adapt the level of logs in dev for example

lib/archethic/utils/hydrating_cache/cache_entry.ex
```diff
@@ -0,0 +1,229 @@
...
+    send(from, {:ok, data.value})
+    {:next_state, :running, data}
+  end
+
+  def handle_event({:call, from}, {:register, fun, key, ttl, refresh_interval}, :running, data) do
+    ## Registering a new hydrating function while previous one is running
+
+    ## We stop the hydrating task if it is already running
+    case data.running_func_task do
+      pid when is_pid(pid) -> Process.exit(pid, :brutal_kill)
```
tenmoves, line 103: it's probably better to use another signal than brutal kill, to be able to trap the exit and throw some logs

config/dev.exs
netboz: I will remove it, this was for testing multiple services

netboz: thanks

lib/archethic/utils/hydrating_cache/cache_entry.ex
netboz: yes, there are still some traces I need to remove

lib/archethic/utils/hydrating_cache/hydrating_cache.ex
```diff
@@ -0,0 +1,222 @@
...
+      before it is discarded if the value is not refreshed.
+    - `refresh_interval`: how often (in milliseconds) the function is
+      recomputed and the new value stored. `refresh_interval` must be strictly
+      smaller than `ttl`. After the value is refreshed, the `ttl` counter is
+      restarted.
+
+  The value is stored only if `{:ok, value}` is returned by `fun`. If `{:error,
+  reason}` is returned, the value is not stored and `fun` must be retried on
+  the next run.
+  """
```
tenmoves, line 39: :+1:

netboz:
I need to do a final refinement, like to remove unused mocks and I would like to add metrics also, if possible.

lib/archethic/oracle_chain/services/uco_price.ex
```diff
@@ -21,30 +23,34 @@ defmodule Archethic.OracleChain.Services.UCOPrice do
...
-          end
-        end,
-        on_timeout: :kill_task
-      )
-      |> Stream.filter(&match?({:ok, {:ok, _}}, &1))
-      |> Stream.map(fn
-        {_, {_, result = %{}}} ->
+      Enum.map(providers(), fn provider ->
+        case HydratingCache.get(
+               :"Elixir.Archethic.Utils.HydratingCache.uco_service",
```
samuelmanzanera, line 28: Why not using direct atom ?

lib/archethic/utils/hydrating_cache/cache_entry.ex
```diff
@@ -0,0 +1,278 @@
+defmodule Archethic.Utils.HydratingCache.CacheEntry.StateData do
```
samuelmanzanera, line 1: I think you can use `defstruct` right inside CacheEntry.

lib/archethic/oracle_chain/services/uco_price.ex
netboz: :) Right, this is definitely something I need to refine

bchamagne (commented):
2 suggestions for me:

# folder
there is an "archethic_cache" folder at root, I would move the HydratingCache in here. 

# config
I think there are 2 configurations mixed in 1.
Here you have a UCOPrice config that contains both providers & cache configuration (refreshrate/ttl).
Then in the Supervisor, you transform this into {MFA}.
I suggest to have the {MFA} directly in a HydratingCache config (adding services should not require code changes in the cache). 

```
uco_providers =  [
  Archethic.OracleChain.Services.UCOPrice.Providers.CoinX,
  Archethic.OracleChain.Services.UCOPrice.Providers.CoinY,
  Archethic.OracleChain.Services.UCOPrice.Providers.CoinZ
  ]

config :archethic, Archethic.OracleChain.Services.UCOPrice,
  providers: uco_providers
  
config :archethic, Archethic.Utils.HydratingCache,
   [<some_other_cache>] ++ 
    Enum.map(uco_providers,&({&1, {&1, :fetch, [["eur", "usd"]]}, 60_000, :infinity}))
```

bchamagne:
I updated a few things as we discussed, I suggest to read commit by commit.

lib/archethic/oracle_chain/supervisor.ex
```diff
@@ -19,22 +18,28 @@ defmodule Archethic.OracleChain.Supervisor do
...
+    Application.get_env(:archethic, Archethic.OracleChain, [])
+    |> Keyword.get(:services, [])
+    |> Keyword.values()
+    |> Enum.map(fn service_module ->
+      # we expect a list of providers in config for each services
+      providers =
+        Application.get_env(:archethic, service_module, [])
+        |> Keyword.get(:providers, [])
+
+      {HydratingCache, [service_module, providers]}
```
samuelmanzanera, line 42: Does it mean one cache per service?
Why don't we define HypdratingCacheSupervisor, where each child would have a HydratingCache (service) which its children as CacheEntry (provider)

samuelmanzanera (changes requested):

bchamagne (approved):
Looks good to me, just a slight comment about  `Archethic.OracleChain.ServiceCacheSupervisor` which I believe is not helpful, but no harm here.

samuelmanzanera:
> Looks good to me, just a slight comment about `Archethic.OracleChain.ServiceCacheSupervisor` which I believe is not helpful, but no harm here.

Yes, the point is to enfoce implementation to leverage hydrating cache for the fetching.

lib/archethic/oracle_chain/services/hydrating_cache.ex
```diff
@@ -0,0 +1,120 @@
...
+        try do
+          {:ok, apply(m, f, a)}
+        rescue
+          _ ->
+            :error
+        end
+      end)
+
+    # start a new hydrate timer
+    Process.send_after(self(), :hydrate, refresh_interval)
```
Neylix, line 81: It would be better to start the new timer when the task is finished to avoid creating new task while an existing one is running

Neylix (commented):
I'm wondering if the refresh interval should be expressed in cron instead of a time. Because let's say a interval of once a day, if a node start at 00:00 and another at 10:00, there might be a big difference between the 2 data fetched.
Indeed the variance resolver should handle it, but I think a cron can provide more utilities like fetching a data from the last minute between T and T-1. This would be easier using a cron than a millisecond time.

- [x] Handle CRON intervals

lib/archethic/oracle_chain/services/hydrating_cache.ex
```diff
@@ -132,4 +134,12 @@ defmodule Archethic.OracleChain.Services.HydratingCache do
   def handle_info(:discard_value, state) do
     {:noreply, %State{state | value: nil, ttl_timer: nil}}
   end
+
+  defp next_tick_in_seconds(refresh_interval) do
+    if is_binary(refresh_interval) do
```
samuelmanzanera, line 139: You should leverage `CronParser.parse(extended_mode)` instead, it's more accurate.

samuelmanzanera: As there is not proper error handling, I think let it crash would be ok.

bchamagne: as discussed, it is fine to let it crash.

Neylix (approved):
Very nice !
Even in local with multiple nodes, we don't have limit reached warning from API :)

PR #858, opened 2023-01-23
added get_token_id function in smart contract library

### LINKED ISSUE #640: Add token ID in smart contral library

According to AEIP2, multiple tokens can be created in the same transaction chain.
There is also a possibility to mint the same token multiple times in the same chain. These token will have a different address, but graphql API `token` provide a token id which uniquely identify the token base on it's properties and genesis address.
Using this token id would be really needed for smart contract development 

Smart contract should provide a function to get this token id like `token_id(token_address)`

### Additional context

_No response_

tenmoves:
# Description

Fixes #640 

## Type of change

# How Has This Been Tested?

- unit tests 
- manual tests
image
image
image

# Checklist:

### DIFF
--- a/lib/archethic/contracts/interpreter/utils.ex
+++ b/lib/archethic/contracts/interpreter/utils.ex
@@ -287,6 +287,15 @@ defmodule Archethic.Contracts.Interpreter.Utils do
     {node, acc}
   end
 
+  # Whitelist the get_token_id/0 function
+  def prewalk(
+        node = {{:atom, "get_token_id"}, _, [_address]},
+        acc = {:ok, %{scope: scope}}
+      )
+      when scope != :root do
+    {node, acc}
+  end
+
   # Whitelist the timestamp/0 function in condition
   def prewalk(node = {{:atom, "timestamp"}, _, _}, acc = {:ok, %{scope: scope}})
       when scope != :root do
--- a/lib/archethic/contracts/interpreter/utils.ex
+++ b/lib/archethic/contracts/interpreter/utils.ex
@@ -287,7 +287,7 @@ defmodule Archethic.Contracts.Interpreter.Utils do
     {node, acc}
   end
 
-  # Whitelist the get_token_id/0 function
+  # Whitelist the get_token_id/1 function
   def prewalk(
         node = {{:atom, "get_token_id"}, _, [_address]},
         acc = {:ok, %{scope: scope}}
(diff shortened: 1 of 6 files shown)


lib/archethic/contracts/interpreter/utils.ex
```diff
@@ -270,6 +270,15 @@ defmodule Archethic.Contracts.Interpreter.Utils do
     {node, acc}
   end
 
+  # Whitelist the get_token_id/0 function
```
samuelmanzanera, line 273: Why using `token_id/0` ?
I think you could use token_id/1 in the Utils, removing the need to integrate it in the condition parser.

tenmoves: sorry it's a typo there is no get_token_id/0 only get_token_id/1
but why it would remove the need to integrate it in the condition parser ?

samuelmanzanera: I think it could be available anywhere in the actions' parser as well.

Neylix (changes requested):

lib/archethic/utils.ex
```diff
@@ -913,4 +916,62 @@ defmodule Archethic.Utils do
     |> Stream.take_while(&(NaiveDateTime.compare(&1, end_of_month_datetime) in [:lt]))
     |> Enum.count()
   end
+
+  @doc """
+  get token properties based on the genesis address and the transaction
+  """
+  @spec get_token_properties(binary(), Transaction.t()) :: {:ok, map()} | {:error, atom()}
```
Neylix, line 923: You may specify the possible atom return when an error occur, it will be easier to understand the possible result

lib/archethic/utils.ex
```diff
@@ -913,4 +916,62 @@ defmodule Archethic.Utils do
...
+      })
+      when type in [:token, :mint_rewards] do
+    case Jason.decode(content) do
+      {:ok, map} ->
+        result = %{
+          genesis: genesis_address,
+          name: Map.get(map, "name", ""),
+          supply: Map.get(map, "supply"),
+          symbol: Map.get(map, "symbol", ""),
+          type: type,
```
Neylix, line 939: The type should be retrieve from the map, the expected type are `fungible` or `non-fungible` and not the transaction type

Neylix (changes requested):

lib/archethic/utils.ex
tenmoves: done

lib/archethic/utils.ex
tenmoves: done

Neylix (approved):

PR #859, opened 2023-01-24
quorum during self repair should fetch from all nodes

### LINKED ISSUE #856: Quorum during self repair should fetch from all the node

During the self repair, we use the quorum read to fetch the missed transactions or the previous summary aggregate.
But if there is some synchronization error, the 3 requested node from the quorum may return that the transaction does not exist and so the self repair crashes.
If the transaction is in the beacon summary it means that it exists and so it should be fetched during self repair.

To avoid this issue there is to options: 
- In the quorum read we pass the consistency level to the number of authorized and available nodes, so the self repair will fetch the data from all storage node. It is maybe to much and cost a lot of networking
- If the quorum read return an unexpected response (transaction does not exists, transaction invalid ...), we remove from the node list the previously requested nodes, and retry the quorum until we get the expected response or the node list is empty

tenmoves:
# Description

During the self repair, we use the quorum read to fetch the missed transactions and the previous summary aggregate.
But if there is some synchronization error, the 3 requested node from the quorum may return that the transaction does not exist and so the self repair crashes.
If the transaction is in the beacon summary it means that it exists and so it should be fetched during self repair.

If the quorum read return an unexpected response (transaction does not exists, transaction invalid ...), we remove from the node list the previously requested nodes, and retry the quorum until we get the expected response or the node list is empty

Fixes #856 

## Type of change

# How Has This Been Tested?

- Unit tests

# Checklist:

### DIFF
--- a/lib/archethic/p2p.ex
+++ b/lib/archethic/p2p.ex
@@ -741,7 +741,7 @@ defmodule Archethic.P2P do
             message,
             conflict_resolver,
             acceptance_resolver,
-            consistency_level - 1,
+            consistency_level,
             timeout,
             quorum_result
           )
--- a/lib/archethic/p2p.ex
+++ b/lib/archethic/p2p.ex
@@ -710,7 +710,7 @@ defmodule Archethic.P2P do
               message,
               conflict_resolver,
               acceptance_resolver,
-              consistency_level - 1,
+              consistency_level,
               timeout,
               quorum_result
             )
@@ -723,7 +723,7 @@ defmodule Archethic.P2P do
             message,
             conflict_resolver,
             acceptance_resolver,
-            consistency_level - 1,
+            consistency_level,
             timeout,
             result
           )
@@ -758,9 +758,7 @@ defmodule Archethic.P2P do
     else
       # If the results differ, we can apply a conflict resolver to merge the result into
       # a consistent response
-      resolved_result = conflict_resolver.(distinct_elems)
-
-      resolved_result
+      conflict_resolver.(distinct_elems)
     end
   end
 end
--- a/lib/archethic/p2p.ex
+++ b/lib/archethic/p2p.ex
@@ -633,10 +633,19 @@ defmodule Archethic.P2P do
 
   def quorum_read(nodes, message, conflict_resolver, timeout, consistency_level) do
     nodes
+    |> filter_and_prioritize_nodes_for_quorum_read()
+    |> do_quorum_read(message, conflict_resolver, timeout, consistency_level, nil)
+  end
+
+  @doc """
+  Filter and prioritize nodes to be used in quorum read
+  """
+  @spec filter_and_prioritize_nodes_for_quorum_read(list(Node.t())) :: list(Node.t())
+  def filter_and_prioritize_nodes_for_quorum_read(nodes) do
+    nodes
     |> Enum.filter(&Node.locally_available?/1)
     |> nearest_nodes()
     |> unprioritize_node(Crypto.first_node_public_key())
-    |> do_quorum_read(message, conflict_resolver, timeout, consistency_level, nil)
   end
 
   defp do_quorum_read([], _, _, _, _, nil), do: {:error, :network_issue}
--- a/lib/archethic/p2p.ex
+++ b/lib/archethic/p2p.ex
@@ -620,6 +620,7 @@ defmodule Archethic.P2P do
           message :: Message.t(),
           conflict_resolver :: (list(Message.t()) -> Message.t()),
           timeout :: non_neg_integer(),
+          acceptance_resolver :: (Message.t() -> boolean()),
           consistency_level :: pos_integer()
         ) ::
           {:ok, Message.t()} | {:error, :network_issue}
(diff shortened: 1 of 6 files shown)


tenmoves:
what happens if we have both network_issue and transaction_not_exists which one should we raise ?
should we expect other errors than these two mentioned ?

lib/archethic/self_repair/sync/transaction_handler.ex
```diff
@@ -66,25 +66,61 @@ defmodule Archethic.SelfRepair.Sync.TransactionHandler do
...
-
-        raise "Transaction doesn't exist"
-
-      {:error, :network_issue} ->
-        Logger.error("Cannot fetch the transaction to sync",
-          transaction_address: Base.encode16(address),
-          transaction_type: type
-        )
+      {:error, error} ->
+        do_retry_download_transaction(address, type, download_nodes, error)
```
samuelmanzanera, line 78: Why not leverage `P2P.quorum_read` but extend it to support the return of the list of responding nodes, so you could discard them in the next iteration.
Hence, all the logic wouldn't need to be reimplemented.

tenmoves: Good idea, if you think we will need to use this iteration process somewhere else.

tenmoves: at first we were thinking that it is mostly needed here so there was no need to change the api behaviour

lib/archethic/p2p.ex
```diff
@@ -620,6 +620,7 @@ defmodule Archethic.P2P do
           message :: Message.t(),
           conflict_resolver :: (list(Message.t()) -> Message.t()),
           timeout :: non_neg_integer(),
+          acceptance_resolver :: (Message.t() -> boolean()),
```
samuelmanzanera, line 623: I think we should move it to the end (last parameter), to be backward compatible

lib/archethic/p2p.ex
```diff
@@ -699,12 +754,13 @@ defmodule Archethic.P2P do
...
     if length(distinct_elems) == 1 do
-      {:ok, List.first(distinct_elems)}
+      List.first(distinct_elems)
     else
       # If the results differ, we can apply a conflict resolver to merge the result into
       # a consistent response
       resolved_result = conflict_resolver.(distinct_elems)
-      {:ok, resolved_result}
+
+      resolved_result
```
samuelmanzanera, line 763: I think to reduce boilerplate, you could handle the acceptance criteria there and return {:ok, _} or {:error, _}, then catch on the do_quorum_read to retry or not, like:

```elixir
case  do_quorum(results, conflict_resolver) do
   {:ok, result} ->
     {:ok, result}
   {:error, _} ->
      do_read_quorum(rest, ....)
end
```

samuelmanzanera (changes requested):

lib/archethic/p2p.ex
tenmoves: the problem is we will need to pass nodes to do_quorum and it will be a cyclic dependecy between do_read_quorum ad do_quorum

lib/archethic/p2p.ex
tenmoves: I checked we don't use consistency_level anywhere in the project that's why I put it last as it will always take default value

lib/archethic/p2p.ex
samuelmanzanera: Why will we need to pass nodes to `do_quorum` ?

tenmoves: maybe i didn't understand you correctly,  can you explain a bit more your first comment

samuelmanzanera: ok I get, it's ok I think

lib/archethic/p2p.ex
```diff
@@ -675,22 +692,60 @@ defmodule Archethic.P2P do
...
+
+          if acceptance_resolver.(quorum_result) do
+            {:ok, quorum_result}
+          else
+            do_quorum_read(
+              rest,
+              message,
+              conflict_resolver,
+              acceptance_resolver,
+              consistency_level - 1,
```
Neylix, line 713: I think you should not reduce the consistency level as we want at least 3 results (by defaut), so here only 2 nodes will be asked instead of 3. Then 1, then 0.

Neylix (changes requested):

lib/archethic/p2p.ex
tenmoves: ty

test/archethic/p2p_test.exs
```diff
@@ -135,6 +135,36 @@ defmodule Archethic.P2PTest do
...
+      )
+      |> expect(
+        :send_message,
+        1,
+        fn _node, _message, _timeout ->
+          {:ok, %NotFound{}}
+        end
+      )
+
+      assert {:ok, %NotFound{}} =
```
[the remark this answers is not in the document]
tenmoves: I can't reproduce it on my machine it's weird, normally this shouldn't happen but I will add a timer.sleep() before the NotFound message to make sure that the other messages have arrived.

test/archethic/self_repair/sync/transaction_handler_test.exs
```diff
@@ -130,6 +130,37 @@ defmodule Archethic.SelfRepair.Sync.TransactionHandlerTest do
              )
   end
 
+  test "download_transaction/2 should download the transaction even after a first failure" do
```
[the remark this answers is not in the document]
tenmoves: you are right

Neylix (approved):

PR #861, opened 2023-01-25
SC: regex_scan/2 & regex_replace/3

bchamagne:
Add 2 library functions to manipulate string with regular expressions.
These can be used in both `condition` & `action` blocks

### DIFF
--- a/lib/archethic/contracts/interpreter/library.ex
+++ b/lib/archethic/contracts/interpreter/library.ex
@@ -56,6 +56,58 @@ defmodule Archethic.Contracts.Interpreter.Library do
     end
   end
 
+  @doc """
+  Extract data from string using capture groups
+
+  ## Examples
+
+      iex> Library.regex_scan("foo", "bar")
+      []
+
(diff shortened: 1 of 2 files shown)


lib/archethic/contracts/interpreter/library.ex
```diff
@@ -56,6 +56,65 @@ defmodule Archethic.Contracts.Interpreter.Library do
...
+      iex> Library.regex_scan("A0B1C2,123\\nD3E4F5,456\\n", "^(\\\\w+),(\\\\d+)$")
+      [["A0B1C2", "123"], ["D3E4F5", "456"]]
+
+  """
+  @spec regex_scan(binary(), binary()) :: list(binary())
+  def regex_scan(text, pattern) when is_binary(text) and is_binary(pattern) do
+    case Regex.compile(pattern, "m") do
+      {:ok, pattern} ->
+        Regex.scan(pattern, text, capture: :all_but_first)
+        |> Enum.map(fn
```
Neylix, line 85: You can use `Enum.flat_map(& &1)`

Neylix:
I think this PR can be closed as these functions are added in the PR #862 which also add a lot of other utils function

PR #862, opened 2023-01-25
SC: plenty of utils funcs

bchamagne:
This branch should be merged onto|after branch `sc-regex`. (ignore first 2 commits)

This PR adds plenty of utils functions (available in both `condition` & `action` blocks)

### DIFF
--- a/lib/archethic/contracts/interpreter/library.ex
+++ b/lib/archethic/contracts/interpreter/library.ex
@@ -241,6 +241,64 @@ defmodule Archethic.Contracts.Interpreter.Library do
   def size(map) when is_map(map), do: map_size(map)
 
   @doc """
+  Append item to the list (slow)
+
+  ## Examples
+
+    iex> Library.append(1, [])
+    [1]
+
+    iex> Library.append([2], [1])
+    [1, [2]]
+  """
+  @spec append(any(), list(any())) :: list(any())
+  def append(item, list) do
+    list ++ [item]
+  end
+
+  @doc """
+
+  Prepend item to the list (fast)
+
+  ## Examples
+    iex> Library.prepend(1, [])
+    [1]
+
+    iex> Library.prepend( [2], [1])
+    [[2], 1]
+  """
+  @spec prepend(any(), list(any())) :: list(any())
+  def prepend(item, list) do
+    [item | list]
+  end
+
+  @doc """
+  Concat both list
+
+  ## Examples
+
+    iex> Library.concat([], [])
+    []
+
+    iex> Library.concat("", "")
+    ""
+
+    iex> Library.concat("hello", " world")
+    "hello world"
+
+    iex> Library.concat([1,2], [3,4])
+    [1,2,3,4]
+  """
+  @spec concat(list() | String.t(), list() | String.t()) :: list() | String.t()
+  def concat(list1, list2) when is_list(list1) and is_list(list2) do
+    list1 ++ list2
+  end
+
+  def concat(str1, str2) when is_binary(str1) and is_binary(str2) do
+    str1 <> str2
+  end
+
+  @doc """
   Get the genesis address of the chain
 
   """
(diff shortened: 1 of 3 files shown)


tenmoves (approved):

lib/archethic/contracts/interpreter/library.ex
```diff
@@ -181,6 +300,90 @@ defmodule Archethic.Contracts.Interpreter.Library do
...
+    Map.put(map, key, value)
+  end
+
+  @doc """
+  Return the head of list
+
+  iex> Library.head([1,2,3])
+  1
+  """
+  @spec head(list()) :: any()
```
samuelmanzanera, line 382: Should we do a function for this or implement list[0]  ?

lib/archethic/contracts/interpreter/library.ex
```diff
@@ -9,6 +9,66 @@ defmodule Archethic.Contracts.Interpreter.Library do
...
+    2
+
+    iex> Library.at([], 0)
+    nil
+
+    iex> Library.at([], 0, "hello")
+    "hello"
+  """
+  def at(list, idx, default \\ nil) do
+    Enum.at(list, idx, default)
```
samuelmanzanera, line 27: Should we do a function for this or support list[i] ?

bchamagne: TBH, I think we should all-in on the imperative way. Most people are used to it. Of course it's more work for the interpreter but that shouldn't be a concern.

loop: 
```
for (idx in list) do 

end
```

list access:
```
list[idx]
```

samuelmanzanera: I think the list access by index is indeed needed.

lib/archethic/contracts/interpreter/library.ex
```diff
@@ -9,6 +9,66 @@ defmodule Archethic.Contracts.Interpreter.Library do
...
+  @doc """
+  Convert a string to an integer
+
+  ## Examples
+
+    iex> Library.int("12345")
+    12345
+  """
+  @spec int(String.t()) :: integer()
+  def int(text) do
```
Neylix, line 39: I think the function name could be more explicit so that anybody could understand by reading the code. Maybe something like `string_to_int`, `string_to_float`, `number_to_string`.

lib/archethic/contracts/interpreter/library.ex
```diff
@@ -56,6 +116,65 @@ defmodule Archethic.Contracts.Interpreter.Library do
...
+      iex> Library.regex_scan("A0B1C2,123\\nD3E4F5,456\\n", "^(\\\\w+),(\\\\d+)$")
+      [["A0B1C2", "123"], ["D3E4F5", "456"]]
+
+  """
+  @spec regex_scan(binary(), binary()) :: list(binary())
+  def regex_scan(text, pattern) when is_binary(text) and is_binary(pattern) do
+    case Regex.compile(pattern, "m") do
+      {:ok, pattern} ->
+        Regex.scan(pattern, text, capture: :all_but_first)
+        |> Enum.map(fn
```
Neylix, line 145: You can use `Enum.flat_map(& &1)`

Neylix (changes requested):

lib/archethic/contracts/interpreter/library.ex
bchamagne: I don't want to flatten everything, I only want to flat if there's one capture group, similar behaviour as `regex_extract`.

lib/archethic/contracts/interpreter/library.ex
samuelmanzanera: After discussion and thinking, we might leverage a more functional approach. Just we need to ensure the naming of the function would be straightforward to understand without the knowledge of Elixir or functional programming

bchamagne:
I will include this work in another branch/PR

PR #863, opened 2023-01-25
SC: ActionInterpreter.execute might return nil to not create a next tx

bchamagne:
# Description

This PRs allows smart contracts to end without a "next_transaction".
I know @Neylix wants to go one step further (allow contract to have a next transaction even when it ends on a non-transaction-statements) but I believe it should be done in a new issue. (CREATED: https://github.com/archethic-foundation/archethic-node/issues/870)

## Type of change

# How Has This Been Tested?

Tested with a smart contract that use this code:

```
actions triggered_by: interval, at: "0 * * * *" do	
	calls = get_calls() 
	if calls != [] do 
		[...]
		add_uco_transfers(transfers)
	end
end
```

# Checklist:

### DIFF
--- a/lib/archethic/contracts/worker.ex
+++ b/lib/archethic/contracts/worker.ex
@@ -103,11 +103,16 @@ defmodule Archethic.Contracts.Worker do
       contract_transaction = Constants.to_transaction(contract_constants)
 
       with true <- ConditionInterpreter.valid_conditions?(transaction_condition, constants),
-           next_tx <- ActionInterpreter.execute(Map.fetch!(triggers, :transaction), constants),
+           next_tx when not is_nil(next_tx) <-
+             ActionInterpreter.execute(Map.fetch!(triggers, :transaction), constants),
            {:ok, next_tx} <- chain_transaction(next_tx, contract_transaction) do
         handle_new_transaction(next_tx)
         {:noreply, state}
       else
+        nil ->
+          # contract did not create a next tx
+          {:noreply, state}
+
         false ->
           Logger.debug("Incoming transaction didn't match the condition",
             transaction_address: Base.encode16(incoming_tx.address),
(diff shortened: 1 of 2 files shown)


lib/archethic/contracts/worker.ex
```diff
@@ -103,11 +103,16 @@ defmodule Archethic.Contracts.Worker do
       contract_transaction = Constants.to_transaction(contract_constants)
 
       with true <- ConditionInterpreter.valid_conditions?(transaction_condition, constants),
-           next_tx <- ActionInterpreter.execute(Map.fetch!(triggers, :transaction), constants),
+           next_tx when not is_nil(next_tx) <-
```
samuelmanzanera, line 106: It might be better to pattern match on the Transaction struct instead of `not nil`

samuelmanzanera (changes requested):

lib/archethic/contracts/worker.ex
```diff
@@ -218,18 +223,26 @@ defmodule Archethic.Contracts.Worker do
...
-             next_tx <- ActionInterpreter.execute(Map.fetch!(triggers, :oracle), constants),
+             next_tx = %Transaction{} <-
+               ActionInterpreter.execute(Map.fetch!(triggers, :oracle), constants),
              {:ok, next_tx} <- chain_transaction(next_tx, contract_transaction),
              :ok <- ensure_enough_funds(next_tx, address) do
           handle_new_transaction(next_tx)
         else
+          nil ->
+            # returned by ActionInterpreter.execute when contract did not create a next tx
+            Logger.error("Oracle contract did not trigger a new tx",
```
samuelmanzanera, line 242: Why should it be an error ?

Neylix: Indeed it should not be an error as it may be expected. I alos think that the log error when the conditions are not valids should not be an error also, but just an information.

Neylix (approved):

Issue #852, opened 2023-01-19
SC: Worker should not be executed while replicating a transaction from self repair


Neylix:
### Describe the problem you discovered

During the replication of a transaction, if the transaction has a recipient, the storage node triggers the execution of the corresponding smart contract.
This works as expected while replicating a new transaction, but the execution should not be triggered if the replication comes from the self repair as it is a past transaction and the execution has already been done by other nodes.

### Describe the solution you'd like

In the Replication module, we should pass the parameter `self_repair?` to the `Contracts.load_transaction` and do not execute smart contract if it comes from self repair.

PR #864, opened 2023-01-25
SC: Worker should not be executed while replicating a transaction from self repair#852

apoorv-2204:
# Description

Fixes # 

- [ ] #852 

## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/lib/archethic/contracts/loader.ex
+++ b/lib/archethic/contracts/loader.ex
@@ -34,7 +34,7 @@ defmodule Archethic.Contracts.Loader do
       _ -> true
     end)
     |> Stream.map(fn {:ok, tx} -> tx end)
-    |> Stream.each(&load_transaction(&1, true))
+    |> Stream.each(&load_transaction(&1, from_db: true))
     |> Stream.run()
 
     {:ok, []}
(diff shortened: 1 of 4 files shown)


lib/archethic/replication.ex
```diff
@@ -564,14 +565,15 @@ defmodule Archethic.Replication do
   - Transactions with smart contract deploy instances of them or can put in pending state waiting approval signatures
   - Code approval transactions may trigger the TestNets deployments or hot-reloads
   """
-  @spec ingest_transaction(Transaction.t(), boolean()) :: :ok
-  def ingest_transaction(tx = %Transaction{}, io_transaction?) do
+  @spec ingest_transaction(Transaction.t(), boolean(), boolean()) :: :ok
+  def ingest_transaction(tx = %Transaction{}, io_transaction?, self_repair? \\ false) do
```
Neylix, line 569: I think there is no need to have a default value, the parameter is always present when we call this function. Just need to add a false parameter in bootstrap_test

lib/archethic/replication.ex
```diff
@@ -564,14 +565,15 @@ defmodule Archethic.Replication do
...
+    # also called from bootstrap
```
Neylix, line 570: I don't get this commentary, this function is not called from bootstrap. Only in replication and bootstrap_test

Neylix (changes requested):

Neylix (approved):

lib/archethic/contracts/loader.ex
```diff
@@ -43,16 +43,22 @@ defmodule Archethic.Contracts.Loader do
...
+  @spec load_transaction(Transaction.t(), list()) :: :ok
+  def load_transaction(tx, opts \\ []) do
+    from_db = Keyword.get(opts, :from_db, false)
+    from_self_repair = Keyword.get(opts, :from_self_repair, false)
 
-  def load_transaction(
+    do_load_transaction(tx, from_db, from_self_repair)
+  end
+
+  def do_load_transaction(
```
samuelmanzanera, line 54: I think it should be private.

samuelmanzanera (changes requested):

samuelmanzanera (approved):

Neylix (approved):

PR #878, opened 2023-01-31
Add algorithm to compute the network coordinates

### LINKED ISSUE #155: Network Coordinates
This epic is about the management of the network coordinates recorded in the BeaconChain and used to get the closest nodes approximation within the network

### LINKED ISSUE #159: Create a program to compute network patch
This feature should make available a program in C - to make efficient the algorithm - to build Eigenvectors and Eigenvalues from a list of N*N matrix of latencies and compute the patch

A port will be created to communicate with.
We can maybe explore as well `Nx` which supports matrix operation

samuelmanzanera:
# Description

Fixes #159 

## Type of change

# How Has This Been Tested?

TODO: need to provide more way to tests (Awaiting response from @internet-zero)

# Checklist:


lib/archethic/beacon_chain/network_coordinates.ex
```diff
@@ -0,0 +1,191 @@
...
+  @type latency_patch :: {String.t(), String.t()}
+
+  @doc """
+  Compute the network patch based on the matrix latencies
+
+  The matrix must be an Hollow matrix, symmetric and all the latencies must be positive
+
+  In physicis, the center of mass of a distribution of mass in the space is the unique point where
+  the weighted relative position of the distributed mass sums to zero. Here, the sum of vectors to
+  all nodes from the center of mass is zero (All the nodes being the unit of mass)
```
tenmoves, line 17: little typo In physics, the center of mass for a distribution of masses

apoorv-2204:
rebase?

bchamagne (commented):
I can neither approve nor deny this PR. I don't have the maths skills to read this.
I know it works when I give it an hermitian matrix, but I don't get what the matrix is in our use case. 

FYI, I had to install `bazel` on my archlinux to make XLA compile.

lib/archethic/beacon_chain/network_coordinates.ex
```diff
@@ -0,0 +1,193 @@
...
+  """
+
+  @digits ["F", "E", "D", "C", "B", "A", "9", "8", "7", "6", "5", "4", "3", "2", "1", "0"]
+
+  @type latency_patch :: {String.t(), String.t()}
+
+  @doc """
+  Compute the network patch based on the matrix latencies
+
+  The matrix must be an Hollow matrix, symmetric and all the latencies must be positive
```
tenmoves, line 13: Hollow matrix can refer to many things, maybe we could rephrase is in a more formal way like this: 
"The matrix M must respect the following conditions 
- M is hollow ie M is an n × n matrix A = (aij) where aij = 0 whenever i = j (i.e. aii = 0 for all i)
- M must be symmetric and positive for all i and for all j aij = aji and aij >= 0"

lib/archethic/beacon_chain/network_coordinates.ex
```diff
@@ -0,0 +1,193 @@
...
+  In physics, the center of mass for a distribution of masses in the space is the unique point where
+  the weighted relative position of the distributed mass sums to zero. Here, the sum of vectors to
+  all nodes from the center of mass is zero (All the nodes being the unit of mass)
+
+  The distance between center of mass and each node is then reduced to this formula using law of cosine
+  and center of mass:
+  `$$ dic^2 =  {1\over 2} \sum_{j=1}^{n} dij^2 - {1\over 2}\sum_{j=2}^{n}\sum_{k=1}^{j-1} djk^2 $$`
+
+  Then using the laws of Cosine, we can compute a gram matrix which would help to plot the nodes in a 2D plan:
+  `$$ gij =  {1\over 2} (dic^2 + djc^2 - dij^2) $$`
```
tenmoves, line 24: I can't display correctly the formula are you using some special extension ?

samuelmanzanera: I used Markdown editor like: https://stackedit.io/ with Mathematics plugin

lib/archethic/beacon_chain/network_coordinates.ex
```diff
@@ -0,0 +1,193 @@
+defmodule Archethic.BeaconChain.NetworkCoordinates do
+  @moduledoc """
+  Network coordinates is a way to map latency between nodes and used to determine the closest nodes
```
cluster, line 3: Perhaps you could provide more details on how to interpret the results and obtain the closest nodes

samuelmanzanera: This is already implemented here:

https://github.com/archethic-foundation/archethic-node/blob/39a8efde015bd7fedce84cd79decaefce911e18c/lib/archethic/p2p.ex#L348-L375

cluster: ok! thank you for pointing that out 🙂
From my understanding of `GeoPatch`, we can have a third digit that corresponds to the precision. Do you think there is a way that would make sense to calculate the precision in the case of network latency? 🤔

lib/archethic/beacon_chain/network_coordinates.ex
```diff
@@ -0,0 +1,193 @@
...
+  # end
+
+  defp compute_distance_from_center_mass(tensor) do
+    matrix_size = Nx.size(tensor[0])
+
+    a =
+      tensor
+      |> Nx.power(2)
+      |> Nx.sum(axes: [:column])
+      |> Nx.multiply(Nx.tensor(1.0 / matrix_size, type: {:f, 64}))
```
tenmoves, line 69: I think the correct syntax is type: :f64

lib/archethic/beacon_chain/network_coordinates.ex
samuelmanzanera: Yes it will be based on the bandwidth

lib/archethic/beacon_chain/network_coordinates.ex
tenmoves: so in  here a little typo the correct formula is 1/n not 1/2 (in the implementation it's correct, typo only in doc)

lib/archethic/beacon_chain/network_coordinates.ex
```diff
@@ -0,0 +1,193 @@
...
+      1..(matrix_size - 1)
+      |> Enum.map(fn i ->
+        excluded_first_row_tensor
+        |> Nx.slice([i - 1, 0], [1, i])
+        |> Nx.power(2)
+        |> Nx.sum()
+      end)
+      |> Nx.stack()
+      |> Nx.sum()
+      |> Nx.to_number()
```
tenmoves, line 83: it may be worth it to precise that that the indexes start at 0 vs 1 in the formula

tenmoves (approved):

lib/archethic/beacon_chain/network_coordinates.ex
samuelmanzanera: Usually, algorithm start indexes a 1, while programming language start indexes at 0. I guess this is a standard way of thinking.

PR #888, opened 2023-02-06
Restart self repair when it crashes

### LINKED ISSUE #871: Try to re run self repair if it fails

For some reason the self repair may fail. For example if during the self repair a node has a connection problem, it will not fetch the beacon summaries and so it will crash the self repair. But when the node is connected again, it continue to validate transaction but it's p2p view may be invalid or have missed some transaction (node shared secret for example). This causing election error.

To avoid this problem, if the self repair ends with an error, we should immediately try to rerun it until it run succesfully. This should happen for bootstrap sync and self repair scheduler I think

If the self repair crahes, the node should not continue to do things. For this we need to stop all scheduler and stop the HTTP / P2P listener.
To implement this behavior we can manage with the PubSub with up and down event

tenmoves:
# Description

Fixes #871 

## Type of change

# How Has This Been Tested?

- Unit tests
-  Simulated manual tests

# Checklist:

### DIFF
--- a/lib/archethic/pub_sub.ex
+++ b/lib/archethic/pub_sub.ex
@@ -112,19 +112,9 @@ defmodule Archethic.PubSub do
 
   @spec register_to_node_status :: {:error, {:already_registered, pid}} | {:ok, pid}
   @doc """
-  Register a process to notify node up
+  Register a process to notify node status
   """
-  def register_to_node_status() do
-    Registry.register(PubSubRegistry, :node_status, [])
-  end
-
-  @doc """
-  UnRegister a process to notification node up
-  """
-  @spec unregister_to_node_status :: :ok
-  def unregister_to_node_status() do
-    Registry.unregister(PubSubRegistry, :node_status)
-  end
+  def register_to_node_status(), do: Registry.register(PubSubRegistry, :node_status, [])
 
   @doc """
   Register a process to a new transaction publication by type
--- a/lib/archethic/pub_sub.ex
+++ b/lib/archethic/pub_sub.ex
@@ -104,27 +104,26 @@ defmodule Archethic.PubSub do
   end
 
   @doc """
-  Notify that Node is Up
+  Notify Node status (up or down)
   """
-  @spec notify_node_up() :: :ok
-  def notify_node_up() do
-    dispatch(:node_up, :node_up)
-  end
+  @spec notify_node_status(:node_up | :node_down) :: :ok
+  def notify_node_status(:node_up), do: dispatch(:node_status, :node_up)
+  def notify_node_status(:node_down), do: dispatch(:node_status, :node_down)
 
-  @spec register_to_node_up :: {:error, {:already_registered, pid}} | {:ok, pid}
+  @spec register_to_node_status :: {:error, {:already_registered, pid}} | {:ok, pid}
   @doc """
   Register a process to notify node up
   """
-  def register_to_node_up() do
-    Registry.register(PubSubRegistry, :node_up, [])
+  def register_to_node_status() do
+    Registry.register(PubSubRegistry, :node_status, [])
   end
 
   @doc """
   UnRegister a process to notification node up
   """
-  @spec unregister_to_node_up :: :ok
-  def unregister_to_node_up() do
-    Registry.unregister(PubSubRegistry, :node_up)
+  @spec unregister_to_node_status :: :ok
+  def unregister_to_node_status() do
+    Registry.unregister(PubSubRegistry, :node_status)
   end
 
   @doc """
@@ -173,6 +172,11 @@ defmodule Archethic.PubSub do
   @spec register_to_node_update() :: {:ok, pid()}
   def register_to_node_update do
     Registry.register(PubSubRegistry, :node_update, [])
+  end
+
+  @spec unregister_to_node_update() :: :ok
+  def unregister_to_node_update do
+    Registry.unregister(PubSubRegistry, :node_update)
   end
 
   @doc """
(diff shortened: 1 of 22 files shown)


lib/archethic/pub_sub.ex
```diff
@@ -104,28 +104,17 @@ defmodule Archethic.PubSub do
...
-  @spec register_to_node_up :: {:error, {:already_registered, pid}} | {:ok, pid}
-  @doc """
-  Register a process to notify node up
-  """
-  def register_to_node_up() do
-    Registry.register(PubSubRegistry, :node_up, [])
-  end
+  @spec notify_node_status(:node_up | :node_down) :: :ok
+  def notify_node_status(:node_up), do: dispatch(:node_status, :node_up)
+  def notify_node_status(:node_down), do: dispatch(:node_status, :node_down)
```
Neylix, line 111: You could simplify by doing
```elixir
def notify_node_status(status), do: dispatch(:node_status, status)
```

lib/archethic/self_repair.ex
```diff
@@ -48,12 +50,30 @@ defmodule Archethic.SelfRepair do
...
+        Logger.info("Bootstrap Sync succeded in loading missed transactions !")
+      else
+        Logger.error(
+          "Bootstrap Sync failed to load missed transactions after max retry of #{@max_retry_count} !"
+        )
+      end
 
       # At the end of self repair, if a new beacon summary as been created
       # we run bootstrap_sync again until the last beacon summary is loaded
+      start_date = DateTime.utc_now()
```
Neylix, line 75: Now you should use `SelfRepair.last_sync_date` as start_date

lib/archethic/self_repair/scheduler.ex
```diff
@@ -92,33 +94,50 @@ defmodule Archethic.SelfRepair.Scheduler do
...
+
+    if Archethic.up?() do
+      :persistent_term.put(:archethic_up, :down)
+      PubSub.notify_node_status(:node_down)
+    end
+
+    new_state = Map.update(state, :retry_count, 1, &(&1 + 1))
+
+    if new_state.retry_count > @max_retry_count do
+      schedule_sync(new_state.interval)
```
Neylix, line 135: You should add the timer returned by the function and add in in the state

test/archethic/reward/scheduler_test.exs
```diff
@@ -156,6 +156,29 @@ defmodule Archethic.Reward.SchedulerTest do
               }} = :sys.get_state(pid)
     end
 
+    test "should wait for :node_down message to stop the scheduler" do
```
Neylix, line 159: I don't understand this test, you may need to send a node_up and then ensure it stops when sending a node_down

test/archethic/shared_secrets/node_renewal_scheduler_test.exs
```diff
@@ -158,6 +158,31 @@ defmodule Archethic.SharedSecrets.NodeRenewalSchedulerTest do
               }} = :sys.get_state(pid)
     end
 
+    test "should wait for node down message to stop the scheduler" do
```
Neylix, line 161: Same as RewardScheduler test for this one

Neylix (changes requested):

lib/archethic/self_repair.ex
tenmoves: right thank you

Neylix (approved):

lib/archethic/self_repair.ex
```diff
@@ -48,17 +50,35 @@ defmodule Archethic.SelfRepair do
...
+            catch
+              _, _ -> {:cont, :error}
+            end
+          end)
+
+      if loaded_missed_transactions? do
+        Logger.info("Bootstrap Sync succeded in loading missed transactions !")
+      else
+        Logger.error(
+          "Bootstrap Sync failed to load missed transactions after max retry of #{@max_retry_count} !"
```
samuelmanzanera, line 69: Does it mean we aren't crashing the process if there is no success?

lib/archethic/self_repair/scheduler.ex
```diff
@@ -92,33 +94,56 @@ defmodule Archethic.SelfRepair.Scheduler do
...
-    # end of node sync message
     if !P2P.available_node?(), do: BootstrapSync.publish_end_of_sync()
-    update_last_sync_date(date)
     Process.demonitor(ref, [:flush])
-    {:noreply, state}
+    {:noreply, new_state}
   end
 
-  def handle_info({:DOWN, _ref, _, _, reason}, state) do
+  def handle_info({:DOWN, _ref, _, _, reason}, state = %{interval: interval}) do
```
samuelmanzanera, line 124: The DOWN message is received, even in the case of success, only when the task process will be terminated.
So, this is not a failure and the retry is irrelevant in that case.

samuelmanzanera: Instead, you might use a try catch in the `Task.async` and return `error` in the catch.
Hence, you could pattern match on the `ok/error` tuple to determine the success or the retry.

Neylix: The DOWN event is not always receive because in the handle info of a success task ({ref, response}) we use `Process.demonitor(ref [:flush])` In that case the message DOWN will not be receive by the scheduler. Otherwise we should have a failed self-repair error log after all self repair even if it is successfull

samuelmanzanera: ok

lib/archethic/self_repair.ex
tenmoves: no we keep it alive, and see if the self repair will work on the next cycle

samuelmanzanera (approved):

PR #892, opened 2023-02-06
Fix bootstrap node index

Neylix:
# Description

Fix a bug when a node goes in first_initialization during bootstrap while the node already exists on the network. The node was not able to correctly retrieve it's index to create the new node transaction in the chain.

Also fix an bug in pending transaction validation while verifying if the transaction already exists

## Type of change

# How Has This Been Tested?

Start 2 nodes, stop 1 and delete it's DB, restart it
It should create a new node transaction in it's node chain.
Then wait for next self repair and try to update the node reward address (using endpoint/settings) and it should create a new transaction in the node chain.

# Checklist:

### DIFF
--- a/lib/archethic/bootstrap.ex
+++ b/lib/archethic/bootstrap.ex
@@ -115,7 +115,6 @@ defmodule Archethic.Bootstrap do
         http_port,
         transport,
         bootstrapping_seeds,
-        last_sync_date,
         network_patch,
         reward_address
       )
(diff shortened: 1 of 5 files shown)


lib/archethic/bootstrap.ex
```diff
@@ -300,19 +301,19 @@ defmodule Archethic.Bootstrap do
...
 
-      _ ->
-        :skip
-    end
+    Crypto.set_node_key_index(length)
 
     tx =
       TransactionHandler.create_node_transaction(ip, port, http_port, transport, reward_address)
 
+    # Reset node index as it will be handled by node transaction replication during self repair
```
samuelmanzanera, line 314: To avoid that, we could use `Transaction.new/3` and pass as argument in `create_node_transaction` the index

samuelmanzanera (approved):

Issue #872, opened 2023-01-26
Welcome node should forward tx if not synced


Neylix:
### Describe the problem you discovered

When a welcome node receive a transaction, it will perform the election for validation nodes and send to them the StartMining message.
But if the welcome node is out of sync and don't have the last node shared secret, it's election will be wrong (election or proof of election).

### Describe the solution you'd like

To fix this, when receiving a transaction, the welcome node should verify that it has the last node shared secret transaction.
If yes, continue as actually, if not we have 3 options:
- Forward the transaction to another authorized node
- Request the last node shared secret transaction and replicate it, then continue to send the transaction (if multiple transaction are received, the node may try to replicate multiple time the shared secret)
- Forward the transaction to another authorized node, and try to replicate the last node shared secret (if multiple transaction are received, the node may try to replicate multiple time the shared secret)

We should take care to continue to have the right welcome node which is the first one the client sent his transaction

internet-zero:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Product-and-Design-Workspace-61efe71b9083a90018250d89/issues/archethic-foundation/archethic-node/872?planning-poker) with Zenhub @Neylix @samuelmanzanera @apoorv-2204 @bchamagne @netboz @tenmoves

PR #896, opened 2023-02-07
Welcome node should forward tx if not synced#872

apoorv-2204:
# Description

When it is the time for protocol to create a new node renewal transaction, there exist a time window where new transactions arrive and election results with anomaly due to missing latest Node shared secret transaction.

###  Solutions and  Further improvements
 - Check for whether a node is synced for every transaction
 - If not schedule a request for resync.
 - Last nss transaction validation stamp and last scheduling date to determine node status.
 
 Drawbacks
 - NSS chain has to be checked for each transaction

 Solution

- [ ] #872 

## Type of change

# How Has This Been Tested?
- manual 

# Checklist:

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -7,15 +7,14 @@ config :git_hooks,
     pre_push: [
       tasks: [
         {:cmd, "mix clean"},
-        {:cmd, "mix git_hooks.install"},
-        {:cmd, "mix hex.outdated --within-requirements"},
         {:cmd, "mix format --check-formatted"},
         {:cmd, "mix compile --warnings-as-errors"},
         {:cmd, "mix credo"},
         {:cmd, "mix sobelow"},
         {:cmd, "mix knigge.verify"},
         {:cmd, "mix test --trace"},
-        {:cmd, "mix dialyzer"}
+        {:cmd, "mix dialyzer"},
+        {:cmd, "mix check.updates"}
       ]
     ]
   ]
--- a/lib/archethic.ex
+++ b/lib/archethic.ex
@@ -3,27 +3,15 @@ defmodule Archethic do
   Provides high level functions serving the API and the Explorer
   """
 
-  alias __MODULE__.Account
-  alias __MODULE__.BeaconChain
-  alias __MODULE__.Crypto
-  alias __MODULE__.Election
-  alias __MODULE__.P2P
-  alias __MODULE__.P2P.Node
-  alias __MODULE__.P2P.Message.Balance
-  alias __MODULE__.P2P.Message.GetBalance
-  alias __MODULE__.P2P.Message.GetCurrentSummaries
-  alias __MODULE__.P2P.Message.GetTransactionSummary
-  alias __MODULE__.P2P.Message.TransactionSummaryMessage
-  alias __MODULE__.P2P.Message.NewTransaction
-  alias __MODULE__.P2P.Message.NotFound
-  alias __MODULE__.P2P.Message.Ok
-  alias __MODULE__.P2P.Message.StartMining
-  alias __MODULE__.P2P.Message.TransactionSummaryList
+  alias Archethic.SharedSecrets
+  alias __MODULE__.{Account, BeaconChain, Crypto, Election, P2P, P2P.Node, P2P.Message}
+  alias __MODULE__.{SelfRepair, TransactionChain}
 
-  alias __MODULE__.TransactionChain
-  alias __MODULE__.TransactionChain.Transaction
-  alias __MODULE__.TransactionChain.TransactionInput
-  alias __MODULE__.TransactionChain.TransactionSummary
+  alias Message.{NewTransaction, NotFound, StartMining, TransactionSummaryList}
+  alias Message.{Balance, GetBalance, GetCurrentSummaries, GetTransactionSummary}
+  alias Message.{StartMining, Ok, TransactionSummaryMessage}
+
+  alias TransactionChain.{Transaction, TransactionInput, TransactionSummary}
 
   require Logger
 
@@ -69,26 +57,42 @@ defmodule Archethic do
         welcome_node_key \\ Crypto.first_node_public_key()
       ) do
     if P2P.authorized_and_available_node?() do
-      do_send_transaction(tx, welcome_node_key)
+      case SharedSecrets.verify_synchronization() do
+        :ok ->
+          do_send_transaction(tx, welcome_node_key)
+
+        :error ->
+          forward_transaction(tx, welcome_node_key)
+
+        {:error, last_address_to_sync} ->
+          SelfRepair.resync(last_address_to_sync)
+          forward_transaction(tx, welcome_node_key)
+      end
     else
-      P2P.authorized_and_available_nodes()
-      |> Enum.filter(&Node.locally_available?/1)
-      |> P2P.nearest_nodes()
-      |> forward_transaction(tx)
+      # node not authorized
+      forward_transaction(tx, welcome_node_key)
     end
   end
 
-  defp forward_transaction([node | rest], tx) do
-    case P2P.send_message(node, %NewTransaction{transaction: tx}) do
+  defp forward_transaction(
+         tx,
+         welcome_node_key,
+         nodes \\ P2P.authorized_and_available_nodes()
+         |> Enum.filter(&Node.locally_available?/1)
+         |> P2P.nearest_nodes()
+       )
+
+  defp forward_transaction(tx, welcome_node_key, [node | rest]) do
+    case P2P.send_message(node, %NewTransaction{transaction: tx, welcome_node: welcome_node_key}) do
       {:ok, %Ok{}} ->
         :ok
 
       {:error, _} ->
-        forward_transaction(rest, tx)
+        forward_transaction(tx, welcome_node_key, rest)
     end
   end
 
-  defp forward_transaction([], _), do: {:error, :network_issue}
+  defp forward_transaction(_, _, []), do: {:error, :network_issue}
 
   defp do_send_transaction(tx = %Transaction{type: tx_type}, welcome_node_key) do
     current_date = DateTime.utc_now()
--- a/lib/archethic/bootstrap/transaction_handler.ex
+++ b/lib/archethic/bootstrap/transaction_handler.ex
@@ -36,7 +36,10 @@ defmodule Archethic.Bootstrap.TransactionHandler do
   @spec do_send_transaction(list(Node.t()), Transaction.t()) ::
           :ok
   defp do_send_transaction([node | rest], tx) do
-    case P2P.send_message(node, %NewTransaction{transaction: tx}) do
+    case P2P.send_message(node, %NewTransaction{
+           transaction: tx,
+           welcome_node: node.first_public_key
+         }) do
       {:ok, %Ok{}} ->
         Logger.info("Waiting transaction validation",
           transaction_address: Base.encode16(tx.address),
--- a/lib/archethic/p2p.ex
+++ b/lib/archethic/p2p.ex
@@ -2,22 +2,11 @@ defmodule Archethic.P2P do
   @moduledoc """
   Handle P2P node discovery and messaging
   """
-  alias Archethic.Crypto
+  alias Archethic.{Crypto, TaskSupervisor, TransactionChain}
 
-  alias __MODULE__.BootstrappingSeeds
-  alias __MODULE__.Client
-  alias __MODULE__.GeoPatch
-  alias __MODULE__.MemTable
-  alias __MODULE__.MemTableLoader
-  alias __MODULE__.Message
-  alias __MODULE__.Node
+  alias Archethic.{TransactionChain.Transaction, Utils}
 
-  alias Archethic.TaskSupervisor
-
-  alias Archethic.TransactionChain
-  alias Archethic.TransactionChain.Transaction
-
-  alias Archethic.Utils
+  alias __MODULE__.{BootstrappingSeeds, Client, GeoPatch, MemTable, MemTableLoader, Message, Node}
 
   require Logger
 
--- a/lib/archethic/self_repair.ex
+++ b/lib/archethic/self_repair.ex
@@ -4,23 +4,11 @@ defmodule Archethic.SelfRepair do
   the bootstrapping phase and stores last synchronization date after each cycle.
   """
 
-  alias __MODULE__.Notifier
-  alias __MODULE__.NotifierSupervisor
-  alias __MODULE__.RepairRegistry
-  alias __MODULE__.RepairWorker
-  alias __MODULE__.Scheduler
-  alias __MODULE__.Sync
+  alias __MODULE__.{Notifier, NotifierSupervisor, RepairRegistry, RepairWorker}
+  alias __MODULE__.{Scheduler, Sync}
 
-  alias Archethic.{
-    BeaconChain,
-    Crypto,
-    Utils,
-    Contracts,
-    TransactionChain,
-    Election
-  }
-
-  alias Archethic.P2P.Node
+  alias Archethic.{BeaconChain, Crypto, Utils, Contracts, TransactionChain, Election}
+  alias Archethic.{P2P.Node, SharedSecrets, OracleChain, Reward, Replication}
 
   alias Crontab.CronExpression.Parser, as: CronParser
   alias Crontab.Scheduler, as: CronScheduler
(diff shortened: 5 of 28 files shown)


lib/archethic/shared_secrets.ex
```diff
@@ -212,4 +218,86 @@ defmodule Archethic.SharedSecrets do
...
+    validate_scheduling_time() || validate_last_address()
+  end
+
+  @spec validate_scheduling_time() :: boolean()
+  def validate_scheduling_time() do
+    last_scheduling_date =
+      DateTime.utc_now()
+      |> __MODULE__.get_last_scheduling_date()
+
+    case genesis_address(:node_shared_secrets) do
```
Neylix, line 263: You could simplify this function using a with statement: 
```elixir
with last_scheduling_date <- get_last_scheduling_date(DateTime.utc_now()),
     genesis_address when not is_nil(genesis_address) <-
       genesis_address(:node_shared_secrets),
     {:ok, %Transaction{validation_stamp: %ValidationStamp{timestamp: validation_timestamp}}} <-
       TransactionChain.get_last_transaction(genesis_address, validation_stamp: [:timestamp]) do
  validation_timestamp >= last_scheduling_date
else
  _ -> false
end
```

lib/archethic/shared_secrets.ex
```diff
@@ -212,4 +218,86 @@ defmodule Archethic.SharedSecrets do
...
+
+          _ ->
+            false
+        end
+    end
+  end
+
+  @spec validate_last_address() :: boolean()
+  def validate_last_address() do
+    case genesis_address(:node_shared_secrets) do
```
Neylix, line 285: Same as previous function, you can use a with statement to better handle errors

lib/archethic/self_repair/resync.ex
```diff
@@ -0,0 +1,55 @@
...
+
+  @spec node_out_of_sync() :: :ok
+  def node_out_of_sync() do
+    first_address = Crypto.genesis_address(:node_shared_secrets)
+
+    case SelfRepair.repair_in_progress?(first_address) do
+      false ->
+        SelfRepair.start_worker(
+          first_address: first_address,
+          storage_address: first_address,
```
Neylix, line 18: Here the RepairWorker expect the storage address to be the transaction to replicate. It will not fetch the last transaction of the chain and replicate it. So here you should provide the transaction you want to replicate (in fact the last nss transaction).
To achieve this, maybe the function `P2P.is_synced_node?()` could be renamed to `P2P.verify_synchronization` and this funtion could return either `:ok` if it is synchronized or `{:error, last_nss_address}`. So the right address to synchronize could be passed to the `Resync.node_out_of_sync` function.

Neylix (changes requested):

lib/archethic.ex
```diff
@@ -60,17 +47,36 @@ defmodule Archethic do
...
         welcome_node_key \\ Crypto.first_node_public_key()
       ) do
-    if P2P.authorized_and_available_node?() do
-      do_send_transaction(tx, welcome_node_key)
-    else
-      P2P.authorized_and_available_nodes()
-      |> Enum.filter(&Node.locally_available?/1)
-      |> P2P.nearest_nodes()
-      |> forward_transaction(tx)
+    handle_new_tx(tx, welcome_node_key)
```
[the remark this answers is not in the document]
apoorv-2204: sorry , my bad

lib/archethic.ex
```diff
@@ -69,16 +56,32 @@ defmodule Archethic do
...
 
-  defp forward_transaction([node | rest], tx) do
+  defp forward_transaction(
+         tx,
+         nodes \\ P2P.authorized_and_available_nodes()
+         |> Enum.filter(&Node.locally_available?/1)
+         |> P2P.nearest_nodes()
+       )
+
+  defp forward_transaction(tx, [node | rest]) do
```
Neylix, line 84: The parameter of the recursive function call are inverted on line 90

lib/archethic/shared_secrets.ex
```diff
@@ -212,4 +218,83 @@ defmodule Archethic.SharedSecrets do
...
+      false ->
+        validate_last_address()
+    end
+  end
+
+  @spec validate_scheduling_time() :: boolean()
+  def validate_scheduling_time() do
+    with genesis_address when is_binary(genesis_address) <- genesis_address(:node_shared_secrets),
+         last_scheduling_date <- get_last_scheduling_date(DateTime.utc_now()),
+         {:ok, %Transaction{validation_stamp: %{__struct__: _, timestamp: validation_timestamp}}} <-
```
Neylix, line 268: `__struct__` should be the real struct name `ValidationStamp`

lib/archethic/shared_secrets.ex
```diff
@@ -212,4 +218,83 @@ defmodule Archethic.SharedSecrets do
...
+      validation_timestamp >= last_scheduling_date
+    else
+      _ -> false
+    end
+  end
+
+  @spec validate_last_address() :: {:ok, :node_in_sync} | {:error, binary() | :node_out_of_sync}
+  def validate_last_address() do
+    case genesis_address(:node_shared_secrets) do
+      nil ->
```
Neylix, line 279: It's not necessary to verify if the genesis_address is not nil as this function will never be called if the self repair failed which assign the genesis addresses.

lib/archethic.ex
```diff
@@ -69,16 +56,32 @@ defmodule Archethic do
         welcome_node_key \\ Crypto.first_node_public_key()
       ) do
     if P2P.authorized_and_available_node?() do
-      do_send_transaction(tx, welcome_node_key)
+      case P2P.verify_synchronization() do
+        {:ok, :node_in_sync} ->
+          do_send_transaction(tx, welcome_node_key)
+
+        {:error, :node_out_of_sync} ->
```
Neylix, line 63: I think we should not start a repair if the verify synchronization did not retrieve the last_address. Starting a resync makes no sensse as we don't know if we are sync or not. So the verify_synchronization could just return :ok, {:error, last_address} or :error.
In case of just error we forward the tx without starting a repair.

Neylix (changes requested):

test/archethic/shared_secrets_test.exs
```diff
@@ -26,4 +28,109 @@ defmodule Archethic.SharedSecretsTest do
       assert SharedSecrets.has_origin_public_key?(pb_key)
     end
   end
+
+  describe "is_synchronized?" do
+    setup do
```
Neylix, line 33: Can't you use `ArchethicCase` which provide a lot of default mock ?

apoorv-2204: to make test asynchronous.

lib/archethic.ex
```diff
@@ -69,26 +56,41 @@ defmodule Archethic do
...
-  defp forward_transaction([node | rest], tx) do
+  defp forward_transaction(
+         tx,
+         nodes \\ P2P.authorized_and_available_nodes()
+         |> Enum.filter(&Node.locally_available?/1)
+         |> P2P.nearest_nodes()
+       )
+
+  defp forward_transaction(tx, [node | rest]) do
     case P2P.send_message(node, %NewTransaction{transaction: tx}) do
```
Neylix, line 84: When a node forward a transaction, the destination node will use the sender of the `NewTransaction` message as the welcome node, but there is some case where we can chain multiple forward. If so, the original welcome node will be loosed.
We should add the original welcome node in the `NewTransaction` message so in all case, the right welcome node will be provided to the validation nodes

Neylix (changes requested):

lib/archethic/bootstrap/transaction_handler.ex
```diff
@@ -36,7 +36,10 @@ defmodule Archethic.Bootstrap.TransactionHandler do
   @spec do_send_transaction(list(Node.t()), Transaction.t()) ::
           :ok
   defp do_send_transaction([node | rest], tx) do
-    case P2P.send_message(node, %NewTransaction{transaction: tx}) do
+    case P2P.send_message(node, %NewTransaction{
+           transaction: tx,
+           welcome_node: Crypto.first_node_public_key()
```
Neylix, line 41: At this point, the real welcome node key will be the the one which will receive the transaction `node.first_public_key`. Because here the node act as a client.

Neylix (changes requested):

Neylix (approved):

lib/archethic/shared_secrets.ex
```diff
@@ -212,4 +218,76 @@ defmodule Archethic.SharedSecrets do
...
+      true ->
+        {:ok, :node_in_sync}
+
+      false ->
+        validate_last_address()
+    end
+  end
+
+  @spec validate_scheduling_time() :: boolean()
+  def validate_scheduling_time() do
```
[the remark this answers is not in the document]
samuelmanzanera: This function should be either `valid_scheduling_time?` or returning ok/error tuple

lib/archethic/shared_secrets.ex
```diff
@@ -212,4 +218,76 @@ defmodule Archethic.SharedSecrets do
...
+         {:ok, %Transaction{validation_stamp: %{__struct__: _, timestamp: validation_timestamp}}} <-
+           TransactionChain.get_last_transaction(genesis_address, validation_stamp: [:timestamp]) do
+      validation_timestamp >= last_scheduling_date
+    else
+      _ -> false
+    end
+  end
+
+  @spec validate_last_address() :: {:ok, :node_in_sync} | {:error, binary() | :node_out_of_sync}
+  def validate_last_address() do
```
samuelmanzanera, line 277: Why is this function public ?

lib/archethic/shared_secrets.ex
```diff
@@ -212,4 +218,76 @@ defmodule Archethic.SharedSecrets do
...
+    with genesis_address when is_binary(genesis_address) <-
+           genesis_address(:node_shared_secrets),
+         {:ok, quroum_last_address} when is_binary(quroum_last_address) <-
+           TransactionChain.resolve_last_address(genesis_address),
+         {local_last_address, _} <- TransactionChain.get_last_address(genesis_address),
+         {true, _} <- {quroum_last_address == local_last_address, quroum_last_address} do
+      {:ok, :node_in_sync}
```
samuelmanzanera, line 284: I think the `ok` should be sufficient

samuelmanzanera (changes requested):

Makefile
```diff
@@ -6,6 +6,7 @@ TPMFLAGS = -ltss2-esys -ltss2-rc -ltss2-mu -lcrypto
 all: compile_c_programs
 
 compile_c_programs:
+	mix git_hooks.install
```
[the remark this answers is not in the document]
apoorv-2204: if we try to do git push origin ... it breaks the call due to git hooks not installed,
Since compile c programs is ran every time,

config/config.exs
```diff
@@ -7,8 +7,7 @@ config :git_hooks,
     pre_push: [
       tasks: [
         {:cmd, "mix clean"},
-        {:cmd, "mix git_hooks.install"},
-        {:cmd, "mix hex.outdated --within-requirements"},
```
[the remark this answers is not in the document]
apoorv-2204: if we try to do git push origin ... it breaks the call , hence we should not always update deps with each push call.
The mix git_hooks,install is not helping here

lib/archethic_web/channels/user_socket.ex
```diff
@@ -8,6 +8,7 @@ defmodule ArchethicWeb.UserSocket do
 
   require Logger
 
+  alias Archethic
```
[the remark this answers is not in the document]
apoorv-2204: To fix dialyzer issue,

lib/archethic/shared_secrets.ex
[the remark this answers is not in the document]
apoorv-2204: for testing??

lib/archethic/shared_secrets.ex
apoorv-2204: testing??

Makefile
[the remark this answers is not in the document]
samuelmanzanera: Maybe you can override `mix compile` command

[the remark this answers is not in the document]
apoorv-2204: why didnt I thought of that

lib/archethic/shared_secrets.ex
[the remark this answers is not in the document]
samuelmanzanera: As this function is only about testing the two functions, I think the tests cases can handle the different functions.

lib/archethic_web/channels/user_socket.ex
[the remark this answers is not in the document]
apoorv-2204: caused by Archethic.up?

test/archethic_test.exs
```diff
@@ -1,33 +1,158 @@
...
+  alias Message.{NotFound, StartMining, TransactionChainLength, TransactionInputList}
+  alias Message.{NewTransaction}
+
+  alias TransactionChain.{Transaction, TransactionData}
+  alias TransactionChain.{TransactionInput, VersionedTransactionInput}
 
   import Mox
 
+  setup do
+    nss_genesis_address = "nss_genesis_address"
```
bchamagne, line 18: is it possible to add this setup as a named setup in ArchethicCase and use `setup [:nss_mock_setup]` to reference it where you need it?

Neylix (changes requested):

lib/archethic/shared_secrets.ex
```diff
@@ -212,4 +218,76 @@ defmodule Archethic.SharedSecrets do
...
+     Validation time is same as time at which tx was created.
+
+   manual:
+
+   Quorum last address from the network and compare locally
+
+  """
+  @spec is_synchronized?(:node_shared_secrets) ::
+          {:ok, :node_in_sync} | {:error, binary() | :node_out_of_sync}
+  def is_synchronized?(:node_shared_secrets) do
```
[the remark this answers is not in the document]
Neylix: As mentionned, the `is` prefix induce that the function return a boolean (is_synchronized -> true / false). Maybe you can rename it as the P2P function `verify_synchronization`

Neylix (changes requested):

lib/archethic/self_repair.ex
```diff
@@ -217,4 +205,56 @@ defmodule Archethic.SelfRepair do
...
+        )
+
+      pid ->
+        add_repair_addresses(pid, first_address, [])
+    end
+
+    :ok
+  end
+
+  @spec resync_protocol_chain(atom(), list(P2P.Node.t()) | []) :: :ok | :error
```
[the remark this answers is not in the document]
apoorv-2204: its confusing whether we mean network or the arch ethic network.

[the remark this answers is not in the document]
Neylix: In any cas we should be consistent using only network or only protocol, but not both a the same time. But in fact there is "Network Transactions Chains" in the yellow paper so I guess we should keep network

lib/archethic/p2p.ex
```diff
@@ -761,4 +750,11 @@ defmodule Archethic.P2P do
       conflict_resolver.(distinct_elems)
     end
   end
+
+  @doc """
+  Whether a node is synchronised or not is determined by whether the protocol chain node shared secrets are in sync.
+  When a new node renewal tx is produced, it is feasible that a node does not have that tx, resulting in an anomaly in election.
+  """
+  @spec verify_synchronization() :: :ok | :error | {:error, binary()}
+  def verify_synchronization(), do: SharedSecrets.is_synchronized(:node_shared_secrets)
```
samuelmanzanera, line 759: I don't get the purpose of this function because first it's on the P2P layer which handles the P2P communication, second it's just a delegation to SharedSecrets. I suggest in the Archethic module to directly check if the `SharedSecrets.checkSynchronization()` return `ok` before to send the tx, etc...

mix.exs
```diff
@@ -131,6 +131,15 @@ defmodule Archethic.MixProject do
 
   defp aliases do
     [
+      compile: ["git_hooks.install", "compile"],
+      "dev.update_deps": [
```
samuelmanzanera, line 135: when this task will be called ?

apoorv-2204: i added new aliases for updating deps, instead of doing one by one manually

lib/archethic/self_repair.ex
```diff
@@ -217,4 +205,56 @@ defmodule Archethic.SelfRepair do
...
+      true ->
+        Logger.info("Enforced Resync: No new transaction to sync", transaction_type: type)
+        :ok
+
+      e ->
+        Logger.debug("Enforced Resync: Error #{inspect(e)}", transaction_type: type)
+    end
+  end
+
+  defp get_genesis_address(type) do
```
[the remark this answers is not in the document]
apoorv-2204: I want to use case here.it intentional.

lib/archethic/p2p.ex
apoorv-2204: idea wast to ask is this node synchronized? thats why it was in p2p

lib/archethic/self_repair.ex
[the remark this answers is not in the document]
samuelmanzanera: I think the pattern matching improves the readability and eases the understanding.

mix.exs
samuelmanzanera: Yes but we should have a way to avoid outdated deps

Neylix: AN idea could be to let the deps outdated check in the CI but at the end and non blocking.
So outdated PR will not block a commit push, but we will see that there is outdated deps after all commands are done

apoorv-2204: > Yes but we should have a way to avoid outdated deps

we have to do it manually, as it may include feature that break existing code.

lib/archethic/self_repair.ex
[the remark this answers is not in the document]
apoorv-2204: its done

mix.exs
samuelmanzanera: > AN idea could be to let the deps outdated check in the CI but at the end and non blocking. So outdated PR will not block a commit push, but we will see that there is outdated deps after all commands are done

With this approach, I'm afraid we will not look at it, unless the CI is failing. In CI's job, we just want it to pass.

samuelmanzanera (approved):
Code is better now, thanks for the refactor !

Neylix (approved):

PR #901, opened 2023-02-10
#3 Improve tests

apoorv-2204:


test/archethic/beacon_chain/subset_test.exs
```diff
@@ -42,31 +58,18 @@ defmodule Archethic.BeaconChain.SubsetTest do
...
-        {:ok, %Ok{}}
-    end)
-
     assert %{
              current_slot: %Slot{
                end_of_node_synchronizations: [%EndOfNodeSync{public_key: ^public_key}]
              }
            } = :sys.get_state(pid)
+
+    Process.sleep(100)
```
tenmoves, line 67: we should think of another alternative than idling the processes if we can

apoorv-2204: the test closes while the other child processes keep running in need of Mox

tenmoves: can't we send a brutal_kill to these processes ?

apoorv-2204:
``` @spec broadcast_message(list(Node.t()), Message.request()) :: :ok
  def broadcast_message(nodes, message) do
    Task.Supervisor.async_stream_nolink(TaskSupervisor, nodes, &send_message(&1, message),
      ordered: false,
      on_timeout: :kill_task,
      timeout: Message.get_timeout(message) + 2000
    )
    |> Stream.run()
  end```

we need to identify child proceses, which counts this also.

apoorv-2204: > can't we send a brutal_kill to these processes ?

They are difficult to identify, I tried to use Process.list, and then process .info.
Is it possible?
what do you think ?
@tenmoves

bchamagne: FYI, I tried this https://github.com/archethic-foundation/archethic-node/commit/1bebfeab4faca0b14fa310eef77e5cfd9f074ca8 . This is trigger 1 or 2 times every run, but it does not fix much :(

bchamagne (approved):

PR #895, opened 2023-02-07
Smart Contract Interpreter versionning

bchamagne:
# Description

Introduce smart contract versioning. 
The user will be able to choose it's version by setting a `@version` attribute at the top of her/his code. No `@version` attributes is equivalent to 0.0.1.
Part of #894

There are 2 versions available:
1. legacy / functional lang (= 0.0.1)
2. imperative lang (= 1.0.0)

I move current language into a version0 folder. 
The imperative lang is not implemented in this PR but I prepared the ground in a version1 interpreter. 
Some files has been renamed to match their modules names.

## TODO
- [x] Unplug the worker from version0 and plug it to the dispatcher (Archethic.Contracts.Interpreter) instead. 
- [x] Unplug the contracts from version0 and plug it to the dispatcher instead.
- [x] Should I restrict versions to only those existing? 

## Examples

```
@version "1.0.0" 
condition ....
actions ...
```

```
@version "0.0.1"
condition ...
actions ...
```

```
condition ...
actions ...
```

# Type of change

- New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Unit tests :heavy_check_mark: 
Legacy smart contract  :heavy_check_mark: 

# Checklist:

- My code follows the style guidelines of this project
- I have performed a self-review of my own code
- I have commented my code, particularly in hard-to-understand areas
- I have made corresponding changes to the documentation
- My changes generate no new warnings
- I have added tests that prove my fix is effective or that my feature works
- New and existing unit tests pass locally with my changes
- Any dependent changes have been merged and published in downstream modules

PR #903, opened 2023-02-16
Smart Contract Interpreter v1 (mutable bindings/modules/for loop)

### LINKED ISSUE #894: Smart Contract Language: imperative over functional
We'd like our smart contract to be less functional. 
This is a big feature that introduce an entirely new language version (from 0.0.1 to 1.0.0). 

- [x] Smart Contract Language semantic versioning 
- [x] Mutable bindings
- [x] Functions are all part of a module
- [x] For loops 

# Semantic versioning
- Version 0.0.1 will be the legacy code (I call it the functional lang)
- Version 1.0.0 will be the language designed by this issue (I call it the imperative lang)
- Version 1.x.x will be the evolution of the imperative language
- Version 2.x.x and others will be when/if we introduce breaking changes

# Mutable bindings
Users should be able to create variables and modify them.
Variables will be scoped by block.

```
name = "Anonymous"

if true do 
  name = "Roger"
end

# here, name == "Roger"
```

```
if true do 
  name = "Roger"
  ...
end

# here, name is undefined
```

# For loop 
Because we will have mutable bindings, the for loop is the easiest way for user to handle lists.
```
for item in list do 
  ...
end
```

# Functions are grouped by modules
Instead of having all functions available directly, we will put them in modules.

```
Map.put(map, key, value)
Map.get(map, key, default)
Contract.set_content(string)
Contract.set_type(string)
Chain.get_first_address(string) 
Chain.get_genesis_address(string)
...
```

bchamagne:
# Description

I am creating this PR on top of branch #895 . 
Here are the features of this new interpreter: 
1. scoped mutable bindings
2. functions grouped into modules 
3. for loop

I reused as much of the previous interpreter as possible. You'll often see Version0 calls from within Version1. 
This PR is so big, the diff will not help you. To review, just open the /archethic/contracts/interpreter folder and read everything from there (you can skip Version0 folder, it is the previous interpreter)

Fixes #894

## TODO 

- [x] for loop
- [x] documentation

## Type of change

# How Has This Been Tested?

Many unit tests were added. 
Tested manually with the transactionBuilder

# Checklist:

### DIFF
--- a/lib/archethic/contracts/interpreter/version1/action_interpreter.ex
+++ b/lib/archethic/contracts/interpreter/version1/action_interpreter.ex
@@ -174,6 +174,19 @@ defmodule Archethic.Contracts.Interpreter.Version1.ActionInterpreter do
     throw({:error, node, "Forbidden to use if as an expression."})
   end
 
+  # forbid "for" as an expression
+  defp prewalk(
+         node =
+           {:=, _,
+            [
+              {{:atom, _}, _, nil},
+              {{:atom, "for"}, _, _}
+            ]},
+         _acc
+       ) do
+    throw({:error, node, "Forbidden to use for as an expression."})
+  end
+
   # whitelist assignation & write them to scope
   # this is done in the prewalk because it must be done before the "variable are read from scope" step
   defp prewalk(
--- a/lib/archethic/contracts/interpreter/version1/action_interpreter.ex
+++ b/lib/archethic/contracts/interpreter/version1/action_interpreter.ex
@@ -156,6 +156,14 @@ defmodule Archethic.Contracts.Interpreter.Version1.ActionInterpreter do
     {node, acc}
   end
 
+  # forbid "if" as an expression
+  defp prewalk(
+         node = {:=, _, [_, {:if, _, _}]},
+         _acc
+       ) do
+    throw({:error, node, "Forbidden to use if as an expression."})
+  end
+
   # whitelist assignation & write them to scope
   # this is done in the prewalk because it must be done before the "variable are read from scope" step
   defp prewalk(
--- a/lib/archethic/contracts/interpreter/version1/action_interpreter.ex
+++ b/lib/archethic/contracts/interpreter/version1/action_interpreter.ex
@@ -113,16 +113,12 @@ defmodule Archethic.Contracts.Interpreter.Version1.ActionInterpreter do
         ast,
         acc,
         fn node, acc ->
-          # IO.inspect(node, label: "prewalk")
           prewalk(node, acc)
         end,
         fn node, acc ->
-          # IO.inspect(node, label: "postwalk")
           postwalk(node, acc)
         end
       )
-
-    # IO.inspect(new_ast, label: "new_ast", limit: :infinity)
 
     new_ast
   end
--- a/lib/archethic/contracts/interpreter/version1/action_interpreter.ex
+++ b/lib/archethic/contracts/interpreter/version1/action_interpreter.ex
@@ -241,6 +241,42 @@ defmodule Archethic.Contracts.Interpreter.Version1.ActionInterpreter do
     {new_node, new_acc}
   end
 
+  # Map access non-nested (x[y])
+  defp prewalk(
+         _node = {{:., _, [Access, :get]}, _, [{{:atom, map_name}, _, nil}, accessor]},
+         acc
+       ) do
+    # accessor can be a variable, a function call, a dot access, a string
+    new_node =
+      quote do
+        get_in(
+          Process.get(:scope),
+          Scope.where_is(Process.get(:scope), unquote(acc), unquote(map_name)) ++
+            [unquote(map_name), unquote(accessor)]
+        )
+      end
+
+    {new_node, acc}
+  end
+
+  # Map access nested (x[y][z])
+  defp prewalk(
+         _node = {{:., _, [Access, :get]}, _, [first_arg = {{:., _, _}, _, _}, accessor]},
+         acc
+       ) do
+    {nested, new_acc} = prewalk(first_arg, acc)
+
+    new_node =
+      quote do
+        get_in(
+          unquote(nested),
+          [unquote(accessor)]
+        )
+      end
+
+    {new_node, new_acc}
+  end
+
   # for var in list
   defp prewalk(
          _node =
--- a/lib/archethic/contracts/interpreter/version1/action_interpreter.ex
+++ b/lib/archethic/contracts/interpreter/version1/action_interpreter.ex
@@ -317,6 +317,11 @@ defmodule Archethic.Contracts.Interpreter.Version1.ActionInterpreter do
        ) do
     absolute_module_atom = Archethic.Contracts.Interpreter.Version1.Library.Contract
 
+    # check function exists
+    unless Library.function_exists?(absolute_module_atom, function_name) do
+      throw({:error, node, "unknown function: Contract.#{function_name}"})
+    end
+
     # check function is available with given arity
     # (we add 1 to arity because we add the contract as 1st argument implicitely)
     unless Library.function_exists?(absolute_module_atom, function_name, length(args) + 1) do
--- a/lib/archethic/contracts/interpreter/version1/action_interpreter.ex
+++ b/lib/archethic/contracts/interpreter/version1/action_interpreter.ex
@@ -241,21 +241,27 @@ defmodule Archethic.Contracts.Interpreter.Version1.ActionInterpreter do
     {new_node, new_acc}
   end
 
-  # for var: list
+  # for var in list
   defp prewalk(
          _node =
            {{:atom, "for"}, meta,
             [
-              [{{:atom, var_name}, list}],
+              {:in, _,
+               [
+                 {{:atom, var_name}, _, nil},
+                 list
+               ]},
               [do: block]
             ]},
          acc
        ) do
-    # wrap in a block to be able to pattern match it to create a scope
     ast =
       {{:atom, "for"}, meta,
        [
+         # we change the "var in list" to "var: list" (which will be automatically converted to %{var => list})
+         # to avoid the "var" interpreted as a variable (which would have been converted to get_in/2)
          [{{:atom, var_name}, list}],
+         # wrap in a block to be able to pattern match it to create a scope
          [do: AST.wrap_in_block(block)]
        ]}
 
@@ -367,7 +373,7 @@ defmodule Archethic.Contracts.Interpreter.Version1.ActionInterpreter do
     {new_node, acc}
   end
 
-  # for var: list
+  # for var in list
   defp postwalk(
          _node =
            {{:atom, "for"}, _,
--- a/lib/archethic/contracts/interpreter/version1/action_interpreter.ex
+++ b/lib/archethic/contracts/interpreter/version1/action_interpreter.ex
@@ -240,6 +240,11 @@ defmodule Archethic.Contracts.Interpreter.Version1.ActionInterpreter do
 
     functionAtom = String.to_existing_atom(functionName)
 
+    # check the type of the args
+    unless absoluteModuleAtom.check_types(functionAtom, args) do
+      throw({:error, node, "invalid arguments for function Contract.#{functionName}"})
+    end
+
     new_node =
       quote do
         Process.put(
--- a/lib/archethic/contracts/interpreter/version1/action_interpreter.ex
+++ b/lib/archethic/contracts/interpreter/version1/action_interpreter.ex
@@ -5,18 +5,21 @@ defmodule Archethic.Contracts.Interpreter.Version1.ActionInterpreter do
   alias Archethic.TransactionChain.Transaction
   alias Archethic.TransactionChain.TransactionData
 
-  alias Archethic.Contracts.Interpreter.Version1.CommonInterpreter
   alias Archethic.Contracts.Interpreter.ASTHelper, as: AST
+  alias Archethic.Contracts.Interpreter.Version1.CommonInterpreter
+  alias Archethic.Contracts.Interpreter.Version1.Scope
 
   @doc """
   Parse the given node and return the trigger and the actions block.
   """
   @spec parse(Macro.t()) :: {:ok, atom(), Macro.t()} | {:error, Macro.t(), String.t()}
   def parse({{:atom, "actions"}, _, [keyword, [do: block]]}) do
-    # We parse the outer block outside of the macro.traverse
-    # so we can keep a clean acc that is used only for scoping variables.
     trigger_type = extract_trigger(keyword)
+
+    # We only parse the do..end block with the macro.traverse
+    # this help us keep a clean accumulator that is used only for scoping.
     actions_ast = parse_block(AST.wrap_in_block(block))
+
     {:ok, trigger_type, actions_ast}
   catch
     {:error, node} ->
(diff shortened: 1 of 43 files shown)


apoorv-2204:
very complex, 
may be the biggest pr of aenode

bchamagne:
Here's the code proposition to handle the faucet via a smart contract: 

```
@version "1.0.0"
condition inherit: [
	uco_transfers: true,
	content: true
]

# every minutes, send UCOs to the addresses that asked for it
# a chain can do the faucet once every 2 minutes
actions triggered_by: interval, at: "0 * * * *" do
	now = Time.now() # Current time
	faucet_delay = 2 * 60 # 2 min delay
	calls = Contract.get_calls() # amount of calls since last tick

	# work only if there is 1..* call
	if calls != [] do
		# remove the addresses that have waited long enough
		content = ""
		for line: Regex.scan(contract.content, "^(\\w+),(\\d+)$") do
			address = List.take_element_at_index(line, 0)
			at = List.take_element_at_index(line, 1)

			if now - String.to_int(at) <= faucet_delay do
				content = "#{content}#{address},#{at}\n"
			end
		end

		# get the genesis and if allowed, add the transfer
		for call: calls do
			genesis_address = Chain.fetch_genesis_address(call.address)

			if Regex.extract(content, genesis_address) == "" do
				# transfer 5 uco
				Contract.add_uco_transfer(to: genesis_address, amount: 500000000)

				# update state
				content = "#{content}#{genesis_address},#{now}\n"
			end
		end

    # update the state
		Contract.set_content(content)
	end
end
```

Works fine.

bchamagne:
I will wait for feedback before working on documentation

lib/archethic/contracts/interpreter/version1/common_interpreter.ex
```diff
@@ -0,0 +1,168 @@
...
+  def prewalk(node = {{:., _, [{:__aliases__, _, _}, _]}, _, _}, acc), do: {node, acc}
+  def prewalk(node = {:., _, [{:__aliases__, _, _}, _]}, acc), do: {node, acc}
+
+  # whitelisted modules
+  def prewalk(node = {:__aliases__, _, [atom: module_name]}, acc)
+      when module_name in @modules_whitelisted,
+      do: {node, acc}
+
+  # internal modules (Process/Scope/Kernel)
+  def prewalk(node = {:__aliases__, _, [atom]}, acc) when is_atom(atom), do: {node, acc}
```
samuelmanzanera, line 82: maybe we should restrict the atom here to avoid ast like:

```elixir
{{:., [line: 1],
  [
    {:__aliases__, [line: 1], [:Kernel]},
    :spawn
  ]}, [line: 1],
 [
   {:fn, [line: 1],
    [{:->, [line: 1], [[], 10]}]}
 ]}
```

```elixir
{{:., [line: 1],
  [
    {:__aliases__, [line: 1],
     [:Process]},
    :exit
  ]}, [line: 1],
 [{:pid, [line: 1], nil}]}
```

which could be dangerous

lib/archethic/contracts/interpreter/version1/condition_interpreter.ex
```diff
@@ -0,0 +1,152 @@
...
+
+  @doc """
+  Return true if the given conditions are valid on the given constants
+  Here we can use version0 because the validation logic does not change.
+
+  We'll just need to use the same scope as version0 to make it work. (to_boolean_expression does this)
+  """
+  @spec valid_conditions?(Conditions.t(), map()) :: bool()
+  def valid_conditions?(conditions, constants) do
+    Archethic.Contracts.Interpreter.Version0.ConditionInterpreter.valid_conditions?(
```
samuelmanzanera, line 60: Maybe using some alias ;)

test/archethic/contracts/interpreter/version1/action_interpreter_test.exs
```diff
@@ -294,25 +294,25 @@ defmodule Archethic.Contracts.Interpreter.Version1.ActionInterpreterTest do
...
-    #            |> Interpreter.sanitize_code()
-    #            |> elem(1)
-    #            |> ActionInterpreter.parse()
-    # end
+    test "should be able to use loop" do
+      code = ~S"""
+      actions triggered_by: transaction do
+        result = 0
+
+        for i: [1,2,3] do
```
samuelmanzanera, line 318: Shouldn't it be easier to understand or read with: `for i in [1, 2, 3]` ?

lib/archethic/contracts/interpreter/version1/action_interpreter.ex
```diff
@@ -0,0 +1,404 @@
...
+    # initiate a transaction that will be use by the "Contract" module
+    next_tx = %Transaction{data: %TransactionData{}}
+
+    # we use the process dictionary to store our scope
+    # because it is mutable.
+    #
+    # constants should already contains the "magic" variables
+    #   - "contract": current contract transaction
+    #   - "transaction": the incoming transaction (when trigger=transaction)
+    Process.put(
```
samuelmanzanera, line 53: Why do you leverage process directory instead of direct bindings in the `Code.eval_quoted` ?

samuelmanzanera:
Except few comments, really great work 🚀

lib/archethic/contracts/interpreter/version1/common_interpreter.ex
bchamagne: At first I thought the same, but since users cannot create atoms, code is much shorter this way. Should I make the change?

test/archethic/contracts/interpreter/version1/action_interpreter_test.exs
bchamagne: It's actually pretty easy to change, I started with the  `in` and changed for this syntax later. Just tell me and I'll update it

bchamagne:
FYI, there's a way to replace `Process.get/1`and `Process.put/2 ` by using macro hygiene instead. 

`var!(:toto, :context1)`
`var!(:toto, :context2)`

Since they are in a different context, here the 2 `toto` are different and there's no risk of collision. 
https://elixir-lang.org/getting-started/meta/macros.html#macro-hygiene

The issue with this is that the context is an `atom`. Which means we'd have to do some tricky atom manipulation for scopes.

test/archethic/contracts/interpreter/version1/action_interpreter_test.exs
samuelmanzanera: Yes indeed in JS it's more like for ..of but I prefer the syntax and meaning of for in

samuelmanzanera:
> FYI, there's a way to replace `Process.get/1`and `Process.put/2 ` by using macro hygiene instead.
> 
> `var!(:toto, :context1)` `var!(:toto, :context2)`
> 
> Since they are in a different context, here the 2 `toto` are different and there's no risk of collision. https://elixir-lang.org/getting-started/meta/macros.html#macro-hygiene
> 
> The issue with this is that the context is an `atom`. Which means we'd have to do some tricky atom manipulation for scopes.

No worry because the Process directory is by process as long as the execution of a contract is isolated into a single one, no risk of collision.

lib/archethic/contracts/interpreter/version1/common_interpreter.ex
samuelmanzanera: It's ok indeed

lib/archethic/contracts/interpreter/version1/condition_interpreter.ex
```diff
@@ -179,22 +173,20 @@ defmodule Archethic.Contracts.Interpreter.Version1.ConditionInterpreter do
...
-        Archethic.Contracts.Interpreter.Version1.Library.Contract.get_calls(
-          get_in(
-            Process.get(:scope),
-            [unquote(global_variable), "address"]
-          )
-        )
-      end
-
-    {new_node, acc}
+    # new_node =
```
bchamagne, line 181: I disabled `Contract.get_calls()` from the `condition` blocks for now, because it'd require a quorum and not a read from DB. (because the `condition inherit` validator is not the worker).

samuelmanzanera (approved):

Bantarus:
Hello @bchamagne , @Neylix , @samuelmanzanera ,

Bravo for the work on the interpreted language !
I hope it is not too late but here my feedbacks on the language form .

Before i say anything , you can take into account i'm not familiar with writting language interpreter with elixir so take what is coming considering i'm a noobie here.

- I would go with a syntax even more similar to Javascript/typescript ( like replacing "do .. end" bloc with bracket  {} )  since it is the most used language in web, to facilitate the onboarding of web developers.
-  It would be nice to see a structured language like in oriented progamming language ,  could you provide an example if i have multiples actions ?  Can i call an action ( without trigger ) within another action ? I understand that the action is our function paradigm.
- We could use the condition as a rooting feature like in web api  to execute a certain portion/action of my code in my smart contract not only a absolute condition at the smart contract level.

Here an exemple to illustrate my words.
`@version 1

condition inherit {
  uco_transfers: true,
  content: true
}

action processFaucetRequests {
  let now = Time.now() // Current time
  let faucet_delay = 6 * 60 * 60 // 6 hour delay
  let calls = Contract.get_calls() // list of calls since last tick

  // work only if there is at least one call
  if calls != [] {
    // remove the addresses that have waited long enough
    let content = ""
    for line in Regex.scan(contract.content, "^(\\w+),(\\d+)$") {
      let address = List.at(line, 0)
      let at = List.at(line, 1)

      if now - String.to_int(at) <= faucet_delay {
        content = "#{content}#{address},#{at}\n"
      }
    }

    // get the genesis and if allowed, add the transfer
    for call in calls {
      let genesis_address = Chain.get_genesis_address(call.address)

      if Regex.extract(content, genesis_address) == "" {
        // transfer 5 uco
        Contract.add_uco_transfer(to: genesis_address, amount: 5)

        // update state
        content = "#{content}#{genesis_address},#{now}\n"
      }
    }

    // update the state
    Contract.set_content(content)
  }
}

contract FaucetContract {
  triggers {
    interval("0 0 * * *") {
      processFaucetRequests()
    }
  }

  conditions {
    inherit {
      uco_transfers: true,
      content: true
    }
  }

  actions {
    processFaucetRequests
  }
}
`

Also the trigger could be a decorator of an action like this "@trigger_datetime("2023-05-01T00:00:00Z")".
Voila c'est tout ^^.

Bonne journée :)

bchamagne:
Hey @Bantarus , thank you for your feedback, I appreciate. 
Here's my thoughts: 

The language use an intermediary AST which is elixir's. This helped us tremendously because we did not have to create an entire interpreter. Therefore we are limitated by what is considered "valid elixir": we cannot change the `do...end` syntax by curly braces.

I think the named `action` block, without trigger, is a really nice way to have functions in the language. We'll definitely keep that in mind. 

What you propose is very similar to something I drafted a few months ago:
 
```
contract([
	condition_inherit: fn out_tx ->
		 assert Regex.match?(out_tx.content, ".*")
	end,
	triggers: [
		[
			trigger: "0 0 0 * * * *",
			code: fn contract ->
				Contract.set_content "hello"
			end,
			condition: []
	    ],
		[
			trigger: "transaction",
			conditions: [
				transaction: fn in_tx ->
					assert List.first(in_tx.uco_transfers).amount > 0
				end
			],
			code: fn contract, in_tx ->
				Contract.set_type "transfer"
				Contract.add_uco_transfer to: "000030831178cd6a49fe446778455a7a980729a293bfa16b0a1d2743935db210da76", amount: 1337 
			end
	    ]	
	]
])
```
Anyway, it's too big of a breaking change for now. Maybe in a later version?
Cheers!

samuelmanzanera:
> Hello @bchamagne , @Neylix , @samuelmanzanera ,
> 
> Bravo for the work on the interpreted language ! I hope it is not too late but here my feedbacks on the language form .
> 
> Before i say anything , you can take into account i'm not familiar with writting language interpreter with elixir so take what is coming considering i'm a noobie here.
> 
> * I would go with a syntax even more similar to Javascript/typescript ( like replacing "do .. end" bloc with bracket  {} )  since it is the most used language in web, to facilitate the onboarding of web developers.
> * It would be nice to see a structured language like in oriented progamming language ,  could you provide an example if i have multiples actions ?  Can i call an action ( without trigger ) within another action ? I understand that the action is our function paradigm.
> * We could use the condition as a rooting feature like in web api  to execute a certain portion/action of the code in the smart contract, not only an absolute condition at the smart contract level.
> 
> Here an exemple to illustrate my words.
> 
> ```
> @version 1
> 
> action processFaucetRequests {
>   let now = Time.now() // Current time
>   let faucet_delay = 6 * 60 * 60 // 6 hour delay
>   let calls = Contract.get_calls() // list of calls since last tick
> 
>   // work only if there is at least one call
>   if calls != [] {
>     // remove the addresses that have waited long enough
>     let content = ""
>     for line in Regex.scan(contract.content, "^(\\w+),(\\d+)$") {
>       let address = List.at(line, 0)
>       let at = List.at(line, 1)
> 
>       if now - String.to_int(at) <= faucet_delay {
>         content = "#{content}#{address},#{at}\n"
>       }
>     }
> 
>     // get the genesis and if allowed, add the transfer
>     for call in calls {
>       let genesis_address = Chain.get_genesis_address(call.address)
> 
>       if Regex.extract(content, genesis_address) == "" {
>         // transfer 5 uco
>         Contract.add_uco_transfer(to: genesis_address, amount: 5)
> 
>         // update state
>         content = "#{content}#{genesis_address},#{now}\n"
>       }
>     }
> 
>     // update the state
>     Contract.set_content(content)
>   }
> }
> 
> contract FaucetContract {
>   triggers {
>     interval("0 0 * * *") {
>       processFaucetRequests()
>     }
>   }
> 
>   actions {
>     processFaucetRequests {
>       condition inherit {
>         uco_transfers: true,
>         content: true
>       }
>     }
>   }
> }
> ```
> 
> Also the trigger could be a decorator of an action like this "`@trigger_datetime("2023-05-01T00:00:00Z")`". Voila c'est tout ^^.
> 
> Bonne journée :)

Hello. Thanks for the contribution.
I just wanted to mention some motivation about this syntax and the smart contract language regarding the points you mentioned. Hopefully this would help to understand the choices we made.

 > I would go with a syntax even more similar to Javascript/typescript ( like replacing "do .. end" bloc with bracket {} ) since it is the most used language in web, to facilitate the onboarding of web developers.
As mentioned @bchamagne, we are using Elixir to build the interpreted language, hence it's a subset of the Elixir language. But even without it, we designed the language to be as a domain specific language, close to a specification, while allowing developers to leverage a programmatic approach. In that direction, the language is then designed to be easy to read and understand for non-developers if needed.
Hence, the `do...end` statement helps to understand what happens inside a block like an algorithm handwritten and can be "read". Another assumption is because the Elixir syntax sugar (do..end) is very close to Ruby which is one of the most used language worldwide and being known to be simple to learn and read.

> It would be nice to see a structured language like in oriented progamming language , could you provide an example if i have multiples actions ? Can i call an action ( without trigger ) within another action ? I understand that the action is our function paradigm.
The purpose of this language is not to be a generic programming language but a dedicated one for the Archethic's interpreter. 
The action for me is not really a function paradigm but more a trigger code. The reason is to avoid some security flows known in Solidity being a Turing complete language. Since the beginning, we didn't want to make a Turing complete language. We have to see the action block as a code which is triggered by an even in the network, like a serverless function/lambda like the AWS Cloud.
However, we are still thinking of a way to branch different actions block based on a name. Hence, the clients and dApps could target a specific action block like a function name.

Bantarus:
Hello @samuelmanzanera , 

Thanks you for the detailed answer. 
C'est clair comme de l'eau de roche :)

Keep up the incredible work !

PR #923, opened 2023-02-28
Fix slot missing transaction

Neylix:
# Description

Actually, when it's the time for a new slot, the slot timer send a event to all the subset to process the slot creation.
But due to some latency for a slot node to receive the ReplicationAttestation message or simply the latency for a transaction to be replicated, the slot node may have build and sent the slot before receiving all the slot transaction.

For example:
image

To solve this problem, here is the new workflow: 
- We don't verify that a replication attestation is in the right slot time. Any attestation can be in any slot.
- When building the summary, we control that the number of confirmations of an attestation reach a certain threshold (35%) of the expected confirmations number. If it not reaches the threshold, the attestation is postponed to the next summary.
- As we can have postponed attestation in a next summary, we ensure the attestation is between 2 previous summary and current summary.

Here is an example:

image

## Type of change

# Checklist:

### DIFF
--- a/lib/archethic/beacon_chain/replication_attestation.ex
+++ b/lib/archethic/beacon_chain/replication_attestation.ex
@@ -209,6 +209,37 @@ defmodule Archethic.BeaconChain.ReplicationAttestation do
     end
   end
 
+  @doc """
+  Take a list of attestations and reduce them to return a list of unique attestation
+  for a transaction with all the confirmations
+  """
+  @spec reduce_confirmations(Enumerable.t(t())) :: Stream.t(t())
+  def reduce_confirmations(attestations) do
+    attestations
+    |> Stream.transform(
+      # start function, init acc
+      fn -> %{} end,
+      # reducer function, return empty enum, accumulate replication attestation by address in acc
+      fn attestation = %__MODULE__{
+           transaction_summary: %TransactionSummary{address: address},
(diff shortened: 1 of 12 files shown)


Neylix:
After some discussion, a better way to handle this problem has been found: 
When a node receive a replication attestation it will add it to the current slot if the tx timestamp is over than 2 previous summary time even if the slot time is not the right one.
When it is a slot time, every node will send the slot if it is not empty event if the node is not an elected node for the current slot.
On the summary time, before adding a replication attestation, the node ensure that there is enough confirmations with a threshold of 35% minimum (threshold if there is at least 10 expected replication nodes) If the threshold is not reached, the attestation is postponed to the next summary by just adding the attestation in the current slot.
During self repair, the same threshold control is performmed

bchamagne:
As discussed, there is an issue with the transaction from last slot being duplicated in 2 summaries (S and S+1).

bchamagne (approved):
Tested with 3 nodes (minimum_nodes_for_threshold = 2)
I added a 10s sleep to 2/3 nodes.
I created a transfer few second before summary 
Transfer postponed as expected. :heavy_check_mark: 

Screenshot 2023-03-07 at 11-46-16 ARCHETHIC

samuelmanzanera (approved):

PR #927, opened 2023-03-08
Set network coordinates from BeaconChain

### LINKED ISSUE #72: Fetch other beacon summary to build network coordinates map
BeaconChain must implement the consolidation of the network coordinates by fetching other beacon summaries to get the whole map of the network and determine the network coordinates patch.

This feature should also add a new entry in the beacon summary data to hold the network patch of the nodes involved in the beacon chain's subset.

samuelmanzanera:
# Description

This PR implement the full integration of the network patch in the beacon by building the latency matrix from all the subsets to determine the global network patch.

Fixes #72 

## Type of change

# How Has This Been Tested?

After running two nodes, the default network patch (being the geo patch) should change based on the latency between the nodes and be persisted into the memory table for the P2P view.

# Checklist:


lib/archethic/beacon_chain/network_coordinates.ex
```diff
@@ -193,4 +195,139 @@ defmodule Archethic.BeaconChain.NetworkCoordinates do
...
+        stats =
+          Task.Supervisor.async_stream_nolink(
+            TaskSupervisor,
+            beacon_nodes,
+            &P2P.send_message(&1, %GetNetworkStats{subset: subset}),
+            on_timeout: :kill_task,
+            ordered: false
+          )
+          |> Stream.filter(fn
+            # Filter slots without P2P view
```
Neylix, line 228: typo, should be "filter slot **with** P2P view"

lib/archethic/beacon_chain/summary.ex
```diff
@@ -30,7 +30,8 @@ defmodule Archethic.BeaconChain.Summary do
     node_availabilities: <<>>,
     node_average_availabilities: [],
     end_of_node_synchronizations: [],
-    version: 1
+    network_patches: [],
+    version: 2
```
Neylix, line 34: Isn't it better to use a migration task to transform old beacon summaries into the new one with an empty network_patched ?

lib/archethic/beacon_chain/summary_aggregate.ex
```diff
@@ -8,19 +8,21 @@ defmodule Archethic.BeaconChain.SummaryAggregate do
   defstruct [
     :summary_time,
     availability_adding_time: [],
-    version: 1,
+    version: 2,
```
Neylix, line 11: Same here

Neylix (changes requested):
Complex to understand as there is a lot of math and sorted list with index ...

But I can see something that may goes wrong, when the subset enter in `handle_summary` function, it request the network stats of other nodes. Then when stats are retrieved the `pop_slots` function is called on `SummaryCache` which empty the ets table and remove the backup.
But when a node request network stats, the result is streamed from the ets table of `SummaryCache` but if the requested node quickly requested network patches and pop the slots, it will not be able to respond to the `GetNetworkStats` message

lib/archethic/beacon_chain.ex
```diff
@@ -413,4 +414,103 @@ defmodule Archethic.BeaconChain do
...
+  end
+
+  defp clean_outliers(list) do
+    list_size = Enum.count(list)
+
+    # Compute percentiles (P80, P20) to remove the outliers
+    p1 = (0.8 * list_size) |> trunc()
+    p2 = (0.2 * list_size) |> trunc()
+
+    max = Enum.at(list, p1)
```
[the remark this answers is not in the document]
samuelmanzanera: The list is assumed to already sorted by node

lib/archethic/beacon_chain/summary.ex
samuelmanzanera: Maybe 🤔

lib/archethic/beacon_chain.ex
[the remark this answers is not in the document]
Neylix: Yes because actually if the latency you receive is [100, 140, 110] (as in the example) the percentile will be 100 and 110 while it should be 100 and 140.
So you should sort the list to have [100, 110, 140] and then remove the outliers.
And it will make the test value change, actually there is wring expected value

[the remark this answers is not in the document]
samuelmanzanera: Yes it's the expected behavior for percentile. I missed this one.

Neylix (changes requested):

lib/archethic/beacon_chain/subset/summary_cache.ex
```diff
@@ -29,16 +33,54 @@ defmodule Archethic.BeaconChain.Subset.SummaryCache do
...
+        |> Enum.any?(fn
+          {%Slot{slot_time: slot_time}, _} ->
+            DateTime.diff(slot_time, previous_summary_time) >= 0
+
+          %Slot{slot_time: slot_time} ->
+            DateTime.diff(slot_time, previous_summary_time) >= 0
+        end)
+
+      # Clean the slots if there is previous cache information
+      unless contains_next_summary_slots?, do: clean_slots(subset)
```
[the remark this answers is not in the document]
samuelmanzanera: I know it can be a way. This implemention was taken because the slot time event will come before the creation and receiving of the slot from the network.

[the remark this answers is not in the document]
samuelmanzanera: Finally the match spec didn't work with DateTime struct. I figured out another way in 9c8fb4a

lib/archethic/beacon_chain/subset/summary_cache.ex
```diff
@@ -27,18 +31,54 @@ defmodule Archethic.BeaconChain.Subset.SummaryCache do
...
+    previous_summary_time = SummaryTimer.previous_summary(slot_time)
+    first_slot_time = SlotTimer.next_slot(previous_summary_time)
+
+    if slot_time == first_slot_time do
+      Enum.each(
+        BeaconChain.list_subsets(),
+        &clean_previous_summary_cache(&1, previous_summary_time)
+      )
+
+      File.rm(recover_path(previous_summary_time))
```
Neylix, line 52: You may recreate this file with the current slot if there is one in the cache after removing old one

Neylix (changes requested):

Neylix:
I got this error after some time : 
image

A patch has only 2 digits

lib/archethic/beacon_chain/subset/summary_cache.ex
```diff
@@ -27,18 +31,60 @@ defmodule Archethic.BeaconChain.Subset.SummaryCache do
...
+    end
+
+    {:noreply, state}
+  end
+
+  def code_change("1.0.7", state, _extra) do
```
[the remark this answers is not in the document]
samuelmanzanera: Yes right

lib/archethic/p2p/message/get_network_stats.ex
```diff
@@ -0,0 +1,79 @@
...
+  @spec process(t(), Crypto.key()) :: NetworkStats.t()
+  def process(%__MODULE__{subsets: subsets}, _node_public_key) do
+    stats =
+      subsets
+      |> Task.async_stream(fn subset ->
+        stats = BeaconChain.get_network_stats(subset)
+        {subset, stats}
+      end)
+      |> Stream.map(fn {:ok, res} -> res end)
+      |> Enum.reduce(%{}, fn
```
bchamagne, line 69: this looks like a `Enum.into(%{})`

samuelmanzanera: Except we don't want a subset with empty stats. But we can probably do a `Stream.filter` for this purpose.

Neylix (changes requested):
It looks like the problem of the pathc with only 2 digits is still there

lib/archethic/beacon_chain/network_coordinates.ex
```diff
@@ -193,4 +201,276 @@ defmodule Archethic.BeaconChain.NetworkCoordinates do
...
+    end)
+    |> Stream.map(fn {:ok, {:ok, %NetworkStats{stats: stats}}} -> stats end)
+  end
+
+  defp aggregate_stats_per_subset(stats) do
+    stats
+    |> Enum.flat_map(& &1)
+    |> Enum.reduce(%{}, fn {subset, stats}, acc ->
+      Enum.reduce(stats, acc, fn {node, stats}, acc ->
+        Map.update(acc, subset, %{node => stats}, &Map.put(&1, node, stats))
```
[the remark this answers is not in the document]
samuelmanzanera: Yes indeed. We should aggregate maybe using some mean of the latencies retrieved.

lib/archethic/beacon_chain/network_coordinates.ex
```diff
@@ -193,4 +201,276 @@ defmodule Archethic.BeaconChain.NetworkCoordinates do
...
+      end)
+    end)
+  end
+
+  defp update_matrix_from_stats(stats_by_subset, matrix, sorted_node_list) do
+    Enum.reduce(stats_by_subset, matrix, fn {subset, stats}, acc ->
+      sampling_nodes = P2PSampling.list_nodes_to_sample(subset)
+
+      Enum.reduce(stats, acc, fn {node_public_key, stats}, acc ->
+        beacon_node_index =
```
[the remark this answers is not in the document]
Neylix: Nevermind it's a beacon node index

lib/archethic/beacon_chain/subset/summary_cache.ex
```diff
@@ -27,18 +31,62 @@ defmodule Archethic.BeaconChain.Subset.SummaryCache do
...
+    subset
+    |> stream_current_slots()
+    |> Stream.filter(fn
+      {%Slot{slot_time: slot_time}, _} ->
+        DateTime.compare(slot_time, previous_summary_time) == :lt
+
+      %Slot{slot_time: slot_time} ->
+        DateTime.compare(slot_time, previous_summary_time) == :lt
+    end)
+    |> Stream.map(fn item ->
```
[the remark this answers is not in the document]
samuelmanzanera: Yes

lib/archethic/beacon_chain/subset.ex
```diff
@@ -354,7 +356,13 @@ defmodule Archethic.BeaconChain.Subset do
   end
 
   defp handle_summary(time, subset) do
-    beacon_slots = SummaryCache.pop_slots(subset)
+    beacon_slots =
+      subset
+      |> SummaryCache.stream_current_slots()
```
Neylix, line 361: This function is now returning a Stream, so the next Logger does not write what it should

Neylix (commented):
Working well now, I'm just thinking we should create an issue or maybe do it in this one, to add the network patch in the beaconChainSummary graphql API. (in the transform function)

lib/archethic/beacon_chain/summary_aggregate.ex
```diff
@@ -188,6 +199,62 @@ defmodule Archethic.BeaconChain.SummaryAggregate do
...
+      |> Enum.map(fn d ->
+        d
+        |> Enum.take(2)
+        |> Enum.map(&String.to_integer(&1, 16))
+      end)
+      |> Enum.zip()
+      |> Enum.map(fn x ->
+        x
+        |> Tuple.to_list()
+        |> Utils.mean()
```
Neylix, line 237: I think here we should use a median instead of a mean.
Let's say patches are `["F__", "F__", "F__", "0__", "0__"]`
in fact, "0" and "F" are pretty close on the globe, but doing a mean of this will result in (16+16+16+0+0) / 5 = 9.6 = "9"
But "9" is far from "0" or "F" while the result of the aggregation should be "F" or "0". So I guess a median would be better as it will return "F".

lib/archethic/beacon_chain/summary_aggregate.ex
```diff
@@ -188,6 +199,62 @@ defmodule Archethic.BeaconChain.SummaryAggregate do
...
+        |> trunc()
+        |> Integer.to_string(16)
+      end)
+      |> IO.inspect()
```
Neylix, line 241: Remaining inspect()

lib/archethic/beacon_chain/summary_aggregate.ex
```diff
@@ -188,6 +199,62 @@ defmodule Archethic.BeaconChain.SummaryAggregate do
...
+      |> Enum.join()
+
+    bandwidth_patch =
+      splitted_patches
+      |> Enum.map(fn digits ->
+        digits
+        |> List.last()
+        |> String.to_integer(16)
+      end)
+      |> Utils.mean()
```
Neylix, line 251: Same here for median over mean

Neylix (changes requested):

lib/archethic/beacon_chain/summary_aggregate.ex
samuelmanzanera: Why not

samuelmanzanera: Why not

Neylix: But we should take care about the the edge case `["F__", "0__"]` a classical median will do the mean of bot values are there is an even number of values. In this case the median will return "8" which is a wrong value.
But in other way `["2__", "4__"]` the median will return "3__" which is the expected value.
maybe using 0..16 value to represent digits is a wrong way because 0..F digits is are cycle value

lib/archethic/p2p.ex
```diff
@@ -152,11 +164,27 @@ defmodule Archethic.P2P do
...
   """
-  @spec authorized_and_available_node?(Crypto.key()) :: boolean()
-  def authorized_and_available_node?(node_public_key \\ Crypto.first_node_public_key()) do
-    Utils.key_in_node_list?(authorized_and_available_nodes(), node_public_key)
+  @spec authorized_and_available_node?(
+          Crypto.key(),
+          datetime :: DateTime.t(),
+          before? :: boolean()
+        ) :: boolean()
+  def authorized_and_available_node?(
```
Neylix, line 174: Here there is an edge case you don't handle, it's when the first node bootstrap. At this state no nodes are really authorized nor available. So the first node cannot send any transaction until the first the next authorization date.
The previous function was returning the first enrolled node in case no nodes are available and authorized (see `authorized_and_available_nodes` function)

Neylix (changes requested):

lib/archethic/beacon_chain/summary_aggregate.ex
samuelmanzanera: Algorithm are heuristic so they should not be perfect for edge case,. It's just a approximation thought at specific time which was the best approach for normal cases. Of course we can improve later, but I would prefer to ship and see, otherwise it will always be a theory in our mind.

Also for the case when there are only two, I think it's fine to have a approximation , as it might be in bootstrapping mode before 3 nodes views.  The next patch will probably make it  change again.

lib/archethic/beacon_chain/subset/summary_cache.ex
```diff
@@ -27,18 +31,62 @@ defmodule Archethic.BeaconChain.Subset.SummaryCache do
...
+
+      File.rm(recover_path(previous_summary_time))
+    end
+
+    {:noreply, state}
+  end
+
+  def code_change("1.0.7", state, _extra) do
+    next_summary_time = SummaryTimer.next_summary(DateTime.utc_now())
+    File.rename("slot_backup", "slot_backup-#{DateTime.to_unix(next_summary_time)}")
```
bchamagne, line 60: I think `PubSub.register_to_current_epoch_of_slot_time()` added in the init should also be in the code change.

Neylix (approved):

lib/archethic/beacon_chain/subset/summary_cache.ex
Neylix: Yes you're right !

samuelmanzanera: Thanks

Issue #907, opened 2023-02-20
Add pending validation control on replication nodes


Neylix:
### Is your feature request related to a problem?

The module `PendingTransactionValidation` provide a lot of controls to determine if the transaction is allowed to be proceceed or not (network chain control mainnly, transaction type validation ...)
These controls are really important, but they are done only by the validation nodes.

### Describe the solution you'd like

Actually if the replication nodes receive the message `ValidateTransaction` it means that the validation are agreed there is not pending error. But only validation nodes is too few to be sure there is no pending error, so when the replication node validate the transaction (`TransactionValidator.validate()`) it should also run the pending validation controls.

### Additional context

These controls should only be verified during the normal replication. If the replication comes from the self repair, this control can be bypassed

Epic: #212

PR #928, opened 2023-03-09
Add pending validation control on replication nodes #907

apoorv-2204:
# Description

Fixes 

- [ ] #907

## Type of change

# How Has This Been Tested?

- Test A
- Test B

# Checklist:

### DIFF
--- a/lib/archethic/bootstrap/network_init.ex
+++ b/lib/archethic/bootstrap/network_init.ex
@@ -4,37 +4,13 @@ defmodule Archethic.Bootstrap.NetworkInit do
 
   Those functions are only executed by the first node bootstrapping on the network
   """
+  alias Archethic.{Bootstrap, Crypto, Election, Mining, PubSub, Replication}
+  alias Archethic.{Replication, BeaconChain.ReplicationAttestation, SharedSecrets, Reward}
+  alias Archethic.TransactionChain.{Transaction, TransactionData, TransactionSummary}
 
-  alias Archethic.Bootstrap
-
-  alias Archethic.BeaconChain.ReplicationAttestation
-
-  alias Archethic.Crypto
-
-  alias Archethic.Election
-
-  alias Archethic.Mining
-
-  alias Archethic.PubSub
-
-  alias Archethic.Replication
-
-  alias Archethic.SharedSecrets
-
-  alias Archethic.Reward
-
-  alias Archethic.TransactionChain.Transaction
-  alias Archethic.TransactionChain.Transaction.CrossValidationStamp
-  alias Archethic.TransactionChain.Transaction.ValidationStamp
-  alias Archethic.TransactionChain.Transaction.ValidationStamp.LedgerOperations
-  alias Archethic.TransactionChain.Transaction.ValidationStamp.LedgerOperations.UnspentOutput
-
-  alias Archethic.TransactionChain.TransactionData
-  alias Archethic.TransactionChain.TransactionData.Ledger
-  alias Archethic.TransactionChain.TransactionData.UCOLedger
-  alias Archethic.TransactionChain.TransactionData.UCOLedger.Transfer
-
-  alias Archethic.TransactionChain.TransactionSummary
+  alias Transaction.{CrossValidationStamp, ValidationStamp, LedgerOperations}
+  alias ValidationStamp.{LedgerOperations, LedgerOperations.UnspentOutput}
+  alias TransactionData.{Ledger, UCOLedger, UCOLedger.Transfer}
 
   require Logger
 
@@ -225,5 +201,7 @@ defmodule Archethic.Bootstrap.NetworkInit do
     }
 
     PubSub.notify_replication_attestation(attestation)
+
+    :ok
   end
 end
--- a/lib/archethic/mining/pending_transaction_validation.ex
+++ b/lib/archethic/mining/pending_transaction_validation.ex
@@ -1,12 +1,12 @@
 defmodule Archethic.Mining.PendingTransactionValidation do
   @moduledoc false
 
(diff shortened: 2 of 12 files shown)


lib/archethic/p2p/message/replication_error.ex
```diff
@@ -104,8 +107,17 @@ defmodule Archethic.P2P.Message.ReplicationError do
     {%__MODULE__{address: address, reason: reason}, rest}
   end
 
-  @spec deserialize_reason(bin :: bitstring) :: {atom, bitstring}
-  def deserialize_reason(<<nb::8, rest::bitstring>>), do: {error(nb), rest}
+  @spec deserialize_reason(bin :: bitstring) :: {atom, bitstring} | {{atom, bitstring}, bitstring}
+  def deserialize_reason(<<nb::8, rest::bitstring>>) do
```
samuelmanzanera, line 111: You can use a pattern matching here to avoid the case

lib/archethic/replication/transaction_validator.ex
```diff
@@ -63,12 +64,25 @@ defmodule Archethic.Replication.TransactionValidator do
...
       ) do
-    with :ok <- valid_transaction(tx, inputs, true, self_repair?),
+    with :ok <- validate_pending_transaction(tx, self_repair?),
+         :ok <- valid_transaction(tx, inputs, true, self_repair?),
          :ok <- validate_inheritance(tx, previous_transaction, self_repair?) do
       validate_chain(tx, previous_transaction)
     end
   end
 
+  def validate_pending_transaction(_, true), do: :ok
```
samuelmanzanera, line 74: What is the meaning of this true/false ?

samuelmanzanera, line 74: You could use a private function here.

lib/archethic/self_repair.ex
```diff
@@ -234,7 +234,7 @@ defmodule Archethic.SelfRepair do
          {local_last_addr, _} <- TransactionChain.get_last_address(addr),
          false <- rem_last_addr == local_last_addr,
          {:ok, tx} <- TransactionChain.fetch_transaction_remotely(rem_last_addr, nodes),
-         :ok <- Replication.validate_and_store_transaction_chain(tx) do
+         :ok <- Replication.validate_and_store_transaction_chain(tx, false) do
```
samuelmanzanera, line 237: Why false ? Shouldn't it true in the self repair ?

lib/archethic/bootstrap/network_init.ex
```diff
@@ -213,7 +189,7 @@ defmodule Archethic.Bootstrap.NetworkInit do
 
   @spec self_replication(Transaction.t()) :: :ok
   def self_replication(tx = %Transaction{}) do
-    :ok = Replication.validate_and_store_transaction_chain(tx)
+    :ok = Replication.validate_and_store_transaction_chain(tx, true)
```
samuelmanzanera, line 192: I think to avoid conflicts with others PR, you can use `Replication.sync_transaction_chain` as the flag will be removed.

lib/archethic/replication/transaction_validator.ex
```diff
@@ -63,12 +64,25 @@ defmodule Archethic.Replication.TransactionValidator do
         inputs,
         self_repair?
       ) do
-    with :ok <- valid_transaction(tx, inputs, true, self_repair?),
+    with :ok <- validate_pending_transaction(tx, self_repair?),
```
samuelmanzanera, line 67: As the flag will be removed, I think you can omit the self-repair condition in this function.

samuelmanzanera:
Good work anyway ;)

Neylix (changes requested):
When executing the validate pending transaction in the replication module, we should not verify if the transaction exsits.
First because this check is already done before starting the `TransactionValidator.valide`, and this is causing an error for the network transactions on IO nodes.
Abnd for the network transaction the IO nodes receive a message `ReplicateTransactionChain` and they start the function `Replication.validate_and_store_transaction_chain` But as they receive this message after the storage nodes stored this transaction, they raise a transaction already exists error and they do not store the transaction.

lib/archethic/mining/pending_transaction_validation.ex
```diff
@@ -87,7 +87,23 @@ defmodule Archethic.Mining.PendingTransactionValidation do
     end
   end
 
-  defp valid_not_exists(%Transaction{address: address}) do
+  defp valid_not_exists(tx = %Transaction{address: address, type: tx_type}) do
+    case Transaction.network_type?(tx_type) do
```
Neylix, line 91: You can reduce the code to something like
```elixir
with true <- Transaction.network_type?(tx_type),
     true <- TransactionChain.transaction_exists?(address) do
  {:error, "Transaction already exists"}
else
  false ->
    do_valid_not_exists(tx)
end
```

lib/archethic/replication/transaction_validator.ex
Neylix: Yes the self repair flag is not used anymore here

lib/archethic/replication/transaction_validator.ex
```diff
@@ -53,16 +54,35 @@ defmodule Archethic.Replication.TransactionValidator do
...
+          self_repair? :: boolean()
         ) ::
           :ok | {:error, error()}
-  def validate(tx = %Transaction{}, previous_transaction, inputs) do
-    with :ok <- valid_transaction(tx, inputs, true),
+  def validate(
+        tx = %Transaction{},
+        previous_transaction,
+        inputs,
+        self_repair? \\ false
```
Neylix, line 65: You should remove the self repaire flag from this function

Neylix (changes requested):

apoorv-2204:
is any changes required in this pr?

Issue #887, opened 2023-02-05
Improve synchronization when node ip changes


Neylix:
### Describe the problem you discovered

- Actually there is a scheduler that runs every 5 minutes to open port and verify if node ip is changed.
If node ip changes, it create a new node transaction with the new ip and send it to the network.
The problem is that since the node has changed it's ip, other nodes cannot send it request. So the replication of the node transaction will never be sent to the node which changed it's ip. So this node will not load update the last public key of itself and it will result in invalid election error until the next self repair. (Last public key is used for validation node election).

- Other issue is that this node didn't receive transaction during the time between it changes it's ip and the new transaction was replicated by other nodes. So this node may missed some network transaction, (shared secret, nodes, oracle ...)

### Describe the solution you'd like

- When a node create a transaction and send it, it should ask the transaction summary to other nodes and replicate this transaction if the summary exists. Same as in bootstrap transaction node (bootstrap/transaction_handler), but replicate the transaction.

- After the new node transaction is validated (summary retrieved and replication done) the node should resync the network transactions, as it is done in post_bootstrap (function `resync_network_chain`), but it should also include node transaction (for both, post_bootstrap and in network scheduler)

samuelmanzanera:
For the 1st problem, indeed we can wait confirmations of the tx, and request replication.

For the 2nd problem, we might think of a more global solution (not only when the node's IP change). For example, during transaction validation check if the previous transaction exists (in the case of network tx), otherwise download it from authorized nodes, to be able to successfully validate the tx in the latest state.
The replication handles the repair of the sync of network tx (if the last is missing)

PR #930, opened 2023-03-13
improve_synchronization_when_node_ip_changes_#887

apoorv-2204:
# Description

If a node's IP address changes, a new transaction (TX) is generated, but an unusual problem arises: the node becomes invisible to other nodes due to the IP change. While the node with the new IP can create a node TX, it must wait for replication and load the last node public key, which fails because it cannot access the TX it created. The solution involves manually waiting for confirmation and initiating a self-repair process.

Fixes 

- [ ] #887

## Type of change

# Checklist:

### DIFF
--- a/lib/archethic/networking/scheduler.ex
+++ b/lib/archethic/networking/scheduler.ex
@@ -4,25 +4,13 @@ defmodule Archethic.Networking.Scheduler do
   use GenServer
   @vsn Mix.Project.config()[:version]
 
-  alias Archethic.Crypto
+  alias Archethic.{Crypto, P2P, P2P.Node, Networking, TaskSupervisor, Utils, PubSub}
+  alias Archethic.{SelfRepair, TransactionChain}
 
-  alias Archethic.P2P
-  alias Archethic.P2P.Listener, as: P2PListener
-  alias Archethic.P2P.Node
+  alias Networking.{IPLookup, PortForwarding}
+  alias TransactionChain.{Transaction, TransactionData}
 
-  alias Archethic.Networking.IPLookup
-  alias Archethic.Networking.PortForwarding
-
-  alias Archethic.TaskSupervisor
-
-  alias Archethic.TransactionChain
-  alias Archethic.TransactionChain.Transaction
-  alias Archethic.TransactionChain.TransactionData
-
-  alias Archethic.Utils
-
-  alias Archethic.PubSub
-
+  alias P2P.Listener, as: P2PListener
   alias ArchethicWeb.Endpoint, as: WebEndpoint
 
   require Logger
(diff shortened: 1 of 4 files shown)


samuelmanzanera (approved):

lib/archethic/networking/scheduler.ex
```diff
@@ -139,4 +130,19 @@ defmodule Archethic.Networking.Scheduler do
       {:ok, p2p_port, web_port}
     end
   end
+
+  defp handle_new_ip(%Transaction{address: tx_address}) do
+    nodes =
+      P2P.authorized_and_available_nodes()
+      |> Enum.filter(&Node.locally_available?/1)
```
Neylix, line 137: You can sort the node list by nearest node, so the await confirmations would be faster

Neylix (changes requested):

lib/archethic/utils.ex
```diff
@@ -997,4 +993,27 @@ defmodule Archethic.Utils do
...
+          :ok | {:error, :network_issue}
+  def await_confirmation(tx_address, [node | rest]) do
+    case P2P.send_message(node, %GetTransactionSummary{address: tx_address}) do
+      {:ok,
+       %TransactionSummaryMessage{transaction_summary: %TransactionSummary{address: ^tx_address}}} ->
+        :ok
+
+      {:ok, %NotFound{}} ->
+        Process.sleep(200)
+        await_confirmation(tx_address, [node | rest])
```
bchamagne, line 1007: as discussed, infinite loop to remove here

bchamagne (changes requested):

lib/archethic/utils.ex
samuelmanzanera: After discussion with @tenmoves, we should not remove it right away.  
This is still a required "infinite loop". But we can make it "finite". 
This is because the time for the 1st node to replicate the transaction needs be awaiting. However we should use some backoff method with cap to know if we need to switch node in the list (recursive function).

apoorv-2204: We can use Task.async_nolink with Task. yieldall

Neylix: > No sure because it depends on the time of replication. Even if you use a quorum, we would have to retry after some time. Like on the bootstrap.

Yes, but instead to ask to only one node we request to multiple nodes, so if the first requested node is out of sync it's not a problem.
But of course we will need the retry, but no need to manage the nodes to request

apoorv-2204: we can use Task .Superviosr.astream_async_nolink
and Task.await_many

apoorv-2204: when we reaches a successful threshold , depending upon nb of nodes

lib/archethic/networking/scheduler.ex
```diff
@@ -139,4 +130,20 @@ defmodule Archethic.Networking.Scheduler do
...
+      |> P2P.nearest_nodes()
+
+    Utils.await_confirmation(tx_address, nodes)
+
+    types = [:node, :oracle, :node_shared_secrets, :reward]
+
+    Task.Supervisor.async_stream_nolink(Archethic.TaskSupervisor, types, fn type ->
```
bchamagne, line 144: After https://github.com/archethic-foundation/archethic-node/pull/933 merge,  you'd need to use `NetworkChain.asynchronous_resync_many(network_chain_types)` instead.

bchamagne (approved):

Neylix (approved):

PR #933, opened 2023-03-13
"Rolling" SelfRepair

### LINKED ISSUE #912: Constantly re synchonize network transactions

Actually if a node missed some network transaction it will never be agree with other nodes and consensus error is reached. This is normally fixed on the next self repair, but it may occur in multiple hours and this is blocking transaction for a long time.
We should have a resynchronization workflow without waiting for next self repair.

Here is a possible solution: 
When a welcome node receive a transaction, it get from it's DB all the last information about network transaction:
- last oracle transaction address
- last node shared secret transaction address
- last origin transaction address (for all origin type)
- all the last public key of authorized and available nodes

Then it concatenate all these informations, create a hash and add it to the StartMining message

When the validation nodes receive this message, they will get the same informations, hash them and compare it with the one received. Here multiple case can occurs: 
- For validation node:
    - The hash is the same, start transaction mining and return Ok to the welcome node
    - The hash is not the same, do not start mining, start a resync*, respond an error to the welcome node
 - For welcome node:
     - If all node return Ok, nothing more to do
     - If only one or zero node return Ok (only error) then forward the transaction to another authorized node and start a resync* (The transaction will not be validated as there is not enough validation node)
     - If a node responded with an error, but at least two nodes responded Ok, start a resync* (transaction will be validated with the 2 or more Ok nodes)

**We may have the same hash behavior when the validation node send the ValidateTransaction message to the replication nodes.**

\* Resync means verify and replication missing network transaction:
- For scheduled network transaction (oracle, node shared secret), we can compare the previous scheduled time and the timestamp of the last transaction of this chain, if timestamp >= last schedule time it means this chain is synchronize, otherwize we need to request the last transaction and replicate it
- For non scheduled network transaction (node, origin) we need to request the last transaction of these chain and replicate it (if not already stored)

The resync process should use the RepairWorker used by the Notifier workflow to not start replicating multiple time the same transaction if we receive a lot of transaction to mine.

This workflow permit to synchronize along the day the potentially missed transactions. It can be improved but degrading the low message size:
Instead of creating a hash of all the network state, we can pass the timestamp of the last transaction of these chains. So when a node have a difference, we can know which node missed a network transaction and so start the repair only on this node. But the StartMining / ValidateTransaction messages will have an increased data size and the behavior to determine in which case we start or not the mining is more complex to determine

bchamagne:
# Description

Maintain a NetworkView that is used to detect synchronization issue between welcome node & validation nodes. 
When a desynchronization is detected, start the relevant synchronization (either P2P, or the network chains)

Fixes #912

## Type of change

-
# How Has This Been Tested?

Tested multiple scenarios: 

- welcome node is not sync (tx is forwarded)
- 1 validation node is not sync (tx is forwarded)
- 2 validation nodes are not sync (tx pass)

# Checklist:


lib/archethic/self_repair/network_chain_view.ex
```diff
@@ -0,0 +1,171 @@
...
+  @doc """
+  Return the hash of the node chain view
+  """
+  @spec get_node_chain_hash() :: binary()
+  def get_node_chain_hash() do
+    # this is not using the genserver
+    # since the nodes' state is already in the P2P module
+    P2P.authorized_and_available_nodes()
+    |> Enum.map_join(& &1.last_public_key)
+    |> Crypto.hash()
```
samuelmanzanera: I think you could iodata, without the join.

lib/archethic/self_repair/network_chain_view.ex
```diff
@@ -0,0 +1,171 @@
...
+
+  def handle_call(:get_scheduled_chains_hash, _from, state = %State{}) do
+    hash =
+      [
+        state.node_shared_secrets,
+        state.oracle,
+        state.origin |> Enum.join()
+      ]
+      |> Enum.join()
+      |> Crypto.hash()
```
samuelmanzanera: Probably the same here with iodata

lib/archethic.ex
```diff
@@ -116,10 +121,68 @@ defmodule Archethic do
...
+            acc
+            | network_chains_resync_needed: true,
+              p2p_resync_needed: true
+          }
+
+        _, acc ->
+          acc
+      end
+    )
+    |> then(fn result ->
```
samuelmanzanera, line 166: I think you should use a private function; it will be better to image what the function & pipeline does without having to deep dive into the code.

lib/archethic/self_repair/network_view.ex
```diff
@@ -0,0 +1,186 @@
...
+        Logger.info("NetworkView: Waiting for Node to complete Bootstrap. ")
+        PubSub.register_to_node_status()
+        :not_initialized
+      end
+
+    {:ok, state}
+  end
+
+  def handle_call(:get_chains_hash, _from, state = %State{}) do
+    hash =
```
samuelmanzanera, line 106: Why don't you prepare the hash once and cached it, until a new transaction invalid it and needs to be recomputed.
This function is called for each transaction, I'm afraid of the bottleneck. Ideally, an ETS would solve the bottlenecks for this high-frequency call.

lib/archethic.ex
```diff
@@ -116,12 +121,77 @@ defmodule Archethic do
...
+
+      if should_forward_transaction?(aggregated_responses, length(validation_nodes)) do
+        forward_transaction(tx, welcome_node_key)
+      else
+        :ok
+      end
+    end)
+  end
+
+  defp aggregate_start_mining_responses(responses) do
```
samuelmanzanera, line 151: This is better in terms of readability, but if we would like to improve it (for me), something like this would be even better:
```elixir
|> Stream.map(fn {:ok, res} -> res end)
|> Enum.reduce(%{
    ok: 0,
    network_chains_resync_needed: false,
    p2p_resync_needed: false
  }, reduce_start_mining_responses/2)
|> then(...)

defp reduce_start_mining_responses({:ok, %Ok{}}, acc) do
  %{acc | ok: acc.ok + 1}
end

defp reduce_start_mining_responses({:ok, %Error{reason: :network_chains_sync}}, acc) do
   %{acc | network_chains_resync_needed: true}
en

defp reduce_start_mining_responses({:ok, %Error{reason: :p2p_sync}}, acc) do
  %{acc | p2p_resync_needed: true}
end

defp reduce_start_mining_responses({:ok, %Error{reason: :both_sync}}, acc) do
%{ acc | network_chains_resync_needed: true, p2p_resync_needed: true }
end

defp reduce_start_mining_responses(_, acc), do: acc
```

samuelmanzanera (approved):

samuelmanzanera:
Nice PR !

lib/archethic/p2p/message/start_mining.ex
```diff
@@ -4,75 +4,120 @@ defmodule Archethic.P2P.Message.StartMining do
...
-        validation_node_public_keys: validation_node_public_keys
+        validation_node_public_keys: validation_node_public_keys,
+        network_chains_view_hash: network_chains_view_hash,
+        p2p_view_hash: p2p_view_hash
       }) do
     <<Transaction.serialize(tx)::binary, welcome_node_public_key::binary,
       length(validation_node_public_keys)::8,
-      :erlang.list_to_binary(validation_node_public_keys)::binary>>
+      :erlang.list_to_binary(validation_node_public_keys)::binary,
+      byte_size(network_chains_view_hash)::8, network_chains_view_hash::binary,
```
Neylix, line 119: If you use `:crypto.hash` as previously mentionned, you can remove the byte size as the hash will always be 256 bits

lib/archethic/self_repair/network_view.ex
```diff
@@ -0,0 +1,203 @@
...
+  end
+
+  @doc """
+  Return the hash of the P2P view
+  """
+  @spec get_p2p_hash() :: binary()
+  def get_p2p_hash() do
+    # this is not using the genserver
+    # since the nodes' state is already in the P2P module
+    P2P.authorized_and_available_nodes()
```
Neylix, line 60: I think this can also be in the state of the genserver, but just the hash. Because actually every time a new transaction is received or sent, you will compute the hash which will almost be the same all the time. This is a useless effort I think.
You can also use the `load_transaction` function on the `node` transaction and update the hash when you receive a new one.
Then the hash will stay in the state and the `get_p2p_hash` will just do a call to the genserver to get the state

Neylix (changes requested):

bchamagne:
After discussion with @samuelmanzanera , I'll add 2 FSMs to better deal with concurrent runs. 
They will be very simple because they are only 2 states: IDLE | RUNNING.

bchamagne:
I'm also going to change the `cast` into `calls` that returns the task(s). This way if caller wants to wait for task completion, it can.

- [ ] resync functions return the task(s)

lib/archethic/self_repair/network_chain.ex
```diff
@@ -0,0 +1,131 @@
...
+        nodes_to_resync = Enum.filter(nodes, &node_require_resync?/1)
+
+        # Load the latest node transactions
+        Task.Supervisor.async_stream_nolink(
+          Archethic.TaskSupervisor,
+          nodes_to_resync,
+          fn node ->
+            SelfRepair.resync(
+              Crypto.derive_address(node.first_public_key),
+              Crypto.derive_address(node.last_public_key)
```
Neylix, line 55: Here you should use `node.last_address` because deriving the last public key will return you the previous last address as the real last address is the hash of the next public key which we cannot know

Neylix (changes requested):

lib/archethic/utils.ex
```diff
@@ -977,6 +977,30 @@ defmodule Archethic.Utils do
...
+  @doc """
+  Run given function in a task and ensure that it is run at most once concurrently
+  """
+  def run_at_most_once_concurrently(fun, key) do
+    registry = Archethic.AtMostOnceConcurrentlyRegistry
+
+    Task.Supervisor.async_nolink(
+      Archethic.TaskSupervisor,
+      fn ->
+        case Registry.lookup(registry, key) do
```
samuelmanzanera, line 989: I think you can make the lookup before the process's spawning, otherwise the process would be created for nothing.

lib/archethic/p2p/message/start_mining.ex
```diff
@@ -4,75 +4,129 @@ defmodule Archethic.P2P.Message.StartMining do
...
-         {:mining, false} <- {:mining, Mining.processing?(tx.address)} do
-      {:ok, _} = Mining.start(tx, welcome_node_public_key, validation_nodes)
-      %Ok{}
-    else
-      {:election, false} ->
-        Logger.error("Invalid validation node election",
-          transaction_address: Base.encode16(tx.address),
-          transaction_type: tx.type
-        )
+    case check_synchronization(network_chains_view_hash, p2p_view_hash) do
```
samuelmanzanera, line 52: I'm wondering if we can reduce this piece of code.
There is a lot which happening into a single function with multiple calls and with statements.
If we can, it should be great to refactor a bit

Neylix (approved):

Neylix (approved):

Neylix (approved):

Issue #931, opened 2023-03-13
Infinite looping in forward transaction 


bchamagne:
### Describe the problem you discovered

There might be some scenario where the forward transaction mechanism leads to infinite looping.

For example: 
Imagine there are N nodes, and they all are synchronized but NODE_1 and NODE_2 (they are available).

![Screenshot 2023-03-13 at 11-41-46 Online FlowChart   Diagrams Editor - Mermaid Live Editor](https://user-images.githubusercontent.com/74045243/224678944-9e79fd6a-65dd-4a13-bfea-222a07a48104.png)

NODE_1 get `send_new_tx` call, here 2 possibilities:
1/ tx is forwarded to NODE_2 (SEE BELOW)
4/ tx is forwarded to NODE_1 (= INFINITE LOOP)

Lets continue after 1/, there are 3 possibilities here: 
2/ tx is forwarded to NODE_1 (= INFINITE LOOP)
3/ tx is forwarded to NODE_2 (=INFINITE LOOP)
5/ tx is forwarded to NODE_N (OK)

PR #937, opened 2023-03-15
infinite_looping_in_forward_transaction_#931

apoorv-2204:
# Description

Fixes 

- [ ] #931 

## Type of change

# How Has This Been Tested?

- Test A
- Test B

# Checklist:


lib/archethic.ex
```diff
@@ -74,25 +74,49 @@ defmodule Archethic do
...
+        current_node = Crypto.first_node_public_key()
+
+        index = Enum.find_index(nodes, &(&1.first_public_key == current_node))
+        {_l, r} = Enum.split(nodes, index + 1)
+        r
+      else
+        nodes
+      end
+
+    spawn(fn ->
```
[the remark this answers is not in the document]
apoorv-2204: its lightweight and much better than tasks supervisor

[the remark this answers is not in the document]
samuelmanzanera: You should not let process unsupervised in your system, as it can lead to unexpected issues and problems. Agree with @neylix

lib/archethic/p2p/message/new_transaction.ex
```diff
@@ -17,13 +17,8 @@ defmodule Archethic.P2P.Message.NewTransaction do
...
   def process(%__MODULE__{transaction: tx, welcome_node: node_pbkey}, _) do
-    case Archethic.send_new_transaction(tx, node_pbkey) do
-      :ok ->
-        %Ok{}
-
-      {:error, :network_issue} ->
-        %Error{reason: :network_issue}
-    end
+    :ok = Archethic.send_new_transaction(tx, node_pbkey)
+    %Ok{}
```
bchamagne, line 21: why did you remove all the error cases?

lib/archethic.ex
```diff
@@ -74,25 +73,50 @@ defmodule Archethic do
...
-         |> P2P.nearest_nodes()
-       )
+  @spec forward_transaction(tx :: Transaction.t(), welcome_node_key :: Crypto.key()) :: :ok
+  defp forward_transaction(tx, welcome_node_key) do
+    %Node{network_patch: welcome_node_patch} = P2P.get_node_info!(welcome_node_key)
+
+    nodes =
+      P2P.authorized_and_available_nodes()
+      |> Enum.reject(&(&1.first_public_key == welcome_node_key))
+      |> Enum.sort_by(& &1.first_public_key)
```
bchamagne, line 83: is this `sort_by` useful? look like the `nearest_nodes` is resorting it right after?

lib/archethic.ex
```diff
@@ -74,25 +73,50 @@ defmodule Archethic do
...
+      |> P2P.nearest_nodes(welcome_node_patch)
+      |> Enum.filter(&Node.locally_available?/1)
+
+    nodes =
+      if Crypto.first_node_public_key() != welcome_node_key do
+        #  if this node is not the welcome node then select
+        # next node from the this node position in nodes list
+        current_node = Crypto.first_node_public_key()
+
+        index = Enum.find_index(nodes, &(&1.first_public_key == current_node))
```
bchamagne, line 93: since there is a sort by nearest_nodes, is the order of the nodes determinist?

Neylix: The sort is based on the network patch if the welcome node, not the local node one.
So as the network patch of the welcome is the same across all node, the sort result in the same ordre

samuelmanzanera: Yes sure, I didn't see the welcome_node_patch, so definitively as long as the network remains the same across the nodes, it should be deterministic.

lib/archethic.ex
Neylix: This sort is usefull when there is multiple nodes in the same patch. We need to have the same order of node within the same patch, while the nearest_nodes function does not ensure the sort inside a patch

Neylix (commented):
Looks good to me, can your rebase to solve conflict and not override new changes ?

Neylix (changes requested):

test/archethic_test.exs
```diff
@@ -130,6 +241,101 @@ defmodule ArchethicTest do
...
+      end)
+
+      # last address is d/f it returns last address from quorum
+      assert {:error, "willnotmatchaddress"} = SharedSecrets.verify_synchronization()
+
+      # trying to ssend a tx when NSS chain not synced
+      assert :ok = Archethic.send_new_transaction(tx)
+
+      # start repair and should bottleneck requests
+      pid = SelfRepair.repair_in_progress?(nss_genesis_address)
```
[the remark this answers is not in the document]
Neylix: I unresolve the thread as this as not been changed

test/archethic_test.exs
```diff
@@ -130,6 +241,101 @@ defmodule ArchethicTest do
...
+      Process.sleep(100)
+      assert pid != nil
+
+      assert_receive({:forwarded_to_node2, ^tx})
+      Process.sleep(100)
```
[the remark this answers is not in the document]
Neylix: I unresolve the thread as this as not been changed

Neylix (changes requested):

test/archethic_test.exs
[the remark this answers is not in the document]
apoorv-2204: i deleted the test

lib/archethic.ex
```diff
@@ -80,25 +79,50 @@ defmodule Archethic do
...
+        {_l, r} = Enum.split(nodes, index + 1)
+        r
+      else
+        nodes
+      end
+
+    TaskSupervisor
+    |> Task.Supervisor.async_nolink(fn ->
```
samuelmanzanera, line 107: Why do you need other process to send the transaction? 
And also, what is the purpose of the `async`? You are not awaiting the task result.

lib/archethic.ex
```diff
@@ -80,25 +79,50 @@ defmodule Archethic do
...
+      %NewTransaction{transaction: tx, welcome_node: welcome_node_key}
+      |> do_forward_transaction(nodes)
+    end)
+
+    :ok
```
samuelmanzanera, line 112: The network issue is never returned here? I think it would still be useful.

lib/archethic.ex
apoorv-2204: spawn?

apoorv-2204: node 1 waits for node 2 to return ok , node 2 waits for node 3 to send txn and fget back response,

Neylix (approved):

bchamagne (commented):
To make the test work I had to hardcode a network patch, because there is a randomness in local development that breaks this feature. 

I tested this scenario: 

:heavy_check_mark:  3 nodes, I modified the code of 2 nodes to mark them as invalid for processing (to force the forward).  I then sent a transaction to `A` that forward it to `B` that forward it to `C`. 

- I think your last modification modify the return type of `send_new_transaction`, is this ok?
- When you remove the task.async, we now have `A` that wait until timeout, and when the timeout occur, `A` will try to send to `C` (in the same scenario as above), which I believe is not good. 

ps: please keep commit history

apoorv-2204:
> To make the test work I had to hardcode a network patch, because there is a randomness in local development that breaks this feature.
> 
> I tested this scenario:
> 
> ✔️ 3 nodes, I modified the code of 2 nodes to mark them as invalid for processing (to force the forward). I then sent a transaction to `A` that forward it to `B` that forward it to `C`.
> 
> * When you remove the task.async, we now have `A` that wait until timeout, and when the timeout occur, `A` will try to send to `C` (in the same scenario as above), which I believe is not good.
> 
> ps: please keep commit history

Hi, we are relying on timeout to notifiy user. I updated geo patch for null patch for localhost

samuelmanzanera:
> > To make the test work I had to hardcode a network patch, because there is a randomness in local development that breaks this feature.
> > I tested this scenario:
> > ✔️ 3 nodes, I modified the code of 2 nodes to mark them as invalid for processing (to force the forward). I then sent a transaction to `A` that forward it to `B` that forward it to `C`.
> > 
> > * When you remove the task.async, we now have `A` that wait until timeout, and when the timeout occur, `A` will try to send to `C` (in the same scenario as above), which I believe is not good.
> > 
> > ps: please keep commit history
> 
> Hi, we are relying on timeout to notifiy user. I updated geo patch for null patch for localhost

Indeed, I wasn't against the spawn a process it's just the Task.async expected to await something, maybe a start_child would be sufficient. (But QUID of the error network_issue then ?)
Otherwise we will have cascading timeout.

bchamagne:
Maybe I got it wrong, but I believe my last comment is still relevant. We talked about an ACK msg no?

apoorv-2204:
> Maybe I got it wrong, but I believe my last comment is still relevant. We talked about an ACK msg no?

Replay back a message?

Neylix:
> Maybe I got it wrong, but I believe my last comment is still relevant. We talked about an ACK msg no?

The ack message is returned in the `process` function of the message `NewTransaction`

Neylix (approved):

bchamagne (approved):

PR #941, opened 2023-03-15
Rework self repair verification

Neylix:
# Description

Rework the self repair verification when replicating a transaction.
Before, when downloading a transaction, the node has to do some small verification about transaction integrity, validation stamp election ... But this was leading to a lot of synchronization error mainly with invalid election.
Now, the replication attestation with all the storage node confirmations is written in the summary aggregate. So when dowloading a transaction, the node can just verify the confirmation signature to ensure the transaction is valid (if 197 nodes validated the transaction when it happen, it is always valid).
This improve the security level, as well as increasing the self repair speed as the node do not have to download other data to ensure the transaction is valid.
An other improvement is to add a checksum of the validation stamp in the transaction summary. So when a node download a transaction, it can be 100% sure that it is the same as the one in the replication attestation.

Future improvement: aggregate all the storage confirmations signature into one to reduce the storage size if the summary aggregate

**Attention: this issue is not yet retrocompatible with the existing mainnet / testnet datas. An other PR will come later to do a data migration from old summary aggregate version to the new one.**

PS: For review, I tried to make meaningfull commits by subject, they may not work individually, but you can review commits one by one. It should be easier to understand

## Type of change

# How Has This Been Tested?

Unit test

Run multiple node, wait some summaries, start a new one. It should download and repair all the transaction properly.
To have non nominal case, we need to update the code to remove or modify some storage confirmation to raise errors during self repair

# Checklist:


lib/archethic/beacon_chain/replication_attestation.ex
```diff
@@ -163,49 +159,43 @@ defmodule Archethic.BeaconChain.ReplicationAttestation do
     deserialize_confirmations(rest, nb_replicas_signature, [{position, signature} | acc])
   end
 
+  @doc """
+  Return the index of the node based on the authorized node at this timestamp
+  """
+  @spec get_node_index(Crypto.key(), DateTime.t()) :: non_neg_integer()
```
samuelmanzanera, line 176: I think this function should be in the P2P module, as it can be useful elsewhere.

lib/archethic/beacon_chain/replication_attestation.ex
```diff
@@ -163,49 +159,43 @@ defmodule Archethic.BeaconChain.ReplicationAttestation do
...
-        check_summary_consistency? \\ false
-      ) do
+  def validate(%__MODULE__{
+        transaction_summary: tx_summary = %TransactionSummary{timestamp: timestamp},
+        confirmations: confirmations
+      }) do
     tx_summary_payload = TransactionSummary.serialize(tx_summary)
 
-    authorized_nodes = P2P.authorized_and_available_nodes(timestamp)
+    nodes_public_key =
```
samuelmanzanera, line 195: A little type here but helpful in terms of meaning: `node_public_keys`

samuelmanzanera (approved):
Nevertheless good PR ;)

bchamagne (approved):

PR #954, opened 2023-03-20
Add conflict resolution to 2 functions of the TransactionChain module

### LINKED ISSUE #942: Missing conflict resolvers for P2P requests

- `TransactionChain.fetch_genesis_address_remotely` 
- `TransactionChain.fetch_first_transaction_address_remotely`

are missing conflicts resolvers in the quorum calls, causing the inconsistency in the network retrieval and chain persistence.

The genesis address and transaction address requests should return a timestamp to make sure to get the earliest address is taken in the conflict resolver.
**We should discard the default unix time: `DateTime.from_unix!(0)`**

bchamagne:
# Description

Add conflict resolution to `fetch_first_transaction_address_remotely` and `fetch_genesis_address_remotely`.

Fixes #942

## Type of change

# How Has This Been Tested?

Added the unit tests required

# Checklist:

### DIFF
--- a/lib/archethic/p2p/message/get_genesis_address.ex
+++ b/lib/archethic/p2p/message/get_genesis_address.ex
@@ -18,7 +18,19 @@ defmodule Archethic.P2P.Message.GetGenesisAddress do
   @spec process(__MODULE__.t(), Crypto.key()) :: GenesisAddress.t()
   def process(%__MODULE__{address: address}, _) do
     genesis_address = TransactionChain.get_genesis_address(address)
-    %GenesisAddress{address: genesis_address}
+
+    # Genesis is not a transaction, so to get the timestamp for the conflict resolver
+    # we return the timestamp of the 1st transaction
+    timestamp =
+      case TransactionChain.get_first_transaction_address(genesis_address) do
+        {:ok, {_first_tx_address, first_tx_timestamp}} ->
+          first_tx_timestamp
+
+        {:error, :transaction_not_exists} ->
+          DateTime.utc_now()
+      end
+
+    %GenesisAddress{address: genesis_address, timestamp: timestamp}
   end
 
   @spec serialize(t()) :: bitstring()
--- a/lib/archethic/transaction_chain.ex
+++ b/lib/archethic/transaction_chain.ex
@@ -1014,7 +1014,7 @@ defmodule Archethic.TransactionChain do
           {:ok, binary()} | {:error, :network_issue}
   def fetch_genesis_address_remotely(address, nodes) when is_binary(address) do
     conflict_resolver = fn results ->
-      Enum.min_by(results, & &1.timestamp)
+      Enum.min_by(results, & &1.timestamp, DateTime)
     end
 
     case P2P.quorum_read(nodes, %GetGenesisAddress{address: address}, conflict_resolver) do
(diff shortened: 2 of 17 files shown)


lib/archethic/transaction_chain.ex
```diff
@@ -1029,7 +1033,26 @@ defmodule Archethic.TransactionChain do
           {:ok, binary()} | {:error, :network_issue} | {:error, :does_not_exist}
   def fetch_first_transaction_address_remotely(address, nodes)
       when is_binary(address) and is_list(nodes) do
-    case P2P.quorum_read(nodes, %GetFirstTransactionAddress{address: address}) do
+    conflict_resolver = fn results ->
+      results
+      |> Enum.reduce(%NotFound{}, fn
```
Neylix, line 1038: To filter the `%NotFound` you can use the same way as it is done in beacon chain `fetch_summaries_aggregate` :
```elixir
conflict_resolver = fn results ->
  # Prioritize results over not found
  case nil <- Enum.find(results, &match?(%FirstTransactionAddress{}, &1)) do
    nil -> %NotFound{}

    res ->
      Enum.min_by(res, &DateTime.to_unix(&1.timestamp, :millisecond))
  end
end
```

Neylix (changes requested):

lib/archethic/transaction_chain.ex
samuelmanzanera: I think you should also discard the `DateTime.from_unix!(0)`

Neylix: > I think you should also discard the `DateTime.from_unix!(0)`

This would never happen for the first transaction message. When there is no transaction in the chain of the requested address, it return a NotFound message

lib/archethic/p2p/message/get_genesis_address.ex
```diff
@@ -18,7 +18,19 @@ defmodule Archethic.P2P.Message.GetGenesisAddress do
   @spec process(__MODULE__.t(), Crypto.key()) :: GenesisAddress.t()
   def process(%__MODULE__{address: address}, _) do
     genesis_address = TransactionChain.get_genesis_address(address)
-    %GenesisAddress{address: genesis_address}
+
+    # Genesis is not a transaction, so to get the timestamp for the conflict resolver
+    # we return the timestamp of the 1st transaction
+    timestamp =
+      case TransactionChain.get_first_transaction_address(genesis_address) do
```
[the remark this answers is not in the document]
bchamagne: As discussed, as long as a node stores a transaction of the chain, he knows all the addresses. No change required.

lib/archethic/transaction_chain.ex
```diff
@@ -1013,7 +1013,11 @@ defmodule Archethic.TransactionChain do
   @spec fetch_genesis_address_remotely(address :: binary(), list(Node.t())) ::
           {:ok, binary()} | {:error, :network_issue}
   def fetch_genesis_address_remotely(address, nodes) when is_binary(address) do
-    case P2P.quorum_read(nodes, %GetGenesisAddress{address: address}) do
+    conflict_resolver = fn results ->
+      Enum.min_by(results, & &1.timestamp)
```
Neylix, line 1017: You should specify the sorter function for a DateTime like `Enum.min_by(results, & &1.timestamp, DateTime)` as it is described in DateTime doc

Neylix (changes requested):

Neylix (approved):

Neylix (approved):

PR #957, opened 2023-03-23
Initialize migration workflow

Neylix:
# Description

This PR aim to initialize the migration workflow.
Each migration needed will be an elixir scripts in `priv/migration_tasks` according the name to `[version]-[description].exs`.

There is 2 times when the migration is called:
- before starting the application
- during the execution of an upgrade thanks to the appup file. Migrate function call is automatically added at the end of appup

The migration file in DB always contains the last version the node runned the migrate tool, even if there is no migration to execute

## Type of change

# How Has This Been Tested?

- Unit tests
- Create a new migration script that create a file somewhere. Build a release, start the node, run a hot reload, all the file should have been created

# Checklist:

### DIFF
--- a/lib/mix/tasks/migrate.ex
+++ b/lib/mix/tasks/migrate.ex
@@ -18,15 +18,15 @@ defmodule Mix.Tasks.Archethic.Migrate do
         # File does not exist when it's the first time the node is started
         # We create the folder to write the migration file on first start
         migration_file_path |> Path.dirname() |> File.mkdir_p!()
+        File.write(migration_file_path, new_version)
         []
       end
 
-    Enum.each(migrations_to_run, fn module ->
+    Enum.each(migrations_to_run, fn {version, module} ->
       :erlang.apply(module, :run, [])
       unload_module(module)
+      File.write(migration_file_path, version)
     end)
-
-    File.write(migration_file_path, new_version)
   end
 
   defp filter_migrations_to_run(last_version) do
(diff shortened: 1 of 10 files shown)


lib/mix/tasks/migrate.ex
```diff
@@ -6,28 +6,110 @@ defmodule Mix.Tasks.Archethic.Migrate do
...
+      migration_version = Regex.run(~r/[0-9\.]*(?=-)/, file_name) |> List.first()
+      {migration_version, migration_path}
+    end)
+    |> Enum.filter(fn {migration_version, _} ->
+      keep_migration?(current_version, migration_version, function)
+    end)
+    |> Enum.map(fn {version, path} -> {version, Code.eval_file(path)} end)
+    |> Enum.filter(fn
+      {_version, {{:module, module, _, _}, _}} ->
+        case module.__info__(:functions) |> Keyword.has_key?(function) do
```
bchamagne, line 76: You can use `function_exported?/3`

lib/mix/tasks/migrate.ex
```diff
@@ -6,28 +6,110 @@ defmodule Mix.Tasks.Archethic.Migrate do
...
+      do: do_run(version, function)
+
+  def run(_), do: :ok
+
+  defp do_run(current_version, function) do
+    migration_file_path = EmbeddedImpl.db_path() |> ChainWriter.migration_file_path()
+
+    migrations_to_run =
+      if File.exists?(migration_file_path) do
+        read_file(migration_file_path) |> filter_migrations_to_run(function)
```
bchamagne, line 37: There's something I don't get but that might not be an issue, if there are 4 migrations to catch up, this will do the 4 pre_start one after the other. Is it ok ? Don't you need to do pre_start1 then pre_up1 then post_up1 and only then run pre_start2 then pre_up2, then post_up2 etc. ?

Neylix: The case here is that the pre_start is triggered when the node start, and the pre/post upgrade are triggered when the node is currently running and doing an hot upgrade.
When a node is currenlty running it means that it is on the last version, or it will integrate new version one by one (with the governance transactions).
So for a specific version you can run either pre_start or pre/post upgrade but never both.
And in theory the code in pre_start would be the same as pre/post upgrade

samuelmanzanera (approved):

samuelmanzanera (approved):

PR #969, opened 2023-03-31
add AtVsn credo rule

bchamagne:
Add a credo rules to check for `@vsn` attributes on GenServer and GenStateMachine.

Example:
Screenshot_2023-03-31_16-13-34

I find it add ~300ms run time on my computer 5.8s vs 5.5s.

### DIFF
--- a/lib/checks/at_vsn.ex
+++ b/lib/checks/at_vsn.ex
@@ -1,0 +1,41 @@
+defmodule Archethic.Checks.AtVsn do
+  @moduledoc """
+  Check that GenServer & GenStateMachine have a @vsn attribute
+  """
+
+  use Credo.Check
+
+  defstruct is_a_server: false,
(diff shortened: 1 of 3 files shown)


lib/checks/at_vsn.ex
```diff
@@ -0,0 +1,41 @@
+defmodule Archethic.Checks.AtVsn do
+  @moduledoc """
+  Check that GenServer & GenStateMachine have a @vsn attribute
+  """
+
+  use Credo.Check
```
Neylix, line 6: Actually when running credo is inform there is an check that return an issue, but credo does not handle it as an issue so credo pass without error.
To fix this we can add an category to the check like `:warning` or `:consistency` so credo will report it as an found issue
ie: `use Credo.Check, category: :warning`

Neylix (changes requested):

PR #973, opened 2023-04-05
Smart Contracts: Safe floating point maths + automatic amount conversion to 10^8

### LINKED ISSUE #893: Transaction amount should verify it is BigInt

Actually user can create smart contract like this :
```elixir
  # quantite d’UCO reçu
  amount_send = transaction.uco_transfers[contract.address]
  if amount_send > 0.5 do
      # Conversion UCO en token test
      token_to_credit = amount_send * 10 - 0.5

      # Nouvelle transaction
      set_type transfer
      add_token_transfer to: transaction.address, token_address: "000052EE087848D039231528532B53B1182C40508B30B3FBB8AB5F616A46484F6241" , amount: token_to_credit
  end
end
```

The number 0.5 causes problem as all amount should be in BigInt, so in this case, the final amount will be a float causing this error : 

image

At some point (define where) we should ensure that the amount of a transfer is an integer

bchamagne:
# Description

**This PR is based on https://github.com/archethic-foundation/archethic-node/pull/967. To be rebased as soon as it is merged.**

1. User may now code `Contract.add_uco_transfer to: "...", amount: 1` to transfer 1 UCO (before he/she had to specify amount: 100000000 to send 1 UCO. Makes the language friendlier. 
1. Floating points mathematics are now done in BigInt so there is no floating point precision issue.
1. Since we use BigInt for maths, we loose the types of the numbers (1 + 1 == 2.0). To address this inconvenience, I updated the library to have `String.from_number/1` and `String.to_number/1` instead of `String.from_float/1`, `String.from_int/1`, `String.to_float/1`, `String.to_int/1`.

Fixes #893

_ps: I suggest to ignore `test/archethic/contracts/interpreter/condition_interpreter_test.exs` when reviewing. This is just tests adaption due to the function being stricter than before (it is a good thing)._

## Type of change

# How Has This Been Tested?

Unit tests :heavy_check_mark: 
Manual tests: :grey_question: 

# Checklist:

### DIFF
--- a/lib/archethic/contracts/interpreter/action_interpreter.ex
+++ b/lib/archethic/contracts/interpreter/action_interpreter.ex
@@ -4,6 +4,7 @@ defmodule Archethic.Contracts.Interpreter.ActionInterpreter do
   alias Archethic.TransactionChain.Transaction
   alias Archethic.TransactionChain.TransactionData
 
+  alias Archethic.Contracts.ContractConstants, as: Constants
   alias Archethic.Contracts.Interpreter.ASTHelper, as: AST
   alias Archethic.Contracts.Interpreter.CommonInterpreter
   alias Archethic.Contracts.Interpreter.Library
@@ -42,6 +43,13 @@ defmodule Archethic.Contracts.Interpreter.ActionInterpreter do
 
     # initiate a transaction that will be use by the "Contract" module
     next_tx = %Transaction{data: %TransactionData{}}
+
+    # Apply some transformations to the transactions
+    # We do it here because the Constants module is still used by InterpreterLegacy
+    constants =
+      constants
+      |> Constants.map_transactions(&Constants.stringify_transaction/1)
+      |> Constants.map_transactions(&Constants.cast_transaction_amount_to_float/1)
 
     # we use the process dictionary to store our scope
     # because it is mutable.
--- a/lib/archethic/contracts/interpreter/common_interpreter.ex
+++ b/lib/archethic/contracts/interpreter/common_interpreter.ex
@@ -361,12 +361,11 @@ defmodule Archethic.Contracts.Interpreter.CommonInterpreter do
   end
 
   # BigInt mathematics to avoid floating point issues
-  @unit_uco 100_000_000
+  # the `0.0 + x` is used to cast integers to floats
   def postwalk(_node = {:*, meta, [lhs, rhs]}, acc) do
     new_node =
-      quote line: Keyword.fetch!(meta, :line),
-            bind_quoted: [lhs: lhs, rhs: rhs, bigint: @unit_uco] do
-        bigint * lhs * bigint * rhs / (bigint * bigint)
+      quote line: Keyword.fetch!(meta, :line) do
+        Float.floor(0.0 + unquote(lhs) * unquote(rhs), 8)
       end
 
     {new_node, acc}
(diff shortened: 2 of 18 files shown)


lib/archethic/contracts/interpreter/common_interpreter.ex
```diff
@@ -353,6 +360,80 @@ defmodule Archethic.Contracts.Interpreter.CommonInterpreter do
     {new_node, acc}
   end
 
+  # BigInt mathematics to avoid floating point issues
+  # the `0.0 + x` is used to cast integers to floats
+  def postwalk(_node = {:*, meta, [lhs, rhs]}, acc) do
+    new_node =
+      quote line: Keyword.fetch!(meta, :line) do
+        Decimal.to_float(
```
Neylix, line 368: You could create a private function to trunc the decimals as you use the same conversion on every function. Maybe somtheing like
```elixir
Decimal.mult(
  Decimal.from_float(0.0 + unquote(lhs)),
  Decimal.from_float(0.0 + unquote(rhs))
)
|> trunc_decimals()

# Maybe rename this function
defp trunc_decimals(val), do:
  val |> Decimal.round(8, :floor) |> Decimal.to_float()
```

lib/archethic/contracts/interpreter/common_interpreter.ex
```diff
@@ -353,6 +360,80 @@ defmodule Archethic.Contracts.Interpreter.CommonInterpreter do
     {new_node, acc}
   end
 
+  # BigInt mathematics to avoid floating point issues
+  # the `0.0 + x` is used to cast integers to floats
+  def postwalk(_node = {:*, meta, [lhs, rhs]}, acc) do
```
Neylix, line 365: I believe in elixir there is the operator `**` which is the power of a number that replace the erlang `:math.pow(x, y)` maybe you could also handle this operator

lib/archethic/contracts/interpreter/condition_validator.ex
```diff
@@ -16,12 +16,12 @@ defmodule Archethic.Contracts.Interpreter.ConditionValidator do
   """
   @spec valid_conditions?(Conditions.t(), map()) :: boolean()
   def valid_conditions?(conditions = %Conditions{}, constants = %{}) do
+    # Apply some transformations to the transactions
+    # We do it here because the Constants module is still used by InterpreterLegacy
     constants =
```
Neylix, line 21: Same here for legacy interpreter

lib/archethic/contracts/interpreter/action_interpreter.ex
```diff
@@ -43,6 +44,13 @@ defmodule Archethic.Contracts.Interpreter.ActionInterpreter do
     # initiate a transaction that will be use by the "Contract" module
     next_tx = %Transaction{data: %TransactionData{}}
 
+    # Apply some transformations to the transactions
+    # We do it here because the Constants module is still used by InterpreterLegacy
+    constants =
```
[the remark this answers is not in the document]
samuelmanzanera: Indeed, the automatic conversion should be for the new smart contract's interpreter. This would make more sense.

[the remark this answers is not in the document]
Neylix: Comment mislead  me as this will not have any action on InterpreterLegacy.

lib/archethic/contracts/interpreter/common_interpreter.ex
bchamagne: There's no `pow` in the Decimal library yet. This may be done in a later issue.

Neylix (approved):

samuelmanzanera (approved):

PR #979, opened 2023-04-17
Migrate old attestations

### LINKED ISSUE #956: Migrate old self repair / replication attestation to the new implementation

In PR https://github.com/archethic-foundation/archethic-node/pull/941 a new implementation of the replication attestation / summary aggregate / transaction summary has been done.
Theses changes are not retrocompatible yet with the current mainnet version

We need to provide a migration tool to update old summary aggregate and beacon summary.
As well as being able to handle both version of replication attestation

Neylix:
# Description

This issue handle the migration from old transaction summary and replication attestation to the new ones added by #941

It adds a migration script to transform old summary aggregate and summary with the new transaction summary containing the validation stamp checksum

Also migrate the current slots stored in summary cache with the new version.

Slot stored in subset is transformed only when an other process request it

Fixes #956 

## Type of change

# How Has This Been Tested?

First need to create a release on version 1.0.7.
Do do so, go on develop and remove the commits from pr #941 and #927, you can do an interactive rebase `git rebase -i HEAD~30` and drops the following commits

image

Then runs these commands : 
```bash
# create the 1.0.7 release
./scripts/test-release.sh -i
# remove release artefact
git reset --hard
# switch to pr branch 
git switch Migrate-old-attestations
```

Here you can start the 2 nodes created:
`./test_release_node1/bin/archethic_node console`
`ARCHETHIC_CRYPTO_SEED=node2 ARCHETHIC_P2P_PORT=3003 ARCHETHIC_HTTP_PORT=4001 ARCHETHIC_HTTPS_PORT=5001 ./test_release_node2/bin/archethic_node console
`

Now you need to prepare the branch to the new version. You have to update the version number and also remove a code change that will fail (as we already run the new version)
You can apply this patch to do so : 
[patch.txt](https://github.com/archethic-foundation/archethic-node/files/11251157/patch.txt)
`git apply patch.txt`
(You can commit this patch to avoid applying it each time you switch branch)

Create the new release 1.0.8
`./scripts/test-release.sh -p`

You can wait the 2 nodes are authorized and create some summary aggregate.

Then execute the new upgrade
`./scripts/test-release.sh -u`

The upgrade should pass successfully without error, you can access to old summaries using the explorer (make sure the data is not cached)

Next summary time should create summary without error (You may have error with StatsCollector which was not started by default in the appup file)

# Checklist:


lib/archethic/transaction_chain/transaction_summary.ex
```diff
@@ -207,4 +212,55 @@ defmodule Archethic.TransactionChain.TransactionSummary do
...
     }
   end
+
+  @doc """
+  This function will be used during the summary day of 1.0.8 upgrade. This function can be deleted after the upgrade.
+  Migrate this function into files 1.0.8-migrate_old_tx_summaries
+  Transform a TransactionSummary without version into a TransactionSummary v1
+  """
+  @spec transform_1_0_8_summary(t()) :: t()
+  def transform_1_0_8_summary(tx_summary = %__MODULE__{version: 1}), do: tx_summary
```
samuelmanzanera, line 222: Why is it in this file ? I think the migration should have this function, hence do no need to alter this file after the upgrade, so the function will be contextualized for the migration only.

samuelmanzanera: I think it would be clearer to use a `Slot.transform/2` and `TransactionSummary.transform/2` which just do transformation of data structure.

bchamagne (approved):
tested with new changes, migration worked fine

PR #988, opened 2023-04-21
Smart Contract: Better error printing

bchamagne:
# Description

Add a case to the `format_error_reason/2` to better print the AST of `M.F(A, A, A)`. 
This used to return message like this: 

```
"invalid arguments for function List.empty? - {:atom, \"List\"} . :atom => \"empty?\"(12) - L8"
```

## Type of change

# How Has This Been Tested?

Tests added to the suite

# Checklist:

### DIFF
--- a/lib/archethic/contracts/interpreter.ex
+++ b/lib/archethic/contracts/interpreter.ex
@@ -168,6 +168,26 @@ defmodule Archethic.Contracts.Interpreter do
     do_format_error_reason(reason, module, metadata)
   end
 
+  def format_error_reason(
+        {{:., metadata, [{:__aliases__, _, [atom: module_name]}, {:atom, function_name}]}, _,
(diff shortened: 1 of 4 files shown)


samuelmanzanera:
Is it possible to return something like: `"invalid arguments for function List.empty? with [12] - L8"` ?

bchamagne:
> Is it possible to return something like: `"invalid arguments for function List.empty? with [12] - L8"` ?

I modified it a little bit so we don't have a "list of arguments" but directly the function call. 

Examples: 
`invalid function arguments - List.empty?(12) - L4`
`invalid function arity - List.empty?([1], "foobar") - L4`
`unknown function - List.non_existing([1, 2, 3]) - L4`

What do you think?

Neylix (approved):

PR #990, opened 2023-04-21
start app even if P2P port cannot be opened

samuelmanzanera:


bchamagne (commented):
Can't we handle this with `Application.get_env(:archethic, IPLookup) == ManualPortForwarding` instead?

Neylix (approved):

PR #999, opened 2023-04-26
Smart Contracts: Optional inherit condition block

### LINKED ISSUE #945: Simplify smart contract inherit constraints
For now, the inherit constraints are required and explicit about the rules to accept a new transaction.
However, this can be annoying and slow down adoption of smart contract on Archethic.

We should propose a way to avoid the need of explicit rules, which are already in the actions block of the smart contract.
This is related to #925 (validation nodes will fetch the previous transaction to get the smart contract code and try to expect the next one to match the incoming transaction).

The inherit conditions could still be used for other purposes, such as origin family constraints, inherit exceptions (authorizations to upgrade code), etc...

bchamagne:
# Description

Conditions inherit are now optional.
Fixes #945 

## Type of change

# How Has This Been Tested?

Unit tested and tested manually with 3 scenarios:
1. no inherit condition
2. inherit condition that validates the next transaction
3. empty inherit condition that closes the chain

# Checklist:

### DIFF
--- a/lib/archethic/contracts.ex
+++ b/lib/archethic/contracts.ex
@@ -65,15 +65,10 @@ defmodule Archethic.Contracts do
      %Contract{
        version: version,
        triggers: triggers,
-       conditions: %{inherit: inherit_conditions}
+       conditions: conditions
      }} = Interpreter.parse(code)
 
-    constants = %{
-      "previous" => Constants.from_transaction(prev_tx),
-      "next" => Constants.from_transaction(next_tx)
-    }
-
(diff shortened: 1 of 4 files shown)


bchamagne (commented):
I forgot about the `condition inherit: []` that should closes the contract.
I'm changing status to DRAFT.

EDIT: after discussion with @samuelmanzanera , we should keep the same behaviour: `condition inherit: []` will close the contract whereas omitting the condition inherit block will always pass the validation.

- [x] Handle the closing of the chain as before

Neylix (approved):

PR #1006, opened 2023-05-03
Smart Contracts: Forbid new contracts to use the legacy interpreter

### LINKED ISSUE #968: Ensure new smart contract are not legacy one

Actually there is the legacy interpreter and the normal interpreter which support `@version` field.
We should ensure all new smart contract are created with the normal interpreter and are versionned as the legacy interpreter may not be maintained.

In the pending validation we already parse the contract to ensure the code is valid.
We could have another control before parsing, that verify if the last transaction has code or not. 
If the last transaction does not have code, the current transaction should use versionned interpreter with @version. 
If the previous transaction has code, then if the code is legacy, we allow legacy code, if the code is not legacy, the new transaction must also be on versionned interpreter

bchamagne:
# Description

Ensure new smart contracts are not using the legacy interpreter.
Fixes #968 

## Type of change

# How Has This Been Tested?

Unit tested + manual tests 

1. Try to create a legacy contract =  
Screenshot_2023-05-03_11-24-01

2. Create a legacy contract on develop then change branch = contract still runs

# Checklist:

### DIFF
--- a/lib/archethic/mining/pending_transaction_validation.ex
+++ b/lib/archethic/mining/pending_transaction_validation.ex
@@ -130,17 +130,32 @@ defmodule Archethic.Mining.PendingTransactionValidation do
 
   defp validate_contract(%Transaction{data: %TransactionData{code: ""}}), do: :ok
 
-  defp validate_contract(%Transaction{
-         data: %TransactionData{code: code, ownerships: ownerships}
-       })
+  defp validate_contract(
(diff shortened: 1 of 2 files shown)


bchamagne:
Doing this would break all existing Dapps such as the bridge. 

The goal of the issue is to help the users that forgets to write `@version`. We can do so in the Smart Contract Playground. 
I'll create an issue there.

PR #1023, opened 2023-05-10
Handle p2p update from self repair in network view

Neylix:
# Description

The network view is not updated when a new node is available after a self repair. So it was causing error as if a node restarts, it doesn't have the same view as the one that didn't restarted.

To fix this we have a new function `NetworkView.update_p2p_view` which is called at the end of a summary repair.
Also in the state of the network view we can have 2 hashes, a current one and a next one that will replace the current one once the threashold time (availability update) is passed. 

## Type of change

# How Has This Been Tested?

Unit test

Start 2 nodes, wait them to be authorized and available.
Restart one, the view should be the same on both nodes

# Checklist:

### DIFF
--- a/lib/archethic/self_repair/network_view.ex
+++ b/lib/archethic/self_repair/network_view.ex
@@ -1,7 +1,7 @@
 defmodule Archethic.SelfRepair.NetworkView do
   @moduledoc """
   The network view is 2 things:
-  - the P2P view (list of available & authorized nodes)
+  - the P2P view (list of all nodes)
   - the Network chains view (oracle/origin/nodesharedsecrets)
 
   It is useful to compare with other nodes to detect desynchronization.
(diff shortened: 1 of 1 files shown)


lib/archethic/self_repair/network_view.ex
```diff
@@ -173,11 +211,29 @@ defmodule Archethic.SelfRepair.NetworkView do
...
-    |> Enum.map(& &1.last_public_key)
-    |> Enum.sort()
-    |> then(&:crypto.hash(:sha256, &1))
+
+  defp get_p2p_view() do
+    current_hash =
+      P2P.authorized_and_available_nodes()
+      |> Enum.map(& &1.last_public_key)
+      |> Enum.sort()
+      |> then(&:crypto.hash(:sha256, &1))
```
samuelmanzanera, line 220: I think you can refactor into a private function to DRY.

samuelmanzanera (approved):

Neylix:
Finally to simplify the workflow, we just use `P2P.list_nodes` to create the P2P so we don't care about authorized and available, we ensure we have the latest transaction of those chains

bchamagne:
I tested the following scenario and it work as expected: 

1. Start a node A, p2p_hash(A) = x
1. Start another node B, p2p_hash(A) = y & p2p_hash(B) = y
1. Start another node C,  p2p_hash(A) = z & p2p_hash(B) = z & p2p_hash(C) = z

But, now that I see it, we only detect changes when there is a new node. We do not change the hash if a node goes offline.

bchamagne (approved):

PR #1038, opened 2023-05-16
Improve lru read concurrency

Neylix:
# Description

Improve read concurrency on LRU cache. It now directly read the ets table instead of requesting it to the LRU GenServer. The GenServer is only used to manage the LRU workflow (monitor cache size, add value in ets table and remove old value when max size is reached).

Also move the ChainIndex LRU to be started in Embedded Supervisor instead of being linked to ChainIndex 

## Type of change

# How Has This Been Tested?

Unit test + manual test on lot of datas

# Checklist:

### DIFF
--- a/lib/archethic_cache/lru.ex
+++ b/lib/archethic_cache/lru.ex
@@ -25,9 +25,19 @@ defmodule ArchethicCache.LRU do
     GenServer.cast(server, {:put, key, value})
   end
 
-  @spec get(GenServer.server(), term()) :: nil | term()
-  def get(server, key) do
-    GenServer.call(server, {:get, key})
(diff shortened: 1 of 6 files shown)


lib/archethic_cache/lru.ex
```diff
@@ -26,8 +26,18 @@ defmodule ArchethicCache.LRU do
   end
 
   @spec get(GenServer.server(), term()) :: nil | term()
```
samuelmanzanera, line 28: We might indicate it can only be a `GenServer.name()` now

samuelmanzanera (approved):

bchamagne (approved):

PR #1040, opened 2023-05-16
Register an error during mining with invalid smart contract calls

### LINKED ISSUE #847: [SC] Support for error reporting in smart contracts

- wherever during the parsing/interpreting/executing of the contract , if there is an error raise an error
- We should provide better error handling

- we can use try and catch in contracts.ex for error reporting in
   - Action.execute
   - Contract.parse

### Additional context

_No response_

### LINKED ISSUE #881: Support smart contract error on transaction

Transaction's fee is a great way to prevent spamming of transactions is the rules are incorrect.
Sometimes, smart contract calls could be invalid in the case of insufficient fee or invalid conditions.
In that case, we have to support a way to handle the transaction—meaning paying the fees, while reporting it as invalid or incorrect.

The transaction's validation stamp support a field `error` that is only used for now to reject in the replication (if the validation nodes determine the pending transaction is invalid).
We can leverage this field to indicate other types of error which would be accepted by the chain replications nodes but not by the I/O storage nodes. 
Hence, on the explorer the transaction can be flagged as invalid, while reporting the reason.
So in the case of smart contract calls, we can determine the validity of the conditions and the fees, and still preventing the spam.

### Additional context

To handle such use cases, during transaction's validation we would have to fetch the smart contract destination addresses to be able to run the conditions or the applicable fees.
This would increase the transaction validation time, but the fees would have been paid within the transaction.

samuelmanzanera:
# Description

This PR provides code to validate smart contract calls during the mining to ensure the transaction in validation is respecting the SC condition of the targeted contracts in the recipients field

Fixes #881 

## Type of change

# How Has This Been Tested?

- Test A
- Test B

# Checklist:


lib/archethic/contracts/contract.ex
```diff
@@ -52,6 +52,26 @@ defmodule Archethic.Contracts.Contract do
     }
   end
 
+  @doc """
+  Create a contract from a transaction
+  """
+  @spec from_transaction(Transaction.t()) :: {:ok, t()} | {:error, String.t()}
+  def from_transaction(tx = %Transaction{data: %TransactionData{code: code}})
+      when is_binary(code) and code != "" do
```
bchamagne, line 60: this guard will throw, should this be an `{:error, _}` instead?

lib/archethic/contracts.ex
```diff
@@ -143,4 +143,25 @@ defmodule Archethic.Contracts do
...
+  """
+  @spec valid_contract_execution?(Contract.t(), Transaction.t(), list(Transaction.t())) ::
+          boolean()
+  def valid_contract_execution?(contract = %Contract{}, tx = %Transaction{}, calls \\ []) do
+    case Interpreter.execute(:transaction, contract, tx, [tx | calls]) do
+      {:ok, _} ->
+        true
+
+      _ ->
+        false
```
bchamagne, line 158: Do we really want to transform a "invalid condition constraint" into "invalid contract execution"?

test/archethic/mining/smart_contract_validation_test.exs
```diff
@@ -0,0 +1,148 @@
...
+
+      P2P.add_and_connect_node(node)
+
+      recipients = ["addr1", "addr2", "addr3"]
+      contracts = recipients |> SmartContractValidation.fetch_contracts() |> Enum.to_list()
+      assert length(contracts) == 3
+    end
+  end
+
+  describe "validate_remote_code/2" do
```
bchamagne, line 91: wrong naming

test/archethic/mining/smart_contract_validation_test.exs
```diff
@@ -0,0 +1,148 @@
...
+            actions triggered_by: transaction do
+              Contract.set_content "hello"
+            end
+            """
+          }
+        }
+        |> Contract.from_transaction!()
+
+      assert SmartContractValidation.valid_contract_calls?([contract], %Transaction{
+               data: %TransactionData{content: "hi"}
```
bchamagne, line 116: We'll probably have a failure here when #1020 is merged (interpreter will require a validated transaction (with a validationstamp))

lib/archethic/mining/smart_contract_validation.ex
```diff
@@ -0,0 +1,48 @@
...
+    end
+  end
+
+  @doc """
+  Determine if the smart contracts conditions are valid according to the given transaction
+  """
+  @spec valid_contract_calls?(Enumerable.t() | list(Contract.t()), Transaction.t()) :: boolean()
+  def valid_contract_calls?(contracts, transaction = %Transaction{}) do
+    TaskSupervisor
+    |> Task.Supervisor.async_stream_nolink(contracts, &Contracts.valid_contract_execution?(&1, transaction))
```
bchamagne, line 45: As discussed with @neylix, we believe we need to fetch the calls of the contract and pass them as 3rd argument.

lib/archethic/contracts/contract.ex
samuelmanzanera: Sure, as Interpreter.parse already handle the case.

lib/archethic/contracts.ex
samuelmanzanera: Why the invalid condition constraint will be raise ?

lib/archethic/mining/smart_contract_validation.ex
samuelmanzanera: Yes I was thinking about that, I didn't implement it because it will be an expensive operation. We need to figure out a way to assert a transaction without having to pull all the inputs of each destination chain.

lib/archethic/contracts.ex
samuelmanzanera: I think it's ok

bchamagne (changes requested):
Just a few things

bchamagne (approved):

PR #1047, opened 2023-05-23
Smart Contracts: Validation workflow refactor

### LINKED ISSUE #925: Smart Contracts: validation workflow

Currently, there are N storage nodes for the smart contract current transaction. When the contract is triggered, all of them execute the contract, but only the first one creates the "next transaction" (with a fallback mechanism if node is unavailable). 
Then the storages nodes of "next transaction" will check the inherit conditions.

We might want to check that all contracts executions resulted in the same "next transaction".

### Additional context

Epic: #306

bchamagne:
# Description
Validators used to only check the inherit condition. Now they actually execute the code of previous transaction and compare it to current transaction. This ensure the transaction is legitimate.

Fixes #925

## Type of change

# How Has This Been Tested?

Unit tested + Tested with 3 nodes.

# Checklist:

### DIFF
--- a/lib/archethic.ex
+++ b/lib/archethic.ex
@@ -4,16 +4,12 @@ defmodule Archethic do
   """
 
   alias Archethic.Account
-
   alias Archethic.BeaconChain
-
+  alias Archethic.Contracts
   alias Archethic.Contracts.Contract
   alias Archethic.Contracts.Interpreter
-
   alias Archethic.Crypto
-
   alias Archethic.Election
-
   alias Archethic.P2P
   alias Archethic.P2P.Node
   alias Archethic.P2P.Message
@@ -30,11 +26,8 @@ defmodule Archethic do
   alias Archethic.SelfRepair
   alias Archethic.SelfRepair.NetworkChain
   alias Archethic.SelfRepair.NetworkView
-
   alias Archethic.SharedSecrets
-
   alias Archethic.TaskSupervisor
-
   alias Archethic.TransactionChain
   alias Archethic.TransactionChain.Transaction
   alias Archethic.TransactionChain.TransactionInput
@@ -67,32 +60,37 @@ defmodule Archethic do
   @doc """
   Send a new transaction in the network to be mined. The current node will act as welcome node
   """
-  @spec send_new_transaction(Transaction.t()) :: :ok
+  @spec send_new_transaction(Transaction.t(), Crypto.key(), nil | Contract.Context.t()) :: :ok
   def send_new_transaction(
         tx = %Transaction{},
-        welcome_node_key \\ Crypto.first_node_public_key()
+        welcome_node_key \\ Crypto.first_node_public_key(),
+        contract_context \\ nil
       ) do
     if P2P.authorized_and_available_node?() do
       case NetworkChain.verify_synchronization(:node_shared_secrets) do
         :ok ->
-          do_send_transaction(tx, welcome_node_key)
+          do_send_transaction(tx, welcome_node_key, contract_context)
 
         :error ->
-          forward_transaction(tx, welcome_node_key)
+          forward_transaction(tx, welcome_node_key, contract_context)
 
         {:error, addresses} ->
           SharedSecrets.genesis_address(:node_shared_secrets) |> SelfRepair.resync(addresses, [])
 
-          forward_transaction(tx, welcome_node_key)
+          forward_transaction(tx, welcome_node_key, contract_context)
       end
     else
       # node not authorized
-      forward_transaction(tx, welcome_node_key)
+      forward_transaction(tx, welcome_node_key, contract_context)
     end
   end
 
-  @spec forward_transaction(tx :: Transaction.t(), welcome_node_key :: Crypto.key()) :: :ok
-  defp forward_transaction(tx, welcome_node_key) do
+  @spec forward_transaction(
+          tx :: Transaction.t(),
+          welcome_node_key :: Crypto.key(),
+          contract_context :: Contract.Context.t()
+        ) :: :ok
+  defp forward_transaction(tx, welcome_node_key, contract_context) do
     %Node{network_patch: welcome_node_patch} = P2P.get_node_info!(welcome_node_key)
 
     nodes =
@@ -118,7 +116,11 @@ defmodule Archethic do
     TaskSupervisor
     |> Task.Supervisor.start_child(fn ->
       :ok =
-        %NewTransaction{transaction: tx, welcome_node: welcome_node_key}
+        %NewTransaction{
+          transaction: tx,
+          welcome_node: welcome_node_key,
+          contract_context: contract_context
+        }
         |> do_forward_transaction(nodes)
     end)
 
@@ -137,7 +139,7 @@ defmodule Archethic do
 
   defp do_forward_transaction(_, []), do: {:error, :network_issue}
 
-  defp do_send_transaction(tx = %Transaction{type: tx_type}, welcome_node_key) do
+  defp do_send_transaction(tx = %Transaction{type: tx_type}, welcome_node_key, contract_context) do
     current_date = DateTime.utc_now()
     sorting_seed = Election.validation_nodes_election_seed_sorting(tx, current_date)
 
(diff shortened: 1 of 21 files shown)


lib/archethic.ex
```diff
@@ -54,32 +56,37 @@ defmodule Archethic do
   @doc """
   Send a new transaction in the network to be mined. The current node will act as welcome node
   """
-  @spec send_new_transaction(Transaction.t()) :: :ok
+  @spec send_new_transaction(Transaction.t(), Crypto.key(), nil | Contract.Context.t()) :: :ok
```
samuelmanzanera, line 59: I'm wondering if we should not move this function into the `Mining` context, so nested modules under `lib/archethic` will not call the top level `Archethic` but the siblings' context APIs.

lib/archethic/contracts.ex
```diff
@@ -40,98 +37,160 @@ defmodule Archethic.Contracts do
...
-       conditions: conditions
-     }} = Interpreter.parse(code)
-
-    with :ok <- validate_conditions(version, conditions, prev_tx, next_tx),
-         :ok <- validate_triggers(triggers, next_tx, date) do
-      true
+    with :ok <- validate_trigger(trigger_type, trigger, timestamp),
+         {:ok, contract} <- Interpreter.parse_transaction(prev_tx),
+         {:ok, calls} <- TransactionChain.fetch_contract_calls(previous_address) do
+      case execute_trigger(
```
samuelmanzanera, line 80: maybe we can leverage `with` statements here.

lib/archethic/contracts.ex
```diff
@@ -40,98 +37,160 @@ defmodule Archethic.Contracts do
...
+    now = DateTime.utc_now()
 
-        if Interpreter.valid_conditions?(version, inherit_conditions, constants) do
-          :ok
-        else
-          Logger.error("Inherit constraints not respected")
-          {:error, :invalid_inherit_constraints}
-        end
+    # Accept time drifting for 10seconds
+    if DateTime.diff(trigger_datetime, datetime) >= 0 and
```
samuelmanzanera, line 153: Can this be refactored as it's used below in `validate_trigger({:interval, interval}` ?

lib/archethic/contracts.ex
```diff
@@ -151,24 +210,44 @@ defmodule Archethic.Contracts do
...
+         datetime
+       ) do
+    %{
+      "transaction" => Constants.from_transaction(transaction),
+      "contract" => contract_constant,
+      "_time_now" => DateTime.to_unix(datetime)
+    }
+  end
+
+  defp get_condition(:oracle, %Contract{conditions: conditions}),
```
samuelmanzanera, line 245: Don't you think this can be simplied over
```elixir
defp get_condition(condition, %Contract{conditions: conditions}) when condition in [:oracle, :transaction, :inherit] do
    Map.get(conditions, condition)
end
```

lib/archethic.ex
samuelmanzanera: Referenced in #1106

lib/archethic/contracts.ex
```diff
@@ -40,98 +39,157 @@ defmodule Archethic.Contracts do
...
-  defp valid_from_trigger?(
-         {:datetime, datetime},
-         %Transaction{},
-         validation_date = %DateTime{}
-       ) do
-    # Accept time drifing for 10seconds
-    DateTime.diff(validation_date, datetime) >= 0 and
-      DateTime.diff(validation_date, datetime) < 10
+  defp validate_trigger(:transaction, {:transaction, _address}, _) do
+    # maybe check address exist and contain the contract in the recipients?
```
Neylix, line 176: indeed this could be interesting. The function `validate_trigger` could return  `{:ok, maybe_tx}` and in case of oracle or transaction trigger fetch the transaction and verify if it's recipient (resolved address in validation stamp) is the previous contract address.

lib/archethic/contracts/interpreter.ex
```diff
@@ -153,18 +137,39 @@ defmodule Archethic.Contracts.Interpreter do
 
       trigger_code ->
         timestamp_now =
-          time_now(trigger_type, maybe_trigger_tx)
+          case Keyword.get(opts, :time_now) do
+            nil ->
+              # you must use the :time_now opts during the validation workflow
+              # because there is no validation_stamp yet
+              time_now(trigger_type, maybe_trigger_tx)
```
Neylix, line 144: This could be as the default value of `Keyword.get` : `Keyword.get(opts, :time_now, time_now(trigger_type, maybe_trigger_tx))`

lib/archethic/mining.ex
```diff
@@ -36,23 +37,33 @@ defmodule Archethic.Mining do
   @spec start(
           transaction :: Transaction.t(),
           welcome_node_public_key :: Crypto.key(),
-          validation_node_public_keys :: list(Crypto.key())
+          validation_node_public_keys :: list(Crypto.key()),
+          contract_context :: nil | Contract.Context.t()
         ) :: {:ok, pid()}
```
Neylix, line 42: contract_context is always passed to the Mining.start function so the default value is useless I think

lib/archethic/contracts/contract/context.ex
```diff
@@ -0,0 +1,33 @@
+defmodule Archethic.Contracts.Contract.Context do
```
Neylix, line 1: I didn't find any serialization / deserialization of the Context used in NewTransaction / StartMining message

Neylix (changes requested):

lib/archethic/contracts/interpreter.ex
bchamagne: nope it's not possible because the `time_now` cannot be executed if there is no `validation_stamp`. It'd raise.

lib/archethic/contracts/contract/context.ex
bchamagne: wow! this means all my tests were wrong and useless. I'll add missing tests, fix this and retest everything

lib/archethic/p2p/message/start_mining.ex
```diff
@@ -96,12 +100,23 @@ defmodule Archethic.P2P.Message.StartMining do
         welcome_node_public_key: welcome_node_public_key,
         validation_node_public_keys: validation_node_public_keys,
         network_chains_view_hash: network_chains_view_hash,
-        p2p_view_hash: p2p_view_hash
+        p2p_view_hash: p2p_view_hash,
+        contract_context: contract_context
       }) do
+    serialized_contract_context =
+      case contract_context do
```
Neylix, line 107: This condition is often checked, I think it could be easier to accept that a Contract.Context can be nil, so the functions `Contract.Context.serialize` or `deserialize` could return a specific value in case of nil context
```elixir
def serialize(nil), do: <<0::8>>
def serialize(context) do <<1::8, ...>> end

def deserialize(<<0::8, rest::binary>>), do: {nil, rest}
def deserialize(<<1::8, rest::binary>>) do ... end
```

Neylix (changes requested):

lib/archethic/contracts.ex
```diff
@@ -40,98 +41,184 @@ defmodule Archethic.Contracts do
...
 
-  defp valid_from_trigger?(
-         {:datetime, datetime},
-         %Transaction{},
-         validation_date = %DateTime{}
-       ) do
-    # Accept time drifing for 10seconds
-    DateTime.diff(validation_date, datetime) >= 0 and
-      DateTime.diff(validation_date, datetime) < 10
+  defp validate_trigger({:transaction, address}, _validation_datetime, contract_address) do
```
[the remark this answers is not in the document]
bchamagne: as you mentionned, if the transaction exists, it passed the "validate_recipients" step which already check the `condition transaction`. Because when we fetch the transaction, we ask a quorum of nodes, we can trust that without rerun the condition.

lib/archethic/p2p/message/start_mining.ex
samuelmanzanera: What's the point of serialize an nil structure ?
It should be handled upfront.

bchamagne: It's in case the context is not the last part of a struct that contains the context (StartMining/NewTransaction)

samuelmanzanera: Just need to put a bit in the newTx and startling, like in the transaction struct with validation stamp

bchamagne: I reverted to what  it used to be

samuelmanzanera (approved):

PR #1097, opened 2023-06-14
Transform the LRU cache to use ETS with pointer to head/tail

### LINKED ISSUE #1089: LRU: remove the usage of List.delete

List.delete is a slow operation. We can easily have millions of data in the list.

We probably need a different data structure. Maybe a Set? 
Benchmarks are required to determine the best solution.

### Epic

_No response_

samuelmanzanera:
This remove the need to have a list of all values
and have to implement costly move key to front or eviction.

The implementation will leverage mainly key value & atomic counter operations.

Fixes #1089

## Type of change

- Enhancements

# Checklist:

### DIFF
--- a/lib/archethic_cache/lru.ex
+++ b/lib/archethic_cache/lru.ex
@@ -1,64 +1,137 @@
 defmodule ArchethicCache.LRU do
   @moduledoc """
-  A cache that stores the values in an ETS table.
-  There are hooks available to be able to add effects (ex: write to disk).
+  A cache that stores the values in an ETS table using Last Recent Used strategy (first is the most recent used, last is the least recent used)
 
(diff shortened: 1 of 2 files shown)


bchamagne (approved):

Neylix:
We have to take care about the number of LRU we will start.
As the now a LRU starts ets table, we cannot start as many LRU as we want since the number of ets table is limited

samuelmanzanera:
> We have to take care about the number of LRU we will start.
> As the now a LRU starts ets table, we cannot start as many LRU as we want since the number of ets table is limited



According to the doc this is not anymore the case. But even with 1400 tables this would give a lot LRUs to create.

PR #1100, opened 2023-06-15
Refactor P2P.connected? into Connection module

Neylix:
# Description

Refactor to be ready for hot reload

# Checklist:

### DIFF
--- a/lib/archethic/p2p/supervisor.ex
+++ b/lib/archethic/p2p/supervisor.ex
@@ -11,11 +11,20 @@ defmodule Archethic.P2P.Supervisor do
 
   use Supervisor
 
+  @table_name :connection_status
+
   def start_link(args \\ []) do
     Supervisor.start_link(__MODULE__, args, name: Archethic.P2PSupervisor)
   end
 
   def init(_args) do
+    # Start an ets table to manage node connection status.
+    # This reduce the number of message sent to the Connection GenServer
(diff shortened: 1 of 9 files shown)


bchamagne (approved):

lib/archethic/p2p/supervisor.ex
```diff
@@ -11,11 +11,20 @@ defmodule Archethic.P2P.Supervisor do
 
   use Supervisor
 
+  @table_name :connection_status
```
samuelmanzanera, line 14: Maybe you should use a function in the Connection module to give the name of the table.

lib/archethic/p2p/supervisor.ex
```diff
@@ -11,11 +11,20 @@ defmodule Archethic.P2P.Supervisor do
...
+
   def start_link(args \\ []) do
     Supervisor.start_link(__MODULE__, args, name: Archethic.P2PSupervisor)
   end
 
   def init(_args) do
+    # Start an ets table to manage node connection status.
+    # This reduce the number of message sent to the Connection GenServer
+    if :ets.whereis(@table_name) == :undefined do
```
samuelmanzanera, line 23: If we want to be cleaner, the best would be to isolate the table in a genserver which might serve more capability and will support hot reload, if we need to change the state in the ets table easily. 
This gen server could be dedicated to connection information or status and might manage its own table and lifecycle.

PR #1102, opened 2023-06-18
Bugfix : CSS assets are not available in the docker image

zedalaye:
Seems because app folder is copied over the built assets.

Looks like COPYing the app folder once is sufficient to be able to build everything after


bchamagne:
LGTM, I wasn't able to test because I am having an other issue on docker (some missing lib). I'll try to test it tomorrow.

zedalaye:
Just pushed a rebase over the current develop branch.

PR #1135, opened 2023-06-30
Add new option to forward transaction

### LINKED ISSUE #1134: Do not forward a transaction if there is a responsiveness

Basically we only want to forward_transaction in these cases:

- new transaction from the API
- new transaction from the bootstrap
- networking scheduler

_No response_

### Epic

_No response_

Neylix:
# Description

Fixes an issue where a transaction which is forwarded because validation nodes are not responding or returns an synchronization, can create a loop of death between node if this transaction is also catched by node responsivness.

Now all transaction created by GenServer that also spawn a node responsivness, are not forwarded since node responsivness will do it

Fixes #1134 

## Type of change

# How Has This Been Tested?

Unit test

# Checklist:

### DIFF
--- a/lib/archethic.ex
+++ b/lib/archethic.ex
@@ -9,6 +9,7 @@ defmodule Archethic do
   alias Archethic.Contracts.Contract
   alias Archethic.Crypto
   alias Archethic.Election
+  alias Archethic.Mining
   alias Archethic.P2P
   alias Archethic.P2P.Node
   alias Archethic.P2P.Message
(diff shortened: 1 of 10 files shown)


lib/archethic.ex
```diff
@@ -59,104 +60,49 @@ defmodule Archethic do
...
-    nodes =
-      if this_node != welcome_node_key do
-        #  if this node is not the welcome node then select
-        # next node from the this node position in nodes list
-        index = Enum.find_index(nodes, &(&1.first_public_key == this_node))
-        {_l, r} = Enum.split(nodes, index + 1)
-        r
-      else
-        nodes
+      if should_forward_transaction?(forward?, responses, length(validation_nodes)) do
```
samuelmanzanera, line 76: We can probably refactor it to avoid this strange thing with `forward?` `should_forward_transaction?`, something like: `  forward? = Keyword.get(opts, :forward?, should_forward_transaction?(responses, length(validation_nodes)))`

samuelmanzanera, line 76: Here a suggestion to refactor a bit to simplify the forward should_forward_transaction

```suggestion
      forward? = case Keyword.get(opts, :forward?) do
        true ->
          should_forward_transaction?(responses, length(validation_nodes))
        _ ->
          false
      end
      
      maybe_start_resync(responses)

      if forward? do
        forward_transaction(tx, welcome_node_key, contract_context)
      end
```

PR #1141, opened 2023-07-03
Fix stats adding multiple times same summary

Neylix:
# Description

Fixes a problem where is was possible that the stats add several times the same beacon summary.

Also enhance the stats file, since only the last summary stats are used we don't need to store each sats for each summary

## Type of change

# Checklist:


priv/migration_tasks/prod/1.2.4@stats_info.ex
```diff
@@ -0,0 +1,45 @@
+defmodule Migration_1_2_4 do
```
bchamagne, line 1: Should this be 1.3.0 instead?

bchamagne (approved):

PR #1148, opened 2023-07-04
Optimize context retrieving

Neylix:
# Description

On mainnet we can see that some cross validation nodes do not have the time to send the mining context to the coordinator before the 3ms of timeout.
This can be explained due to high latency but also the way a node retrieve the context.
Currently a node fetch the previous transaction and the p2p view concurrently and then fetch the utxos.
This is improved by fetching the utxos concurrently with other fetch.

Also fixed the timestamp of transaction fetching

# Checklist:

### DIFF
--- a/lib/archethic/mining/transaction_context.ex
+++ b/lib/archethic/mining/transaction_context.ex
@@ -58,10 +58,10 @@ defmodule Archethic.Mining.TransactionContext do
       |> Enum.uniq()
 
     prev_tx_task = request_previous_tx(previous_address, prev_tx_nodes_split)
+    utxos_task = request_utxos(previous_address, unspent_outputs_nodes_split)
     nodes_view_task = request_nodes_view(node_public_keys)
 
-    prev_tx = Task.await(prev_tx_task)
-    nodes_view = Task.await(nodes_view_task)
+    [prev_tx, utxos, nodes_view] = Task.await_many([prev_tx_task, utxos_task, nodes_view_task])
 
     %{
       chain_nodes_view: chain_storage_nodes_view,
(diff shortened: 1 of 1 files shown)


bchamagne (approved):

lib/archethic/mining/transaction_context.ex
```diff
@@ -58,9 +58,11 @@ defmodule Archethic.Mining.TransactionContext do
       |> Enum.uniq()
 
     prev_tx_task = request_previous_tx(previous_address, prev_tx_nodes_split)
+    utxos_task = request_utxos(previous_address, unspent_outputs_nodes_split)
     nodes_view_task = request_nodes_view(node_public_keys)
 
     prev_tx = Task.await(prev_tx_task)
+    utxos = Task.await(utxos_task)
```
samuelmanzanera, line 65: You could probably use [Task.await_many](https://hexdocs.pm/elixir/Task.html#await_many/2)

Neylix: Done !

lib/archethic/mining/transaction_context.ex
```diff
@@ -98,23 +91,26 @@ defmodule Archethic.Mining.TransactionContext do
...
-
-          {:error, _} ->
-            nil
-        end
+    Task.Supervisor.async(TaskSupervisor, fn ->
+      # Timeout of 2.5 sec because the coordinator node wait 3 sec to get the context
+      # from the cross validation nodes
+      case TransactionChain.fetch_transaction(previous_address, nodes,
+             search_mode: :remote,
+             timeout: 2500
```
samuelmanzanera, line 99: I'm wondering if it should be based on the variable of the coordinator waiting time to be more dynamic.

Neylix: Deleted the commit, we'll need to rework globally on mining timeout

samuelmanzanera (approved):

Issue #980, opened 2023-04-17
Ensure coordinator node and cross validation nodes are different


Neylix:
### Describe the problem you discovered

When a replication node verify the validation stamps and cross validation stamps signature, it does not ensure the coordinator node and the cross validation nodes are different.
This can be a breach for a malicious node to send a self signed transaction to storage nodes while it should be signed through the distributed workflow with multiple nodes.

### Describe the solution you'd like

In the function `Transaction.valid_stamps_signature?` we should ensure the coordinator node is different from the cross validation nodes.
This check should be done only if the validation node list has at least 2 nodes

PR #1150, opened 2023-07-05
WIP - Overbook validation nodes

### LINKED ISSUE #913: Open more nodes in validation nodes election

Actually for a low UCO transfer amount we have 3 validation nodes. The minimum number of nodes to validate a transaction is 2.
Let's say there is 2 nodes in the network which suddenly disconnect but are still globally available.
If an election of a transaction select 3 nodes including the 2 disconnected one, the transaction will never be validated until theses node will be globally unavailable in the next (or next next) self repair. So the transaction is stuck for a long time and this can be really annoying if it's a network one.

Here is a possible solution:
When a welcome node receive a transaction, it do the election (assume 3 nodes) and 2 of them are locally unavailable (they are maybe disconnected) we know at this time that the transaction will not be validated (only 1 or 0 validation nodes will receive the StartMining message).

Actually the maximum number of validation nodes is 12. So the election may always return 12 nodes and then the welcome node select the needed number of validation node out of these 12 nodes, but only locally available nodes. (This selection may be random ?)
Then for the replication and the validation node, to determine if the election is right, the welcome node can send the list of the 12 nodes from the election (that can be verified using the same election) and the selected one to validate the transaction. The selected ones should be part of the list of the 12 nodes.

With this behavior, we keep the same verification for the election, but we have more flexibility to select locally available nodes to validate the transaction. And do not let transaction waiting for next self repair.

samuelmanzanera:
# Description

This aims to resolve issue regarding the number of valiation nodes which not be availalbe during the transaction's validation and the election chose them.

This proposes an overbooking system for the validations based on the minimum validations nodes and to filter/jump nodes based on the election keeping the sorted algorithms and heuristic constraints.

Fixes #913, #980 

Some changes are still required in that PR to improve the validation election verification

## Type of change

- Enhancement

# How Has This Been Tested?

- Unit test
- Integration test

Use the following patch to use deterministic validation node to help determine the validation to shut down

```
diff --git a/lib/archethic/election.ex b/lib/archethic/election.ex
index 465b55bb..d5d3cf03 100755
--- a/lib/archethic/election.ex
+++ b/lib/archethic/election.ex
@@ -22,29 +22,34 @@ defmodule Archethic.Election do
   Create a seed to sort the validation nodes. This will produce a proof for the election
   """
   @spec validation_nodes_election_seed_sorting(Transaction.t(), DateTime.t()) :: binary()
-  def validation_nodes_election_seed_sorting(tx = %Transaction{}, timestamp = %DateTime{}) do
-    tx_hash =
-      tx
-      |> Transaction.to_pending()
-      |> Transaction.serialize()
-      |> Crypto.hash()
-
-    Crypto.sign_with_daily_nonce_key(tx_hash, timestamp)
+  def validation_nodes_election_seed_sorting(_tx = %Transaction{}, _timestamp = %DateTime{}) do
+    # tx_hash =
+    #   tx
+    #   |> Transaction.to_pending()
+    #   |> Transaction.serialize()
+    #   |> Crypto.hash()
+
+    # Crypto.sign_with_daily_nonce_key(tx_hash, timestamp)
+    <<195, 51, 61, 55, 140, 12, 138, 246, 249, 106, 198, 175, 145, 9, 255, 133, 67,
+    240, 175, 53, 236, 65, 151, 191, 128, 11, 58, 103, 82, 6, 218, 31, 220, 114,
+    65, 3, 151, 209, 9, 84, 209, 105, 191, 180, 156, 157, 95, 25, 202, 2, 169,
+    112, 109, 54, 99, 40, 47, 96, 93, 33, 82, 40, 100, 13>>
   end
 
   @doc """
   Verify if a proof of election is valid according to transaction and the given public key
   """
   @spec valid_proof_of_election?(Transaction.t(), binary, Crypto.key()) :: boolean
-  def valid_proof_of_election?(tx = %Transaction{}, proof_of_election, daily_nonce_public_key)
-      when is_binary(proof_of_election) and is_binary(daily_nonce_public_key) do
-    data =
-      tx
-      |> Transaction.to_pending()
-      |> Transaction.serialize()
-      |> Crypto.hash()
-
-    Crypto.verify?(proof_of_election, data, daily_nonce_public_key)
+  def valid_proof_of_election?(_tx = %Transaction{}, _proof_of_election, _daily_nonce_public_key) do
+      # when is_binary(proof_of_election) and is_binary(daily_nonce_public_key) do
+    # data =
+    #   tx
+    #   |> Transaction.to_pending()
+    #   |> Transaction.serialize()
+    #   |> Crypto.hash()
+
+    # Crypto.verify?(proof_of_election, data, daily_nonce_public_key)
+    true
   end
 
   @doc """
```

You should run at least 6 nodes because the overbooking will try to get 5 nodes by default based on the min validation which is 3.

# Checklist:

### DIFF
--- a/lib/archethic/election/constraints/validation.ex
+++ b/lib/archethic/election/constraints/validation.ex
@@ -98,7 +98,7 @@ defmodule Archethic.Election.ValidationConstraints do
           validation_number
       end
     else
-      min_nb_nodes
+      default_nb_nodes
     end
   end
 
--- a/lib/archethic/election/constraints/validation.ex
+++ b/lib/archethic/election/constraints/validation.ex
@@ -92,7 +92,7 @@ defmodule Archethic.Election.ValidationConstraints do
           nb_authorized_nodes
 
         validation_number < default_nb_nodes ->
-          min_nb_nodes
+          default_nb_nodes
 
         true ->
           validation_number
--- a/lib/archethic/election/constraints/validation.ex
+++ b/lib/archethic/election/constraints/validation.ex
@@ -17,6 +17,8 @@ defmodule Archethic.Election.ValidationConstraints do
   alias Archethic.TransactionChain.TransactionData.Ledger
   alias Archethic.TransactionChain.TransactionData.UCOLedger
 
+  alias Archethic.Utils
+
   @typedoc """
   Each validation constraints represent a function which will be executed during the election algorithms:
   - min_geo_patch: Require number of distinct geographic patch for the elected validation nodes.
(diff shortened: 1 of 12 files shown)


lib/archethic/election/constraints/validation.ex
```diff
@@ -66,30 +68,41 @@ defmodule Archethic.Election.ValidationConstraints do
...
-          min_validation_nodes(nb_authorized_nodes)
+        validation_number < default_nb_nodes ->
+          min_nb_nodes
 
         true ->
           validation_number
       end
     else
-      min_validation_nodes(nb_authorized_nodes)
+      min_nb_nodes
```
Neylix, line 101: This should be the default_nb_nodes

PR #1180, opened 2023-07-19
use PlugCrypto binary to term in bootstrap.info

apoorv-2204:
though its not user-data , but still.

### DIFF
--- a/lib/archethic/db/embedded_impl/bootstrap_info.ex
+++ b/lib/archethic/db/embedded_impl/bootstrap_info.ex
@@ -47,6 +47,6 @@ defmodule Archethic.DB.EmbeddedImpl.BootstrapInfo do
 
   defp load_data(filepath) do
     data = File.read!(filepath)
-    :erlang.binary_to_term(data, [:safe])
+    Plug.Crypto.non_executable_binary_to_term(data, [:safe])
   end
 end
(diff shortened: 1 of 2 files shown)


bchamagne (approved):
I am fine with this, maybe we can add a Credo rules to forbid the usage of `binary_to_term` alltogether.

apoorv-2204:
> maybe we can add a Credo rules to forbid

Credo do has high priority warning, but as of now we are just ignoring it.

Neylix:
If there is configuration that ignore this check, we should not ignore it. Could you update the conf to ensure credo raise a warning if there is a binary to term used ?

apoorv-2204:
> If there is configuration that ignore this check, we should not ignore it. Could you update the conf to ensure credo raise a warning if there is a binary to term used ?

There are also other critical issues reported by credo,  aka priority high . Exiting the credo at priority high will eventually crash as they are not fixed yet

Neylix:
> > If there is configuration that ignore this check, we should not ignore it. Could you update the conf to ensure credo raise a warning if there is a binary to term used ?
> 
> There are also other critical issues reported by credo, aka priority high . Exiting the credo at priority high will eventually crash as they are not fixed yet

Do you mean Credo or Sobelow ?

apoorv-2204:
> > > If there is configuration that ignore this check, we should not ignore it. Could you update the conf to ensure credo raise a warning if there is a binary to term used ?
> > 
> > 
> > There are also other critical issues reported by credo, aka priority high . Exiting the credo at priority high will eventually crash as they are not fixed yet
> 
> Do you mean Credo or Sobelow ?

credo

PR #1187, opened 2023-07-25
Implement json rpc

### LINKED ISSUE #1169: Implement JSON RPC server
We are currently using a REST API to push or request computation . However the REST approach is not practical. 

JSON RPC is well known and used in a lot of distributed ledger.
It's advantage relies on request which are performing operations instead of access and modify resources (used in a REST API). 

Hence this would make sense to design, structure and create a more expressive API for clients applications. 

So the idea would be to transform the current API methods into JsonRpc:
- [x] Send transaction
- [x] Estimate fee
- [x] Simulate SC call
- [x] Add new origin key

We should ensure the clients & SDK support the new queries. 
The nodes  should support the both API until the deprecation policy is applied. 

- [ ] A new entry in the documentation should be present to list the available functions. 

--

This refers to the [AEIP17](https://github.com/archethic-foundation/aeip/blob/main/AEIP-17.md)

Neylix:
# Description

Implement new API using JSON RPC specification.
Did not use any library as I didn't found one that works well with our implementation.

Used json schema ton verify request is compliant with json rpc specification
Implement all existing rest API

Fixes #1169 

## Type of change

# How Has This Been Tested?

Unit test
Manual test with postman

# Checklist:


lib/archethic_web/api/jsonrpc/controllers/jsonrpc_controller.ex
```diff
@@ -0,0 +1,135 @@
...
+    )
+    |> Enum.zip(requests)
+    |> Enum.map(fn
+      {{:ok, {:ok, result}}, request} ->
+        create_valid_response(request, result)
+
+      {{:ok, {:error, reason}}, request} ->
+        create_error_response(request, reason)
+
+      {{:error, :timeout}, request} ->
```
bchamagne, line 66: I think this should be `{:exit, :timeout}`. 

> :kill_task - the task that timed out is killed. The value emitted for that task is {:exit, :timeout}.

bchamagne (approved):

lib/archethic_web/api/jsonrpc/controllers/jsonrpc_controller.ex
Neylix: Updated !
Also added Logger.warning for other exit reason

aime-risson (approved):

PR #1189, opened 2023-07-26
Add Chain.get_transaction/1

### LINKED ISSUE #1167: Being able to fetch transaction from smart contract

Sometimes we need to fetch another transaction within a smart contract, the interpreter should expose a method for this purpose.

The interpreter could expose a function in the `Chain` module as: `Chain.get_transaction` which will do a quorum request to fetch the transaction to accessible in the contract actions.

### Additional context

_No response_

### Epic

_No response_

bchamagne:
# Description

Implement `Chain.get_transaction/1`
Fixes #1167

- [ ] Add mock on the SCPlayground
- [x] Documentation

## Type of change

# How Has This Been Tested?

Unit tested

# Checklist:

### DIFF
--- a/lib/archethic/contracts/interpreter/library/common/chain_impl.ex
+++ b/lib/archethic/contracts/interpreter/library/common/chain_impl.ex
@@ -3,6 +3,8 @@ defmodule Archethic.Contracts.Interpreter.Library.Common.ChainImpl do
   @behaviour Archethic.Contracts.Interpreter.Library.Common.Chain
 
   alias Archethic.Contracts.Interpreter.Legacy
+  alias Archethic.Contracts.Interpreter.Legacy.UtilsInterpreter
+  alias Archethic.Contracts.ContractConstants, as: Constants
 
   @impl Archethic.Contracts.Interpreter.Library.Common.Chain
   defdelegate get_genesis_address(address),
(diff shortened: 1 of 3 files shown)


lib/archethic/contracts/interpreter/library/common/chain_impl.ex
```diff
@@ -18,4 +20,18 @@ defmodule Archethic.Contracts.Interpreter.Library.Common.ChainImpl do
...
+  def get_transaction(address) do
+    address
+    |> UtilsInterpreter.get_address("Chain.get_transaction/1")
+    |> Archethic.search_transaction()
+    |> then(fn
+      {:ok, tx} ->
+        Constants.from_transaction(tx)
+
+      {:error, _} ->
+        nil
```
Neylix, line 34: Currently other `Chain` functions raise an error if the result is invalid. In this case you are returning nil which is not coherent with other functions.
We should have an unified way to return error.

Neylix (changes requested):

lib/archethic/contracts/interpreter/library/common/chain_impl.ex
bchamagne: - [ ] Update all `Chain` functions to return `nil` instead of crashing

bchamagne: - [ ] expect instead of stub

Neylix (approved):

PR #1199, opened 2023-07-27
Scope refactor

aime-risson:
# Description

This pr refactores the Scope module so that the scope is entirely handled by it. 

Fixes # (issue)

## Type of change

non-breaking change

# How Has This Been Tested?

New tests for the scope module have been added

# Checklist:

### DIFF
--- a/lib/archethic/contracts/interpreter/scope.ex
+++ b/lib/archethic/contracts/interpreter/scope.ex
@@ -27,14 +27,14 @@ defmodule Archethic.Contracts.Interpreter.Scope do
     current_scope_hierarchy = get_context_scope_hierarchy(current_context)
     ref = new_ref()
 
-    Process.put(
-      :scope,
-      put_in(Process.get(:scope), [current_context] ++ current_scope_hierarchy ++ [ref], %{})
-    )
+    new_scope =
+      Process.get(:scope)
+      |> put_in([current_context] ++ current_scope_hierarchy ++ [ref], %{})
+      |> update_in([current_context, "scope_hierarchy"], &(&1 ++ [ref]))
 
     Process.put(
       :scope,
-      update_in(Process.get(:scope), [current_context, "scope_hierarchy"], &(&1 ++ [ref]))
+      new_scope
     )
 
     :ok
--- a/lib/archethic/contracts/interpreter/scope.ex
+++ b/lib/archethic/contracts/interpreter/scope.ex
@@ -3,11 +3,7 @@ defmodule Archethic.Contracts.Interpreter.Scope do
   Helper functions to deal with scopes
   """
 
-  @doc """
-  Initialize the scope with given map
-  """
-  @spec init(map()) :: :ok
-  def init(global_variables \\ %{}) do
+  defp init(global_variables) do
     global_variables = Map.put(global_variables, "context_list", [])
 
     Process.put(
--- a/lib/archethic/contracts/interpreter/scope.ex
+++ b/lib/archethic/contracts/interpreter/scope.ex
@@ -68,15 +68,15 @@ defmodule Archethic.Contracts.Interpreter.Scope do
   @spec leave_scope() :: :ok
   def leave_scope() do
     current_context = get_current_context()
+    current_scope_hierarchy = get_context_scope_hierarchy(current_context)
 
-    Process.put(
-      :scope,
-      update_in(
-        Process.get(:scope),
-        [current_context, "scope_hierarchy"],
-        &List.delete_at(&1, -1)
-      )
-    )
+    new_scope =
+      Process.get(:scope)
+      |> update_in([current_context, "scope_hierarchy"], &List.delete_at(&1, -1))
+      |> pop_in([current_context] ++ current_scope_hierarchy)
+      |> elem(1)
+
+    Process.put(:scope, new_scope)
 
     :ok
   end
--- a/lib/archethic/contracts/interpreter/scope.ex
+++ b/lib/archethic/contracts/interpreter/scope.ex
@@ -91,7 +91,7 @@ defmodule Archethic.Contracts.Interpreter.Scope do
     new_scope =
       Process.get(:scope)
       |> Map.delete(current_context)
-      |> Map.update!("context_list", &List.delete_at(&1, 0))
+      |> Map.update!("context_list", fn [_first | rest] -> rest end)
 
     Process.put(:scope, new_scope)
 
--- a/lib/archethic/contracts/interpreter/scope.ex
+++ b/lib/archethic/contracts/interpreter/scope.ex
@@ -8,6 +8,8 @@ defmodule Archethic.Contracts.Interpreter.Scope do
   """
   @spec init(map()) :: :ok
   def init(global_variables \\ %{}) do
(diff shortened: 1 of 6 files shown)


lib/archethic/contracts/interpreter/scope.ex
```diff
@@ -19,27 +21,109 @@ defmodule Archethic.Contracts.Interpreter.Scope do
...
+    Process.put(:scope, new_scope)
+
+    :ok
+  end
+
+  @doc """
+  Leave a scope by removing it from current context's scope hierarchy and deleting its content
+  """
+  @spec leave_scope() :: :ok
+  def leave_scope() do
```
Neylix, line 69: Here you only delete the ref in the scope hierarchy list. You could also delete the ref in the nested scope map, it would behave a bit like a garbage collector

lib/archethic/contracts/interpreter/scope.ex
```diff
@@ -19,27 +21,109 @@ defmodule Archethic.Contracts.Interpreter.Scope do
...
+  Leave a context by removing it from scope and context_list
+  """
+  @spec leave_context() :: :ok
+  def leave_context() do
+    current_context = get_current_context()
+
+    new_scope =
+      Process.get(:scope)
+      |> Map.delete(current_context)
+      |> Map.update!("context_list", &List.delete_at(&1, 0))
```
Neylix, line 94: Here you could replace `&List.delete_at(&1, 0)` by pattern matching like `fn [_first | reset] -> rest end`

lib/archethic/contracts/interpreter/scope.ex
```diff
@@ -8,6 +8,8 @@ defmodule Archethic.Contracts.Interpreter.Scope do
   """
   @spec init(map()) :: :ok
   def init(global_variables \\ %{}) do
```
Neylix, line 10: Scope.init is not used anymore outside of Scope, you can make it private

test/archethic/contracts/interpreter/scope_test.exs
```diff
@@ -5,55 +5,589 @@ defmodule Archethic.Contracts.Interpreter.ScopeTest do
...
+      constants = %{"var1" => %{"prop" => 1}, "var2" => 4}
+      Scope.init(constants)
+      assert %{"context_list" => [], "var1" => %{"prop" => 1}, "var2" => 4} == Process.get(:scope)
+    end
   end
 
-  test "write_at/3" do
-    Scope.init()
+  describe "create_context/0" do
+    test "create_context/0 should create empty context to existing scope" do
```
Neylix, line 17: In test title you don't need to repeat the describe title since the log test concatenate both

test/archethic/contracts/interpreter/scope_test.exs
```diff
@@ -5,55 +5,589 @@ defmodule Archethic.Contracts.Interpreter.ScopeTest do
...
-    Scope.create(hierarchy)
-    Scope.write_at(hierarchy, "var1", 1)
-    assert Scope.read(hierarchy, "var1") == 1
+      Scope.create_context()
 
-    hierarchy = ["depth1", "depth2"]
-    Scope.create(hierarchy)
-    Scope.write_at(hierarchy, "var2", 2)
-    assert Scope.read(hierarchy, "var2") == 2
+      %{"context_list" => [context_ref]} = Process.get(:scope)
```
Neylix, line 22: You can add an assert here in case "context_list" is empty

test/archethic/contracts/interpreter/scope_test.exs
```diff
@@ -5,55 +5,589 @@ defmodule Archethic.Contracts.Interpreter.ScopeTest do
...
+        "var1" => %{"prop" => 1},
+        "var2" => 4,
+        "context_1" => %{
+          "scope_hierarchy" => []
+        }
+      })
+
+      Scope.create_context()
+
+      %{"context_list" => [context_ref_2, _]} = Process.get(:scope)
```
Neylix, line 47: Same here for assert.
You can also match that second arg of the list is the first context 
```elixir
assert %{"context_list" => [context_ref_2, "context_1"]} = Process.get(:scope)
```

test/archethic/contracts/interpreter/scope_test.exs
```diff
@@ -5,55 +5,589 @@ defmodule Archethic.Contracts.Interpreter.ScopeTest do
...
+            "second_scope" => %{}
+          }
+        }
+      }
+
+      Process.put(:scope, scope)
+
+      Scope.leave_scope()
+
+      assert %{
```
Neylix, line 200: Here the test is not fully right as the function `leave_scope` does not delete the scope in the map, but remove only the ref in scope hierarchy.
The test is passing because you pattern match the result. Maybe you should not pattern match but use a strict equal `==` so you ensure the map is the expected one without any other data inside.
You could do it for all tests 

(Tested with `==` test is not passing)

test/archethic/contracts/interpreter/scope_test.exs
```diff
@@ -5,55 +5,589 @@ defmodule Archethic.Contracts.Interpreter.ScopeTest do
...
+                   "my_var" => 42,
+                   "second_scope" => %{
+                     "another_var" => 99
+                   }
+                 }
+               }
+             } = Process.get(:scope)
+    end
+
+    test "should correctly overwrite variable in closest parent scope among multiple parent scopes" do
```
Neylix, line 376: Normally this behavior should never happen since the second "my_var" => 24 should have overwrite the first "my_var" => 11

test/archethic/contracts/interpreter/scope_test.exs
```diff
@@ -5,55 +5,589 @@ defmodule Archethic.Contracts.Interpreter.ScopeTest do
...
+            "second_scope" => %{
+              "var_in_second" => "value_in_second"
+            }
+          }
+        }
+      }
+
+      Process.put(:scope, scope)
+
+      var_value = Scope.read("var_in_second")
```
Neylix, line 452: you could directly
```elixir
assert "value_in_second" == Scope.read("var_in_second")
```

Neylix (changes requested):

lib/archethic/contracts/interpreter/scope.ex
aime-risson: i can't or it won't be tested

Neylix: You can test it in the test of execute_ast function

test/archethic/contracts/interpreter/scope_test.exs
aime-risson: actually does happen in for loops

lib/archethic/contracts/interpreter/scope.ex
aime-risson: actually the execute function wasn't tested but i don't know what tests I should add

Neylix: You can create some AST to be executed as a function, create a scope containing this function and some global variable, call Scope.execute_function and ensure the function is well executed and return expected result

Neylix (approved):

bchamagne (approved):

PR #1210, opened 2023-08-02
Rpc function call

aime-risson:
# Description

Allows to call a smart contract public function through the Json RPC api

Fixes # (issue)

## Type of change

# How Has This Been Tested?

Tested new rpc methods and errors

# Checklist:

### DIFF
--- a/lib/archethic_web/api/ecto_schemas/function_call_payload.ex
+++ b/lib/archethic_web/api/ecto_schemas/function_call_payload.ex
@@ -8,11 +8,12 @@ defmodule ArchethicWeb.API.FunctionCallPayload do
   embedded_schema do
     field(:contract, Address)
     field(:function, :string)
+    field(:args, {:array, :any})
   end
 
   def changeset(params = %{}) do
     %__MODULE__{}
-    |> cast(params, [:contract, :function])
+    |> cast(params, [:contract, :function, :args])
     |> validate_required([:contract, :function])
   end
 end
--- a/lib/archethic_web/api/ecto_schemas/function_call_payload.ex
+++ b/lib/archethic_web/api/ecto_schemas/function_call_payload.ex
@@ -1,0 +1,18 @@
+defmodule ArchethicWeb.API.FunctionCallPayload do
+  @moduledoc false
+  alias ArchethicWeb.API.Types.Address
+
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  embedded_schema do
+    field(:contract, Address)
+    field(:function, :string)
+  end
+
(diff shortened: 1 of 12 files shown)


lib/archethic_web/api/ecto_schemas/function_call_payload.ex
```diff
@@ -0,0 +1,18 @@
+defmodule ArchethicWeb.API.FunctionCallPayload do
+  @moduledoc false
+  alias ArchethicWeb.API.Types.Address
+
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  embedded_schema do
+    field(:contract, Address)
+    field(:function, :string)
```
Neylix, line 10: You should endure args is an array otherwise it will crash later.
Tried `args: "something"` and it crashed

lib/archethic_web/api/jsonrpc/methods/call_contract_function.ex
```diff
@@ -0,0 +1,69 @@
...
+  alias ArchethicWeb.API.JsonRPC.Method
+  alias ArchethicWeb.WebUtils
+
+  @behaviour Method
+
+  @spec validate_params(params :: map()) :: {:ok, params :: map()} | {:error, reasons :: list()}
+  def validate_params(params) do
+    case FunctionCallPayload.changeset(params) do
+      %{valid?: true, changes: changed_params} ->
+        {:ok, Map.put(changed_params, :args, Map.get(params, "args", []))}
```
Neylix, line 16: You could use `Map.update` over `Map.put` and `Map.get`

Neylix (changes requested):

lib/archethic_web/api/jsonrpc/methods/call_contract_function.ex
```diff
@@ -0,0 +1,72 @@
...
+  @spec validate_params(params :: map()) :: {:ok, params :: map()} | {:error, reasons :: list()}
+  def validate_params(params) do
+    case FunctionCallPayload.changeset(params) do
+      %{valid?: true, changes: changed_params} ->
+        {:ok, Map.update(changed_params, :args, [], fn val -> val end)}
+
+      changeset ->
+        reasons =
+          Ecto.Changeset.traverse_errors(changeset, fn
+            {message, [type: {:array, :any}, validation: :cast]} ->
```
Neylix, line 21: This would have been better to catch it in `translate_error` function I think, as it any other time we'll use an array it will be already catched

Neylix (changes requested):

Neylix (approved):

PR #1214, opened 2023-08-04
AEIP-19 token recipients

### LINKED ISSUE #1175: Support recipients in token minting transaction

Sometimes when we mint a token, we want to send at the same time some tokens to a predefined recipient list.
Since the [AEIP19](https://github.com/archethic-foundation/aeip/blob/main/AEIP-19.md) we introduce a specification to support it.

To implement this feature, we have to:
- [x] Adapt the JSON schema to support the `recipients` fields in the token's transaction specification
- [x] During the transaction's validation, the miners will mint the tokens and reduce from the UTXO the list of tokens associated to the list of recipients, and include them in the transaction's movements of the validation stamp
- [x] During the transaction's fee estimation, we have to include the calculation about the number of recipients to deliver the tokens to (may or may not require #1041)

### Additional context

_No response_

### Epic

_No response_

bchamagne:
# Description

/!\ Base branch 1174-token-resupply (AEIP-18) 
This is AEIP-19 implementation. This adds a `recipients` field in the token transactions. Minted tokens can be sent to recipients in the same transaction.

Fixes #1175 

## Type of change

# How Has This Been Tested?

Scenario 1: Create a token with recipients
```yaml
transaction_type: token
content: |
  {
    "supply": 10000000000,
    "type": "fungible",
    "decimals": 8,
    "name": "COINCOIN",
    "allow_mint": true,
    "recipients": [
      {"to": "00002223bbd4ec3d64ae597696c7d7ade1cee65c639d885450ad2d7b75592ac76afa", "amount": 333}
    ],
    "aeip": [2, 18]
  }
```
```
go run main.go send-transaction --config newtoken.yaml --access-seed bastien --index 0
```
- Check the movements in resulting

Scenario 2: 
- Create a resupply with recipients
```yaml
transaction_type: token
content: |
  {
    "aeip": [2,18],
    "supply": 100000000,
    "recipients": [{"to": "00002223bbd4ec3d64ae597696c7d7ade1cee65c639d885450ad2d7b75592ac76afa", "amount": 100000000}],
    "token_reference": "000058C44C933FBD047BF694490371BB41DA037597FFD57970A1DE6832AF990B7752"
  }
```
```
go run main.go send-transaction --config supply.yaml --access-seed bastien --index 1
```
- Check the movements in resulting

Many unit tests

# Checklist:

### DIFF
--- a/lib/archethic/mining/pending_transaction_validation.ex
+++ b/lib/archethic/mining/pending_transaction_validation.ex
@@ -865,7 +865,7 @@ defmodule Archethic.Mining.PendingTransactionValidation do
          %{"type" => "fungible", "allow_mint" => true} <- reference_json_token do
       :ok
     else
-      i when is_integer(i) ->
+      nil ->
         {:error, "Timeout when fetching the reference token or the genesis address"}
 
       {:exit, _} ->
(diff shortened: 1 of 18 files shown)


lib/archethic/mining/pending_transaction_validation.ex
```diff
@@ -825,49 +832,129 @@ defmodule Archethic.Mining.PendingTransactionValidation do
...
+
+    # Shut down the tasks that did not reply nor exit
+    results =
+      [tx_genesis_result, ref_genesis_result, ref_tx_result] =
+      Task.yield_many(tasks)
+      |> Enum.map(fn {task, res} ->
+        res || Task.shutdown(task, :brutal_kill)
+      end)
+
+    with 0 <- Enum.count(results, &(&1 == nil)),
```
Neylix, line 862: I think you don't need this since any failed task will return nil, so the next intruction in the `with` will fail and you'll get a `nil` in the else. So you can just pattern match in the else on `nil`

lib/archethic/mining/pending_transaction_validation.ex
```diff
@@ -825,49 +832,129 @@ defmodule Archethic.Mining.PendingTransactionValidation do
...
       {:error, %Jason.DecodeError{}} ->
         {:error,
          "Invalid token transaction - token_reference exists but does not contain a valid JSON"}
+    end
+  end
 
-      :error ->
-        {:error, "Invalid token transaction - token_reference is not an hexadecimal"}
+  defp verify_token_recipients(json_token) do
+    case json_token["recipients"] || [] do
```
Neylix, line 902: You could directly pattern match in the funtion parameter:
```elixir
defp verify_token_recipients(json_token = %{"recipients" => recipients})
    when is_list(recipients) do
  ...
end

defp verify_token_recipients(_), do: :ok

File: lib/archethic/mining/pending_transaction_validation.ex [hunk 3]
```diff
@@ -825,49 +832,129 @@ defmodule Archethic.Mining.PendingTransactionValidation do
...
+             :ok <- validate_token_recipients_total(recipients, json_token) do
+          validate_token_recipients_addresses(recipients)
+        end
+    end
+  end
+
+  defp validate_token_recipients_addresses(recipients) do
+    valid? =
+      recipients
+      |> Enum.map(& &1["to"])
```
Neylix, line 917: You could optimize this by reading only once the recipients list:
```elixir
Enum.all?(recipients, fn %{"to" => address} -> 
  address |> Base.decode16!(case: :mixed) |> Crypto.valid_address?()
end)
```

lib/archethic/mining/pending_transaction_validation.ex
```diff
@@ -825,49 +832,129 @@ defmodule Archethic.Mining.PendingTransactionValidation do
...
+      :ok
+    else
+      {:error, "Invalid token transaction - invalid recipients addresses"}
+    end
+  end
+
+  defp validate_token_recipients_total(recipients, json_token) do
+    total =
+      recipients
+      |> Enum.map(& &1["amount"])
```
Neylix, line 931: Same here you could read only once the list items:
```elixir
total = Enum.reduce(recipients, 0, fn %{"amount" => amount}, sum -> sum + amount end)
```

lib/archethic/mining/pending_transaction_validation.ex
```diff
@@ -825,49 +832,129 @@ defmodule Archethic.Mining.PendingTransactionValidation do
...
+      |> Enum.sum()
+
+    if total <= json_token["supply"] do
```
Neylix, line 934: You can retrieve suply by pattern matching function paramter to stay consistent with entire project

lib/archethic/replication/transaction_validator.ex
```diff
@@ -349,10 +339,15 @@ defmodule Archethic.Replication.TransactionValidator do
...
+        tokens_to_mint: LedgerOperations.get_utxos_from_transaction(tx, timestamp)
       }
-      |> LedgerOperations.from_transaction(tx, tx.validation_stamp.timestamp)
-      |> LedgerOperations.consume_inputs(tx.address, inputs, tx.validation_stamp.timestamp)
+      |> LedgerOperations.consume_inputs(
+        address,
+        inputs,
+        timestamp
+      )
+      |> elem(1)
```
Neylix, line 350: Here you should keep the `enough_funds?` boolean and return an error of the user don't have enough funds for its transfers and return `insufficient_funds` as it was done before

lib/archethic/transaction_chain/transaction.ex
```diff
@@ -953,4 +949,61 @@ defmodule Archethic.TransactionChain.Transaction do
...
+        []
+    end
+  end
+
+  defp get_movements_from_token_creation(tx_address, json) do
+    recipients = json["recipients"] || []
+    fungible? = json["type"] == "fungible"
+
+    Enum.map(recipients, fn r = %{"to" => address_hex, "amount" => amount} ->
+      token_id = r["token_id"] || if fungible?, do: 0, else: 1
```
Neylix, line 977: Here the token id should be always present if the token is non_fungible otherwise the user could send multiple time the same NFT. (It will result in unsufficient funds error)

lib/archethic/transaction_chain/transaction.ex
```diff
@@ -953,4 +949,61 @@ defmodule Archethic.TransactionChain.Transaction do
...
+          amount: amount,
+          type: {:token, tx_address, token_id}
+        }
+      end
+    end)
+    |> Enum.reject(&is_nil/1)
+  end
+
+  defp get_movements_from_token_resupply(json) do
+    recipients = json["recipients"] || []
```
Neylix, line 994: Same here you could pattern match this in functions parameters

lib/archethic/transaction_chain/transaction.ex
```diff
@@ -953,4 +949,61 @@ defmodule Archethic.TransactionChain.Transaction do
...
+    token_reference = json["token_reference"]
+
+    Enum.map(recipients, fn r = %{"to" => address_hex, "amount" => amount} ->
+      token_id = r["token_id"] || 0
```
Neylix, line 998: Here the user should not be able to set a token_id since it will always be 0 as we cannot user a token resupply on a non-fungible token

Neylix (changes requested):

Neylix (approved):

PR #1219, opened 2023-08-16
AEIP-16 named actions

### LINKED ISSUE #1170: Implement Smart Contract Named Actions

From the [AEIP16](https://github.com/archethic-foundation/aeip/blob/main/AEIP-16.md), we introduce the concept of named actions, in which smart contract actions can be named helping developers and clients to target specific code blocks depending on the use case.

To support it, we have to extend the interpreter with new features.

#### Parsing

- [ ] Actions block should have a new supported argument: `on` as in 
```elixir
actions triggered_by: transaction, on: vote(candidate) do
end

# or
actions triggered_by: transaction, on: update do
end
```

This new argument should accept function name with 0 or N arity.

Conditions

- [ ] Condition transaction should support the named actions called with the `on` parameter as in
```elixir
condition triggered_by: transaction, on: vote, as: [
  content: "hello"
]
```

- [ ] Condition transaction should also adapt a new syntax to support extension with an entire code block to bring more expressiveness for developers as in 

```elixir
condition triggered_by: transaction, on: vote(x, y) do

end
```

#### Execution

The smart contract worker and execution should 
- [ ] forward the call to the right actions block
- [ ] return an error if the named actions is not present

### Epic

_No response_

bchamagne:
# Description

Implement named actions. It introduces a new `transaction.version: 2` with a new format for the recipients. 
Fixes #1170 

- [ ] Update documentation
- [x] Update playground
- [ ] Update LIBJS
- [ ] Update LIBGO
- [ ] Update CLI
- [ ] Update LIBDART

## Type of change

# How Has This Been Tested?

Unit tests + manual testing (= deployment via playground + test via playbook (since CLI is not up to date yet))

# Checklist:


lib/archethic/transaction_chain/transaction/data/recipient.ex
```diff
@@ -0,0 +1,110 @@
...
+  end
+
+  def serialize(%__MODULE__{address: address, action: nil, args: nil}, 2) do
+    <<@unnamed_action::8, address::binary>>
+  end
+
+  def serialize(%__MODULE__{address: address, action: action, args: args}, 2) do
+    # 255 chars should be enough
+    action_bytes = byte_size(action)
+    true = 255 >= action_bytes
```
Neylix, line 43: This should be checked before.
In `TransactionPayload` or even when parsing the contract.

lib/archethic/transaction_chain/transaction/data/recipient.ex
```diff
@@ -0,0 +1,110 @@
...
+  @doc """
+  Deserialize a recipient
+  """
+  @spec deserialize(binary(), pos_integer()) :: {t(), binary()}
+  def deserialize(rest, 1) do
+    {address, rest} = Utils.deserialize_address(rest)
+    {%__MODULE__{address: address}, rest}
+  end
+
+  def deserialize(<<@unnamed_action::8, rest::binary>>, 2) do
```
Neylix, line 61: You could use only 1 bit for the type of action

lib/archethic_web/api/ecto_schemas/schema/transaction_data.ex
```diff
@@ -6,35 +6,123 @@ defmodule ArchethicWeb.API.Schema.TransactionData do
...
+        )
+    end
+  end
+
+  # recipients is a list of either binary or map
+  # so we can't use ecto to cast this value
+  defp maybe_cast_recipients(params = %{"recipients" => recipients}) do
+    recipients =
+      Enum.map(recipients, fn
+        %{"address" => address_hex, "action" => action, "args" => args} ->
```
Neylix, line 57: Maybe you can use `Map.get(recipient, "'action", nil)`

lib/archethic_web/api/graphql/schema/transaction_type.ex
```diff
@@ -58,7 +58,7 @@ defmodule ArchethicWeb.API.GraphQL.Schema.TransactionType do
     field(:code, :string)
     field(:content, :content)
     field(:ownerships, list_of(:ownership))
-    field(:recipients, list_of(:address))
+    field(:recipients, list_of(:recipient))
```
Neylix, line 61: Doing this means that every App that are using this graphql API will be broken when the release will be up on mainnet.
We should find a way to handle this update by versionning the API maybe

Neylix (changes requested):

lib/archethic_web/api/graphql/schema/transaction_type.ex
samuelmanzanera: Actually versioning on GraphQL API is not the most opiniated approach and was a first goal of GraphQL vs Rest approach' 

https://graphql.org/learn/best-practices/#versioning

Neylix: So this means we should have a new field like `namedActionRecipient` ?
This field would return the new format of recipient while the old field `recipients` could continue to return the old format of recipient (list of addresses)

samuelmanzanera: Can be an idea

PR #1226, opened 2023-08-22
JSON RPC methods that used transaction as payload now only takes 1 param

### LINKED ISSUE #1225: JsonRPC params to be similar to Open RPC
This is a subset of #1220 .
In this issue, we want to update the params to have "by-name" convention. 

# Example with send_transaction
## BEFORE
```json
{ 
  "params": {
    "address": "...",
    "type": "...",
  }
}
```

OpenRPC would display a `send_transaction(address, type, data...)` method, which is not expected.

## AFTER

```json 
{
  "params": {
    "transaction": {
      "address": "...",
      "type": "...",
    }
  }
}
```

OpenRPC would display a `send_transaction(transaction)` method, which is expected.

**This require SDK change to those that already implemented JSON RPC**:
- libgo
- libjs
- libdart

bchamagne:
# Description

Wrap the transaction payload inside a "transaction" object for these 3 methods: `SimulateContractExecution`, `EstimateTransactionFee` and `SendTransaction`.

Fixes #1225

## Type of change

**This require the 3 SDKs to be updated.**
- [x] libgo
- [x] libdart
- [x] libjs

# How Has This Been Tested?

Unit tested

# Checklist:

### DIFF
--- a/lib/archethic_web/api/jsonrpc/methods/estimate_transaction_fee.ex
+++ b/lib/archethic_web/api/jsonrpc/methods/estimate_transaction_fee.ex
@@ -19,8 +19,8 @@ defmodule ArchethicWeb.API.JsonRPC.Method.EstimateTransactionFee do
   """
   @spec validate_params(param :: map()) ::
           {:ok, params :: Transaction.t()} | {:error, reasons :: list()}
-  def validate_params(params) do
-    case TransactionPayload.changeset(params) do
+  def validate_params(%{"transaction" => transaction_params}) do
+    case TransactionPayload.changeset(transaction_params) do
       changeset = %{valid?: true} ->
         tx = changeset |> TransactionPayload.to_map() |> Transaction.cast()
         {:ok, tx}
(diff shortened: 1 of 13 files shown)


lib/archethic_web/api/jsonrpc/methods/estimate_transaction_fee.ex
```diff
@@ -19,8 +19,8 @@ defmodule ArchethicWeb.API.JsonRPC.Method.EstimateTransactionFee do
   """
   @spec validate_params(param :: map()) ::
           {:ok, params :: Transaction.t()} | {:error, reasons :: list()}
-  def validate_params(params) do
-    case TransactionPayload.changeset(params) do
+  def validate_params(%{"transaction" => transaction_params}) do
+    case TransactionPayload.changeset(transaction_params) do
```
Neylix, line 23: It could be nice to return an error if `transaction_params` is not a map.
This should be done in `TransactionPayload` I think

Neylix (changes requested):

Neylix (approved):

PR #1234, opened 2023-08-29
Adapt the fee for multi recipients transactions

### LINKED ISSUE #1041: Review of the fee calculation rule for multi recipients

Currently, when a transaction has multiple recipients, the fees are much higher.

Number of recipient (for transfers or smart contractcalls)
1: no more additional fee
> 1: each additional recipient will have an additinal cost of $0.1 UCO(Because sending transaction to multiple leverages more resources in term of network and storage, as the transaction must be replicated in all the chain targets)

The cost is to be rethought

### Additional context

Epic: #257

samuelmanzanera:
# Description

The idea is to ensure a bulk transaction to be more cost efficient than multiple transactions. 

Indeed, the validation is done one time versus multiple times.

The cost is then determined as logarithmic according to the number of recipients

Fixes #1041 

# How Has This Been Tested?

- Unit tests
- Multiple recipients sending with CLI

# Checklist:

### DIFF
--- a/test/archethic/mining/distributed_workflow_test.exs
+++ b/test/archethic/mining/distributed_workflow_test.exs
@@ -1274,7 +1274,7 @@ defmodule Archethic.Mining.DistributedWorkflowTest do
       proof_of_election: Election.validation_nodes_election_seed_sorting(tx, DateTime.utc_now()),
       ledger_operations:
         %LedgerOperations{
-          fee: Fee.calculate(tx, 0.07, timestamp),
+          fee: Fee.calculate(tx, 0.07, timestamp, ArchethicCase.current_protocol_version()),
           transaction_movements: Transaction.get_movements(tx),
           tokens_to_mint: LedgerOperations.get_utxos_from_transaction(tx, timestamp)
         }
--- a/lib/archethic/mining.ex
+++ b/lib/archethic/mining.ex
@@ -27,7 +27,7 @@ defmodule Archethic.Mining do
 
   use Retry
 
-  @protocol_version 1
+  @protocol_version 2
 
   def protocol_version, do: @protocol_version
 
@@ -245,6 +245,13 @@ defmodule Archethic.Mining do
   @doc """
   Get the transaction fee
   """
-  @spec get_transaction_fee(Transaction.t(), float(), DateTime.t()) :: non_neg_integer()
-  defdelegate get_transaction_fee(tx, uco_price_in_usd, timestamp), to: Fee, as: :calculate
+  @spec get_transaction_fee(
+          transaction :: Transaction.t(),
+          uco_price_in_usd :: float(),
+          timestamp :: DateTime.t(),
+          protocol_version :: pos_integer()
+        ) :: non_neg_integer()
+  def get_transaction_fee(tx, uco_price_in_usd, timestamp, proto_version \\ protocol_version()) do
+    Fee.calculate(tx, uco_price_in_usd, timestamp, proto_version)
+  end
 end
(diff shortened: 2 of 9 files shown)


Neylix:
I don't really agree with the log function for additionnal recipient.
Currently the entire validation process cost 1 cts. It represent the work done by welcome nodes, all validation nodes and the storage nodes of the transaction and the recipients.
This new fee calculation means that adding a recipient, which in fact adds only few calculation and one communication at the end with new pool of IO nodes, cost more than the entire validation process.

For me it looks strange to have more cost for something that takes less resources.

Addionnaly, doing this means that sending 2 differents transactions cost less fees than a single transaction with multiple recipients. But for the network, 2 transactions takes way more ressources than a single transaction

samuelmanzanera:
> I don't really agree with the log function for additionnal recipient. Currently the entire validation process cost 1 cts. It represent the work done by welcome nodes, all validation nodes and the storage nodes of the transaction and the recipients. This new fee calculation means that adding a recipient, which in fact adds only few calculation and one communication at the end with new pool of IO nodes, cost more than the entire validation process.
> 
> For me it looks strange to have more cost for something that takes less resources.
> 
> Addionnaly, doing this means that sending 2 differents transactions cost less fees than a single transaction with multiple recipients. But for the network, 2 transactions takes way more ressources than a single transaction.
> 
> It's like if I go to the garage and it cost me 50$ to change a wheel, but 120$ to change 2 wheels while it should be 100$ or less since there is a common setup for 1 or 2 wheels

In fact, this addition with log, ensure than multi recipients would cost less than many transactions.
Hence, there is an incentive to a do bulk transfer, as it would resolve into less resource consumption for the end to end workflow.

There is a property testing trying to ensure this assertion.
This was the goal of this PR to have this kind of calculation at the end.

Neylix:
> > I don't really agree with the log function for additionnal recipient. Currently the entire validation process cost 1 cts. It represent the work done by welcome nodes, all validation nodes and the storage nodes of the transaction and the recipients. This new fee calculation means that adding a recipient, which in fact adds only few calculation and one communication at the end with new pool of IO nodes, cost more than the entire validation process.
> > For me it looks strange to have more cost for something that takes less resources.
> > Addionnaly, doing this means that sending 2 differents transactions cost less fees than a single transaction with multiple recipients. But for the network, 2 transactions takes way more ressources than a single transaction.
> > It's like if I go to the garage and it cost me 50$ to change a wheel, but 120$ to change 2 wheels while it should be 100$ or less since there is a common setup for 1 or 2 wheels
> 
> In fact, this addition with log, ensure than multi recipients would cost less than many transactions. Hence, there is an incentive to a do bulk transfer, as it would resolve into less resource consumption for the end to end workflow.
> 
> There is a property testing trying to ensure this assertion. This was the goal of this PR to have this kind of calculation at the end.

Indeed ... It was late, didn't do the right calculation

test/archethic/mining/distributed_workflow_test.exs
```diff
@@ -1274,7 +1274,7 @@ defmodule Archethic.Mining.DistributedWorkflowTest do
       proof_of_election: Election.validation_nodes_election_seed_sorting(tx, DateTime.utc_now()),
       ledger_operations:
         %LedgerOperations{
-          fee: Fee.calculate(tx, 0.07, timestamp),
+          fee: Fee.calculate(tx, 0.07, timestamp, rchethicCase.current_protocol_version()),
```
bchamagne, line 1277: typo

bchamagne (approved):
Question: why do we change the protocol_version?

samuelmanzanera:
> Question: why do we change the protocol_version?

Because we still need to valid previous transaction's fee with previous algorithm

Neylix (approved):

PR #1249, opened 2023-09-04
Add consume_inputs in transaction's structure

### LINKED ISSUE #1241: AEIP-21: Store utxo on genesis pool

AEIP-21

- [x] Add consumed utxo in validation stamp (need update protocol version)
- [x] Send replication attestations to genesis pool
- [x] Store unspent utxos on genesis pool

### Additional context

_No response_

### Epic

_No response_

samuelmanzanera:
# Description

This PR adds a new field `consumed_inputs` in the transaction's structure, particularly in the ledger operations.
A new protocol version is defined to support backward compatible serialization.

During transaction's validation, the list of unspent outputs consumed will be logged in the validation stamp

Partially resolves #1241

## Type of change

# Checklist:


lib/archethic/transaction_chain/transaction/validation_stamp/ledger_operations.ex
```diff
@@ -280,26 +294,61 @@ defmodule Archethic.TransactionChain.Transaction.ValidationStamp.LedgerOperation
...
-        | new_token_unspent_outputs(
-            tokens_received,
-            tokens_to_spend,
-            change_address,
-            inputs,
-            timestamp
-          )
-      ]
+      %{uco: consumed_uco_utxos, token: consumed_token_utxos} =
+        get_inputs_to_consume(inputs, uco_to_spend, tokens_to_spend)
```
Neylix, line 298: I think the logic of consumed inputs is that if we take the consumed inputs and apply the movement of the transaction, it will result the same as the unspent outputs.
For now as we still put all the utxos in the unspent outputs, I think we should have all the inputs as well in the consumed inputs.
And then on the part 2 of the AEIP21 we will use your calculation of which inputs are really used.

Also in part 2 this algorithm could be done by the genesis pool before sending the inputs to the validation nodes

samuelmanzanera: But the part 1 is also to make sure the list of consumed inputs is the right one, before to actually using it.

Neylix: Hum for me not really.
I think the goal of the first part is to ensure that the genesis pool is storing well what we expect it to store. It should delete the consumed inputs and create the new unspent outputs.
If you don't put all the inputs in the consumed inputs, it means that you could have unspent outputs in the stamp that are already known and stored by the genesis pool. So the genesis pool will have a check to create the utxo only if it does not already exists.
But when we'll go on part 2 this check will not be needed anymore so we will update how the genesis pool store the unspent outputs which will make the part 1 less usefull

samuelmanzanera: The genesis pool need also the list of consumed inputs, to return the only right list of real unspent outputs later on the phase 2.
the genesis would still need to use the unspent outputs in the stamp to reallocate the real new unspent outputs.

samuelmanzanera: I think we should prepare the algorithm and storage to experiment it before real usage.

Neylix:
It look like the consumed inputs does not take the consolidated UCO inputs.
I updated the doctest of the function `consume_inputs` by replacing the transfer amount to 50_000_000

bchamagne (approved):
LGTM, I tested these scenarios:

- transfer uco
- creation of token with recipient
- resupply of token with recipient

Neylix (approved):

PR #1258, opened 2023-09-06
Fix functions with map params

### LINKED ISSUE #1257: Smart Contracts: functions cannot accept maps as parameter

This is a v1.3.0 bug, which is only available in testnet yet.

```elixir
@version 1

condition triggered_by: transaction, as: []
  Contract.set_content get_address(contract)
end


fun get_address(contract) do
  contract.address
end
```

This contract should pass

### Epic

_No response_

bchamagne:
# Description

Functions were failing with map parameters.
Fixes #1257 

## Type of change

# How Has This Been Tested?

Unit tests & playground

# Checklist:

### DIFF
--- a/lib/archethic/contracts/interpreter/scope.ex
+++ b/lib/archethic/contracts/interpreter/scope.ex
@@ -3,7 +3,7 @@ defmodule Archethic.Contracts.Interpreter.Scope do
   Helper functions to deal with scopes
   """
 
-  defp init(global_variables) do
+  def init(global_variables) do
     global_variables = Map.put(global_variables, "context_list", [])
 
     Process.put(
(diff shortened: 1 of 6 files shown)


lib/archethic/contracts/interpreter/scope.ex
```diff
@@ -3,7 +3,7 @@ defmodule Archethic.Contracts.Interpreter.Scope do
   Helper functions to deal with scopes
   """
 
-  defp init(global_variables) do
+  def init(global_variables) do
```
Neylix, line 6: **IMPORTANT:**
You should not make this function public to use it in tests.
Instead you can do, as other tests, create you map and use `Process.put` to create the scope (as the init function do)

Neylix (changes requested):

Neylix (approved):

PR #1284, opened 2023-09-26
Improve SC HTTP to use POST request

Neylix:
# Description

This PR allow users to make request with POST / PUT / DELETE / PATCH request method.
They can also make their own headers

## Type of change

# How Has This Been Tested?

Unit tests
Manual test with simple contract

# Checklist:

### DIFF
--- a/lib/archethic/contracts/interpreter/library/common/http_impl.ex
+++ b/lib/archethic/contracts/interpreter/library/common/http_impl.ex
@@ -17,6 +17,7 @@ defmodule Archethic.Contracts.Interpreter.Library.Common.HttpImpl do
   @behaviour Http
   @threshold 256 * 1024
   @timeout Application.compile_env(:archethic, [__MODULE__, :timeout], 2_000)
+  @allow_http? Application.compile_env(:archethic, [__MODULE__, :allow_http?], false)
 
   @tag [:io]
   @impl Http
@@ -149,15 +150,24 @@ defmodule Archethic.Contracts.Interpreter.Library.Common.HttpImpl do
         |> Keyword.get(:transport_opts, [])
     ]
 
-    with :ok <- validate_scheme(uri.scheme),
-         {:ok, conn} <- Mint.HTTP.connect(:https, uri.host, uri.port, conn_opts),
+    with {:ok, scheme} <- validate_scheme(uri.scheme),
+         {:ok, conn} <- Mint.HTTP.connect(scheme, uri.host, uri.port, conn_opts),
          {:ok, conn, _} <- Mint.HTTP.request(conn, method, path(uri), headers, request_body),
          {:ok, %{body: response_body, status: status}} <- stream_response(conn) do
       {:ok, %{"status" => status, "body" => response_body}}
     end
   end
 
-  defp validate_scheme("https"), do: :ok
+  defp validate_scheme("https"), do: {:ok, :https}
+
+  defp validate_scheme("http") do
+    if @allow_http? do
+      {:ok, :http}
+    else
+      {:error, :not_https}
+    end
+  end
+
   defp validate_scheme(_), do: {:error, :not_https}
 
   # copied over from Mint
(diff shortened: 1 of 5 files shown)


lib/archethic/contracts/interpreter/library/common/http_impl.ex
```diff
@@ -58,74 +61,85 @@ defmodule Archethic.Contracts.Interpreter.Library.Common.HttpImpl do
...
-      raise Library.Error, message: "Http module got called more than once"
-    end
+  def request_many(requests) do
+    check_too_many_calls()
 
-    uris_count = length(uris)
+    requests_count = length(requests)
 
-    if uris_count > 5 do
+    if requests_count > 5 do
```
samuelmanzanera, line 69: Suggestion: to simplify the codebase, you could use a guard upfront . And let pattern matching do the rest for the happy path.

test/support/mock_server.ex
samuelmanzanera, line 1: Suggestion: instead to have to create an entire mock server, you can use this lib for mocking http calls : https://hexdocs.pm/bypass/Bypass.html
It's a bit like 'nock' in JS world.

You wouldn't need a dedicated mocking server.

So within the tests, you can define the kind of endpoint and the expected results

lib/archethic/contracts/interpreter/library/common/http_impl.ex
```diff
@@ -58,74 +61,85 @@ defmodule Archethic.Contracts.Interpreter.Library.Common.HttpImpl do
...
-      end)
-      |> then(fn {bytes_total, results} ->
-        if bytes_total > @threshold do
-          raise Library.Error,
-            message: "Http.request_many/1 sum of responses is bigger than threshold"
-        else
-          results
-        end
+    requests =
+      Enum.map(requests, fn request = %{"url" => url, "method" => method} ->
```
samuelmanzanera, line 74: Suggestion: you could reduce iteration by using this logic within the async_stream_nolink, like:
```elixir
Task.Supervisor.async_stream_nolink( TaskSupervisor, requests, fn request ->
   # Handle assertion and formatting
   %{ "url" => url, "method" => method, "headers" => headers, "body" => body } = prepare_request(request)
   do_request(url, method, headers, body)
end)
...
```

lib/archethic/contracts/interpreter/library/common/http_impl.ex
```diff
@@ -192,14 +206,28 @@ defmodule Archethic.Contracts.Interpreter.Library.Common.HttpImpl do
     end
   end
 
+  defp validate_request(url, method, headers, body)
+       when is_binary(url) and is_binary(method) and is_map(headers) and
+              (is_binary(body) or is_nil(body)) do
```
samuelmanzanera, line 211: Suggestion: ideally I would prefer a `with` statement with appropriate private function to return ok/error tuple, as it would result in more simple and clean function body.

lib/archethic/contracts/interpreter/library/common/http_impl.ex
```diff
@@ -17,133 +17,134 @@ defmodule Archethic.Contracts.Interpreter.Library.Common.HttpImpl do
   @behaviour Http
   @threshold 256 * 1024
   @timeout Application.compile_env(:archethic, [__MODULE__, :timeout], 2_000)
+  @allow_http? Application.compile_env(:archethic, [__MODULE__, :allow_http?], false)
```
samuelmanzanera, line 20: Suggestion: Maybe you can use a list of supported schemes, this way it will be less specific to oriented https vs http but more generic, and I think we can support also later fetching from AEWeb/AEFS itself or other potentiel providers (ie. IPFS)

```elixir
@supported_schemes Application.compile_env(:archethic, [__MODULE__, :supported_schemes], [:https])
```

This will also simplify and optimize the check in `validate_scheme`

```elixir
defp validate_scheme(scheme) when scheme in @supported_schemes, do {:ok, scheme}
defp validate_scheme(_), do: {:error, :not_supported_scheme}
```

Neylix: Done in last commit

samuelmanzanera (approved):

bchamagne (approved):

PR #1296, opened 2023-10-10
Update recipient fee calculation

Neylix:
# Description

Since the user pay the fee for a contract trigger in recipient, the control in the Worker to ensure the contract has the minimum fees is not relevant anymore so it is delete.

Also if a contract call another contract it's transaction was free but we added after the recipient fee which was not logical since the user should pay the fee for this as well

## Type of change

# How Has This Been Tested?

Tested with a contract that call another contract (bridge workflow) the contract has no fund but the transaction is still passing

# Checklist:

### DIFF
--- a/lib/archethic/contracts/worker.ex
+++ b/lib/archethic/contracts/worker.ex
@@ -1,8 +1,6 @@
 defmodule Archethic.Contracts.Worker do
   @moduledoc false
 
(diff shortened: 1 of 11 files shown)


bchamagne (approved):
LGTM but there isn't a single test that uses the `contract_recipient_fee`.

PR #1297, opened 2023-10-11
Handle inputs amount overflow over 2^64

### LINKED ISSUE #1276: Consolidated inputs should handle token with amount > 2^64

Currently all ledger operation, inputs or utxo have an amount encoded on 64 bits.
When creating a token, the maximum amount is limited to 2^64 bits so it fit the maximum amount.

But since we introduced the token resupply in AEIP-18 we can now have multiple inputs of the same token address and the sum of those inputs can be encoded in more than 64 bits.
This can be a risk of loosing data.

We have 2 options: 
- Upgrading the encodage of unspent inputs with varint
- Modify the consolidation algorithm to not create inputs with more than 2^64 amout

### Epic

_No response_

Neylix:
# Description

This PR handle the unspent outputs with an amount over 2^64 - 1. 
Now the amount of an utxo is encoded using VarInt

Fixes #1276

## Type of change

# How Has This Been Tested?

Using blockchain and explorer normally.
Also started a node without the change, stopped the node, restarted it with the change.
Everything still work for old and new utxo serialization

# Checklist:

### DIFF
--- a/lib/archethic/contracts/interpreter/library/common/evm.ex
+++ b/lib/archethic/contracts/interpreter/library/common/evm.ex
@@ -47,6 +47,10 @@ defmodule Archethic.Contracts.Interpreter.Library.Common.Evm do
     check_types(:abi_encode, [first]) && list_or_variable_or_function?(second)
   end
 
+  def check_types(:abi_decode, [first, second]) do
+    binary_or_variable_or_function?(first) && binary_or_variable_or_function?(second)
+  end
+
   def check_types(_, _), do: false
 
   defp binary_or_variable_or_function?(arg) do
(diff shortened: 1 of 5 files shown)


samuelmanzanera (approved):
Seems good
Just if you can add comment to explain the reason or the logic of this function, so in further months when will come back we will know what is the reason of this function ^^

Neylix:
> Seems good Just if you can add comment to explain the reason or the logic of this function, so in further months when will come back we will know what is the reason of this function ^^

Added an explaination on top of the function

Neylix:
As discussed, creating multiple UTXO that could be exaclty the same does not fit the UTXO model.
So now to fox the issue, the unspent output amount is simply encoded using VarInt

bchamagne (approved):

samuelmanzanera (approved):

PR #1304, opened 2023-10-17
Add recipients arguments encoding

samuelmanzanera:
# Description

This PR aims to resolve an issue of inconsistency with JSON encoding, particularly with decimals.

Hence, it introduces a typed encoding approach of arguments and also use a different algorithm for the previous signature verification and the storage.

Previous signature tx's serialization happens with bit encoded on 1 byte (to simplify SDK and client integration), while for storage we can encode into 1 bit (Erlang allow us to optimize bitstring encoding easily)

## Type of change

# Checklist:


lib/archethic/transaction_chain/transaction/data/recipient/arguments_encoding.ex
```diff
@@ -0,0 +1,54 @@
...
+        :extended ->
+          8
+      end
+
+    bin =
+      args
+      |> Enum.map(&TypedEncoding.serialize(&1, bit_size))
+      |> :erlang.list_to_bitstring()
+
+    <<length(args)::8, bin::bitstring>>
```
Neylix, line 26: **Suggestion**
Encoding the length of the args list on 8 bits means the list could have a length of 255 items maximum.
This limitation should also be added in the json schema for the JSON RPC APIs

lib/archethic/transaction_chain/transaction.ex
```diff
@@ -683,71 +685,85 @@ defmodule Archethic.TransactionChain.Transaction do
...
+          type: type,
+          data: data,
+          previous_public_key: nil,
+          previous_signature: nil,
+          origin_signature: nil,
+          validation_stamp: nil
+        },
+        serialization_mode
+      ) do
     <<version::32, address::binary, serialize_type(type)::8,
```
Neylix, line 704: **Suggestion**
Maybe the serialization mode could be added in the serialization, so whenever we need to deserialize a transaction the deserialization will know which mode to use while the program may not know the serialization mode to use before deserialization

lib/archethic/utils/typed_encoding.ex
```diff
@@ -0,0 +1,147 @@
...
+  @type_bool 5
+  @type_nil 6
+
+  @type arg() :: number() | boolean() | binary() | list() | map() | nil
+
+  alias Archethic.Utils
+  alias Archethic.Utils.VarInt
+
+  @spec serialize(arg(), bit_size :: pos_integer()) :: bitstring()
+  def serialize(int, bit_size) when is_integer(int) do
```
Neylix, line 20: To my own it would make more sense to have the serialization mode in entry of this function and regarding it the serialization could adapt the bit size.
```elixir
TypedEncoding.serialize(value, :compact)
```

```elixir
serialize(value, :compact) do
    do_serialize(value, 1)
end

serialize(value, :extended) do
    do_serialize(value, 8)
end
```

This would avoid having case on serialization mode before calling the TypedEncoding.serialize function

And I think the responsability to determine what is the bit size of a compact encoding is to the TypedEncoding module, not the responsability of the module which calls it

test/archethic/transaction_chain/transaction_data/recipients/arguments_encoding_test.exs
```diff
@@ -0,0 +1,30 @@
...
+  use ExUnit.Case
+  use ExUnitProperties
+
+  alias Archethic.TransactionChain.TransactionData.Recipient.ArgumentsEncoding
+
+  property "symmetric encoding/decoding of recipients arguments" do
+    check all(
+            args <-
+              StreamData.list_of(
+                StreamData.one_of([
```
Neylix, line 11: **Suggestion**
The recipient args can also have map, list anf float as value. It could be nice to add them too in the property testing.
But this already tested in the TypedEncodingTest, so it's not mandatory

samuelmanzanera: Yes this was the reason to not include complex types

lib/archethic/transaction_chain/transaction.ex
samuelmanzanera: It can be an idea, but at the end the serialization/deserialization should be the most efficient and will result everytime as the compact mode (except for signature, but this serialization is not persisted)

Neylix: For now I agree it is not usefull. I was thinking for later, but we could change this later if needed

bchamagne (approved):
I tested while developing the new version of the JS SDK.
I was able to trigger contracts with named action.

samuelmanzanera:
:+1:

PR #1308, opened 2023-10-20
Scripts now takes arguments (testnet vs mainnet)

bchamagne:
Scripts now takes arguments so we can deploy both testnet & mainnet on the same machine.
Also change the version of TPM2-TSS because the version was removed from Github 🤦

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -3,8 +3,7 @@
 set -e
 trap 'echo "******* FAILED *******" 1>&2' ERR
 
-SUFFIX="${1:-mainnet}"
-INSTALL_DIR=~/aebot-$SUFFIX
+INSTALL_DIR=~/aebot/
 echo "INSTALL_DIR=$INSTALL_DIR"
 
 SCRIPT_DIR=$(dirname $(readlink -f $0))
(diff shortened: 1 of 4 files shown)


scripts/install.sh
```diff
@@ -138,7 +140,7 @@ make
 IFACE=$(route | grep '^default' | grep -o '[^ ]*$')
 FILENAME=`cat /sys/class/net/$IFACE/address`
 echo Generating Keys - It may take a while...
-./priv/c_dist/tpm_keygen > $INSTALL_DIR/$FILENAME 
+./priv/c_dist/tpm_keygen > $INSTALL_DIR/$FILENAME
```
Neylix, line 143: Ass the INSTALL_DIR finishes with a "/" this will create a path with double "/" no ?

scripts/release.sh
```diff
@@ -3,31 +3,32 @@
 set -e
 trap 'echo "******* FAILED *******" 1>&2' ERR
 
-USER=$(whoami)
-INSTALL_DIR=/home/$USER/aebot/build
```
Neylix, line 7: We should care if you delete this, as some nodes on the mainnet may broke since "/home/$USER "is not the same as "~"

Neylix (changes requested):

PR #1327, opened 2023-11-16
GetNetworkStats improvment: timeout & caching

### LINKED ISSUE #1325: Network state retrieval falls in timeout

During the beacon summary creation, the summary nodes calculate the network patches of all nodes.
To do so, each subset request the networks stats from other nodes. This request is done through the module `StateCollector`. This module aims to request those stats only once for all the subset.
To request network stats, the `StatCollector` sends a message `GetNetworkStats` to the other beacon nodes. The beacon nodes that receive this message will call the function `BeaconChain.get_network_stats` but here is a problem.
Since all beacons nodes request those stats, all beacon nodes received multiple times this message and so call multiple time the function `BeaconChain.get_network_stats` but this function looks like to to a lot of computation on the PC.

On this screenshot we can see that on summary time there is a pic of CPU usage and a down usage of network traffic. The down usage of network traffic could means that the CPU is overused and cannot handle even network requests.

image

On this other screenshot we can see the task responsible to call the function falls in timeout : 

image

We should find a way to not overuse the CPU to calculate the network stats.
We could have the same behavior as `StatCollector` but for the function `BeaconChain.get_network_stats`. Create a GenServer that execute this function only once, but respond to all request with the result.

We could also use profiler such as flamegraph to determine which part of the calculation consume a lot of ressources and try to improve the algorithm.

### Epic

_No response_

bchamagne:
# Description

The GetNetworkStats message is called by many nodes at the same time, the computation it does is heavy. I added a JobCache so the computation is only done once (cache reset after 30s). 
I also added a 5s timer which is bigger than the 3s default.
Fixes #1325

## Type of change

# How Has This Been Tested?

This can only be tested in stress condition. I'll setup some test environment tomorrow.

# Checklist:

### DIFF
--- a/config/test.exs
+++ b/config/test.exs
@@ -22,6 +22,7 @@ config :archethic, Archethic.BeaconChain.SummaryTimer,
   interval: "0 * * * * *"
 
 config :archethic, Archethic.BeaconChain.Subset.SummaryCache, enabled: false
+config :archethic, Archethic.BeaconChain.Subset.StatsCollector, enabled: false
 
 config :archethic, Archethic.Bootstrap, enabled: false
 
--- a/lib/archethic/application.ex
+++ b/lib/archethic/application.ex
@@ -104,7 +104,8 @@ defmodule Archethic.Application do
       MetricSupervisor,
 
       # a registry used in Utils to ensure a function is executed at most once concurrently
-      {Registry, keys: :unique, name: Archethic.RunExclusiveRegistry}
+      {Registry, keys: :unique, name: Archethic.RunExclusiveRegistry},
+      {Registry, keys: :unique, name: Archethic.Utils.JobCacheRegistry}
     ]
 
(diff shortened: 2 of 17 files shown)


Neylix (commented):
I think we can have a better way to handle this heavy calculation process.
The `StatsCollector` module is already doing one part of the network stats calculation by requesting only once the stats for all subsets.
We could reuse this module to also calculate only once the network patch for all the `GetNetworkStats` message a node receive.

Also those calculation/request are done only by the beacon summary node at summary time (00:00 currenlty) and are usefull before the self repair time (00:05 currently)

So we could use the `StatsCollector` module as an orchestrator that will start the 2 task (request/calculation) at summary time (using PubSub event) and delegate the calculation to `JobCache` module. Then store all client that request an information and return the result from the JobCache. Then when the self repair time is triggered (using Pubsub event) the module can clean the JobCache as the result will not be relevant anymore.

Neylix (approved):

PR #1333, opened 2023-11-28
Dashboard does not use metrics anymore

### LINKED ISSUE #1182: Metrics fetching consume too much network data

Currently we have a dashboard display on the explorer. It display some information about number of transaction processed and transaction validation time.
To do this, a node needs to fetch the metrics of all other nodes. These metrics are refreshed every 5 sec.
Here is the calculation of how much data is consume per day:
the number of request is 12 per minutes so 12 * 60 * 24 = 17280 request per node per day
There is currently 14 nodes to fetch on the mainnet so 14 * 17280 = 241 920 request per day
One metrics request has a size of 199 KB so 199 * 241 920 = 45.911 GB

So there is a big problem with this behavior as it consume a huge amount of data.
We can see it throught observer_cli on a node:

image

~730 GB consumed in input and output on a period of 21 days

Find a solution to reduce the impact of the metrics.
Could be to fetch only relevant metrics or find another way to display metrics

### Epic

_No response_

bchamagne:
# Description

With 21 nodes, we used to transmit more than 100GB per node per day to display the dashboard. 

Before: `21 nodes * 300KB of data * 720 requests per hour * 24 hours = 100GB`
After: `21 nodes * 4KB of data * 60 requests per hour * 24 hours = 121MB`

I removed the fetching of metrics from nodes to nodes and replaced it with 2 new processes: 
- DashboardMetrics listens to local events and keep them in buckets
- DashboardMetricsAggregator request regularly other nodes for their DashboardAggregator data 

The aggregation is currently set to 1 minute and we keep 60 minutes of history. This way the dashboard can display the last hour of data as soon as we hit the page.

We could do smaller buckets / longer history if the need arises.

Fixes #1182

Capture d’écran 2023-12-19 à 18 33 13

## Type of change

# How Has This Been Tested?

Unit tests + tested locally with 4 nodes

# Checklist:


test/archethic_web/dashboard_aggregator_test.exs
```diff
@@ -0,0 +1,76 @@
+defmodule ArchethicWeb.DashboardAggregatorTest do
+  use ExUnit.Case
+
+  alias ArchethicWeb.DashboardAggregator
+  alias Archethic.PubSub
+
+  setup do
+    # the DashboardAggregator is already started and supervised
```
Neylix, line 8: This could be avoided by using the `enabled` attribute in test config. So the setup could be just to start the GenServer using `start_supervised`

test/archethic_web/dashboard_aggregator_aggregator_test.exs
```diff
@@ -0,0 +1,139 @@
...
+      first_public_key: other_node_pkey,
+      last_public_key: other_node_pkey,
+      network_patch: "BBB",
+      geo_patch: "BBB",
+      available?: true,
+      authorized?: true,
+      authorization_date: DateTime.utc_now() |> DateTime.add(-1)
+    })
+
+    # the DashboardAggregatorAggregator is already started and supervised
```
Neylix, line 45: Same as DashboardAggregator for test config

Neylix (changes requested):
For the naming you could rename `DashboardAggregator` to `DashboardMetrics` and `DashboardAggregatorAggregator` to `DashboardMetricsAggregator` it looks nicer I think

Neylix:
I also have weird behavior running 2 nodes:

image

Also there is something wrong in the calculation of number of transaction. Currently for a bucket of a timestamp you create a list of all the durations of all nodes, but some durations of different node refer the same transaction, so if you have only one transaction processed by 3 nodes you'll say there where 3 transactions processed in the graph while in fact there were only 1 transaction.

A solution could be to have a map with transaction address => duration, so you could easily calculate the real number of transaction, and it would also be a more accurate calculation of the average transaction validation time

lib/archethic_web/explorer/live/dashboard_live.ex
```diff
@@ -59,30 +57,84 @@ defmodule ArchethicWeb.Explorer.DashboardLive do
...
   defp fetch_stats() do
+    # DashboardMetricsAggregator returns an _UNORDERED_ map
+    # we convert it to a ORDERED list of pairs
     DashboardMetricsAggregator.get_all()
-    |> Enum.map(fn {{node_public_key, datetime}, durations} ->
-      {{Base.encode16(node_public_key), datetime}, durations}
+    |> Enum.map(fn {{node_public_key, datetime}, durations_by_address} ->
+      {{Base.encode16(node_public_key), datetime}, durations_by_address}
+    end)
+    |> Enum.sort(fn {{_, datetime_a}, _}, {{_, datetime_b}, _} ->
```
Neylix, line 66: **Suggestion**
You can use `Enum.sort_by` here
```elixir
Enum.sort_by(fn {{_, datetime}, _} -> datetime end, {:asc, DateTime})
```

Neylix (approved):

bchamagne:
Capture d’écran 2023-12-20 à 19 36 38

lib/archethic_web/explorer/live/dashboard_live.ex
```diff
@@ -4,147 +4,166 @@ defmodule ArchethicWeb.Explorer.DashboardLive do
...
+      {timestamp, %{count: 0, sum: 0}} ->
+        {timestamp, 0}
 
-  defp cumul(
-         acc,
-         {"archethic_mining_full_transaction_validation_duration", %{count: count, sum: sum}}
-       ) do
-    put_in(acc, [Access.key(:cumul, %{}), :transaction_validation], {count, sum})
+      {timestamp, %{count: count, sum: sum}} ->
+        {timestamp, div(sum, count) / 1_000_000}
```
Neylix, line 116: Why do you divide by 1_000_000, this result in 0.00035 ms on the graph which is invalid

bchamagne:
Capture d’écran 2023-12-21 à 20 21 37

Neylix (approved):

PR #1343, opened 2023-12-07
Smart Contracts: InvalidRecipientsExecution now contains the reason

bchamagne:
# Description

This way, if a transaction is refused because a smart contract would have failed, the reason is propagated to the user.

## Type of change

# How Has This Been Tested?

Tested with 3 nodes

# Checklist:

### DIFF
--- a/lib/archethic/p2p/message/validate_smart_contract_call.ex
+++ b/lib/archethic/p2p/message/validate_smart_contract_call.ex
@@ -11,6 +11,8 @@ defmodule Archethic.P2P.Message.ValidateSmartContractCall do
   alias Archethic.Contracts.Contract.Failure
   alias Archethic.Contracts.Contract.ConditionRejected
   alias Archethic.Contracts.Contract.ActionWithTransaction
+  alias Archethic.Contracts.Contract.ConditionRejected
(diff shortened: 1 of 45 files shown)


lib/archethic/p2p/message/validate_smart_contract_call.ex
```diff
@@ -78,7 +80,8 @@ defmodule Archethic.P2P.Message.ValidateSmartContractCall do
 
     with {:ok, contract_tx} <- TransactionChain.get_transaction(recipient_address),
          {:ok, contract} <- Contracts.from_transaction(contract_tx),
-         trigger when not is_nil(trigger) <- Contract.get_trigger_for_recipient(recipient),
+         {:trigger, trigger} when not is_nil(trigger) <-
```
samuelmanzanera, line 83: Maybe you should add a private function to continue to leverage `{:ok, _}` syntax. 
```elixir
defp validate_trigger(recipient) do
  case Contract.get_trigger_for_recipient(recipient) do
    nil ->
      {:error, :invalid_trigger}
    _ ->
      {:ok, trigger}
  end
end
```
This would make the code more readable and consistent

bchamagne:
LGTM

PR #1346, opened 2023-12-11
Reconnection backoff & heartbeat mechanism

### LINKED ISSUE #1323: Nodes should not try to connect to stale nodes

Currently, for each nodes in the network we create a Connection process which holds the tcp connection with the node. If the tcp connection is closed, the process retry to connect to the node every 500ms.
For some reason a node can stop working or can be offline forever. If its happen there is no needs to retry to connect to it every 500ms this it may not be connectable anymore.

If a node becomes completely unavailable the Connection process should be stopped.
If the node is restarted, it will send a message to us and so we should restart the Connection process since the node is back again.

### Additional context

A node is considered a completely unavailable if it is globally unavailable and the Connection availability timer of the Connection process is nil for more than 1 Slot time

### Epic

_No response_

bchamagne:
# Description

- Exponential backoff on reconnect
- Stop trying to reconnect after X times
- Wakeup when the other node connect to us
- Heartbeat to automatically close "stale" connections

- [ ] There is one scenario that is still to be resolved. In case of a "split", both nodes consider the other as "stale". In that scenario, there is no attempt of reconnection ever.

Fixes #1323

## Type of change

# How Has This Been Tested?

Tested locally with multiple nodes and turning them on and off.

# Checklist:

### DIFF
--- a/lib/archethic/p2p/client/connection.ex
+++ b/lib/archethic/p2p/client/connection.ex
@@ -23,6 +23,7 @@ defmodule Archethic.P2P.Client.Connection do
   use GenStateMachine, callback_mode: [:handle_event_function, :state_enter], restart: :temporary
   @vsn Mix.Project.config()[:version]
   @table_name :connection_status
+  @max_reconnect_attempts 5
 
   @doc """
   Starts a new connection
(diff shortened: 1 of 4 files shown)


samuelmanzanera:
Do you think we should add a `code_change` to adapt the state, or prefer restarting the process ?

Neylix (changes requested):
I think you should keep the "wakup" functionnality be cause if a node goes offline for a while, all other nodes will have a reconnection each 24 hours.
But when the node goes online, others nodes may reconnect to it after 24 hours which is too long

lib/archethic/p2p/client/connection.ex
```diff
@@ -381,6 +433,35 @@ defmodule Archethic.P2P.Client.Connection do
...
+        {:connected, socket},
+        data = %{
+          transport: transport,
+          heartbeats_sent: heartbeats_sent,
+          heartbeats_received: heartbeats_received
+        }
+      ) do
+    # disconnect if missed more than 2 heartbeats
+    if heartbeats_sent - heartbeats_received >= 2 do
+      {:next_state, :disconnected, data}
```
Neylix, line 448: I guess here you should also close the socket

Neylix (changes requested):

Neylix (approved):

PR #1347, opened 2023-12-11
Utils.run_exclusive doesn't not send messages anymore

### LINKED ISSUE #1306: Utils.run_exclusive causes GenServer error

When a node is not synchronized it starts a repair process.
This process is started with `NetworkChain.asynchronous_resync` which calls `Utils.run_exclusive`. But this last one create a task using `Task.async_nolink` which means that the process that created this task will receive an event `{ref, res}` and `{:DOWN, ...}`.

Since this function may be called from GenServer if they don't handle the event they'll have a function clause error.
Example from Node responsivness and Node Renewal Scheduler:

image

image

The Utils.run_exclusive should not start an async task that return event to the caller but a "fire and forget" task using `Task.Supervisor.start_child`

### Epic

_No response_

bchamagne:
# Description

Fixes #1306

## Type of change

# How Has This Been Tested?

Unit tested 

# Checklist:

### DIFF
--- a/lib/archethic/utils.ex
+++ b/lib/archethic/utils.ex
@@ -1061,7 +1061,7 @@ defmodule Archethic.Utils do
 
     case Registry.lookup(registry, key) do
       [] ->
-        Task.Supervisor.async_nolink(
+        Task.Supervisor.start_child(
           Archethic.TaskSupervisor,
           fn ->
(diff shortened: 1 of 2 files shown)


samuelmanzanera (approved):
LGTM

Neylix (approved):

PR #1349, opened 2023-12-12
Smart Contracts: Logs for the playground

bchamagne:
# Description

Added a `Playground.print/1` function that will forward the term to the playground's console. 

## Type of change

# How Has This Been Tested?

Unit tests + to be tested with the playground:

- [x] Logs on the condition
- [x] Logs on the named action
- [x] Logs on the unamed action
- [x] Logs on a public function
- [x] Logs on a private function

# Checklist:


samuelmanzanera:
Maybe we should have something less coupled to the playground. Why not using a log mechanism as result of execution like you did without having to branch on playground but the playground can catch it to display on the console. 

My take would be to have something similar on chain in a near future to be able to display logs as EVM offers. 
However we should have a way to restraint it in fee mechanism. Probably also a typing can be interesting

samuelmanzanera:
> Maybe we should have something less coupled to the playground. Why not using a log mechanism as result of execution like you did without having to branch on playground but the playground can catch it to display on the console.
> 
> My take would be to have something similar on chain in a near future to be able to display logs as EVM offers. However we should have a way to restraint it in fee mechanism. Probably also a typing can be interesting

Do you think it's possible to add `Playground.print` from the playground itself by overriding the interpreter ?

Because logs on chain should be more events than just print

PR #1372, opened 2024-01-09
MUCO to UCO in transaction movements

### LINKED ISSUE #1368: MUCO to UCO transformation should be done during validation

Currently when we send a MUCO (mining reward token) the storage node of the recipient convert the MUCO to UCO. But this is only a view of the storage node and this is not written in the transaction validation stamp.

When a transaction has a transfer of MUCO, the corresponding movement should be converted to UCO directly, so it will be written in the validation stamp

### Additional context

We should keep the old behavior to be retro compatible mainly for the self repair

### Epic

_No response_

Neylix:
# Description

This PR update the behavior of converting MUCO to UCO. The conversion is now done during the validation in the movements.
Updated the explorer view to display the conversion

Fixes #1368

## Type of change

# How Has This Been Tested?

- Unit tests
- Update the reward address of a node (localhost:4000/settings)
Using the chain of the new reward address, send the MUCO to another address, in the validation stamp the movement is UCO.
- Also created a transfer of MUCO before applying this PR, then applied the PR with new explorer behavior, it works as it was previously (just display MUCO)

# Checklist:

### DIFF
--- a/lib/archethic/mining/validation_context.ex
+++ b/lib/archethic/mining/validation_context.ex
@@ -64,8 +64,6 @@ defmodule Archethic.Mining.ValidationContext do
   alias Archethic.TransactionChain.Transaction.ValidationStamp
   alias Archethic.TransactionChain.Transaction.ValidationStamp.LedgerOperations
 
-  alias Archethic.TransactionChain.Transaction.ValidationStamp.LedgerOperations.TransactionMovement
-
   alias Archethic.TransactionChain.Transaction.ValidationStamp.LedgerOperations.UnspentOutput
   alias Archethic.TransactionChain.TransactionData
   alias Archethic.TransactionChain.TransactionData.Recipient
@@ -800,7 +798,7 @@ defmodule Archethic.Mining.ValidationContext do
 
   defp get_ledger_operations(
          %__MODULE__{
-           transaction: tx,
+           transaction: tx = %Transaction{address: address, type: tx_type},
            unspent_outputs: unspent_outputs,
            resolved_addresses: resolved_addresses
          },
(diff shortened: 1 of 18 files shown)


bchamagne (approved):
I like what you've done with `ledgerOp` & `consume_input`, looks cleaner. 

Tested and numbers are good.

PREVIOUS INPUTS
Capture d’écran 2024-01-10 à 17 49 35

TX TRANSFER OF 0.01
Capture d’écran 2024-01-10 à 17 49 26

lib/archethic/mining/validation_context.ex
```diff
@@ -800,30 +798,31 @@ defmodule Archethic.Mining.ValidationContext do
...
-    )
+    movements = Transaction.get_movements(tx)
+
+    {sufficient_funds?, ops} =
+      LedgerOperations.consume_inputs(
+        %LedgerOperations{fee: fee},
+        address,
+        unspent_outputs,
+        movements,
+        LedgerOperations.get_utxos_from_transaction(tx, validation_time),
```
samuelmanzanera, line 817: I'm wondering if we should not just concat unspent_outputs with the new unspent outputs  (tokens_to_mint) at first.
This would simplify the code of the `consume inputs` fn , because at the end `tokens_to_mint` is only used to be concatenated into the list of inputs.

samuelmanzanera: After AEIP21, the new tokens will still be reaming in the UTXO and only be in the consumed inputs if the token define minted transfers. So I don't see yet the benefits to keep them separately

Neylix: A transaction could consume inputs not yet created ?

samuelmanzanera: Yeah..maybe not.
It might be strange evenif in a pure a UTXO model , an output should come from an input .

Neylix: In case of token minting, the output could not come from inputs since the transaction is creating new things so I guess it's valid

samuelmanzanera: Token minting is a particular use case for UTXO, not initial thought from it.
I think both could still be valid depending of people point of view but for simplification we can omit it, particularly than after AEIP21 
`new_token_unspent_outputs` might not be used and we would need identification of the `tokens_to_mint`

samuelmanzanera (approved):

PR #1383, opened 2024-01-16
DashboardData uses a different struct

### LINKED ISSUE #1382: DashboardData serialization issue

DashboardData has a property `buckets` which is a `%{DateTime.t() => list()}`. 
The issue is that map order cannot be garanteed and it happens that some nodes have a different map order than other resulting in message that does not decode properly (due to signatures no matching on sender and receiver)

We must use another structure

### Epic

_No response_

bchamagne:
# Description

Changed a map with a list of pairs to guarantee the order. Which is necessary because messages are signed and verified. Different nodes might sort a map differently.

Fixes #1382

## Type of change

- 
# How Has This Been Tested?

Unit tested & locally with few nodes

# Checklist:


lib/archethic/p2p/message/dashboard_data.ex
```diff
@@ -27,20 +27,20 @@ defmodule Archethic.P2P.Message.DashboardData do
...
   def deserialize(<<rest::bitstring>>) do
     {buckets_count, rest} = VarInt.get_value(rest)
-    {buckets, rest} = deserialize_buckets(buckets_count, %{}, rest)
+    {buckets, rest} = deserialize_buckets(buckets_count, [], rest)
 
     {%__MODULE__{buckets: buckets}, rest}
   end
 
-  defp deserialize_buckets(0, acc, rest), do: {acc, rest}
+  defp deserialize_buckets(0, acc, rest), do: {Enum.reverse(acc), rest}
```
samuelmanzanera, line 43: Suggestion: To optimize the computation, I think you can omit `Enum.reverse/1` as the order is not required outside the serialization

Neylix (approved):

samuelmanzanera (approved):

PR #1385, opened 2024-01-16
Integrate genesis address in self repair

### LINKED ISSUE #1242: AEIP-21: Update self repair

_No response_

- [x] Add genesis address in transaction summary
- [x] Determine genesis address for legacy transaction summary (version 1) in self repair to inject into genesis ledger
- [x] Run election to determine if the node should download the transaction based on genesis addresses
- [x] Adapt the notifier to include genesis synchronization

### Additional context

_No response_

### Epic

_No response_

samuelmanzanera:
# Description

This PR fixes #1242 

## Type of change

# How Has This Been Tested?

- Unit test
- Start 2 of 3 nodes from AEIP21 branch
- Upgrade the 2 running nodes with the PR's code
- Await some transfers are done with the new version
- Start the 3rd node with the PR's code release
- Stop one of the first 2 nodes, and restart it.

# Checklist:


lib/archethic/self_repair/notifier.ex
```diff
@@ -141,15 +143,15 @@ defmodule Archethic.SelfRepair.Notifier do
...
     prev_storage_nodes =
-      Election.chain_storage_nodes(address, prev_available_nodes)
+      address
+      |> Election.chain_storage_nodes(prev_available_nodes)
       |> Enum.map(& &1.first_public_key)
 
-    resolved_addresses =
-      transaction_movements
-      |> Enum.map(& &1.to)
+    resolved_addresses = [genesis_address | Enum.map(transaction_movements, & &1.to)]
```
Neylix, line 154: Here the resolved addresses for the genesis address should take in consideration the protocol version of the tx:
- During phase 1, tx will have last address as resolved movement, so the notifier should retrieve genesis address of each movement recipient and send the ShardRepair message to last address io node and genesis io node
- After phase 2, tx will have genesis address as resolved movement, so the notifier could direclty send the message to the genesis address. But legacy transaction will still have last address as resolved movement, so if the transaction is legacy, the notifier should retrieve genesis address for each movement.

And in both case we still need to put the chain genesis address of the transaction as you did.

lib/archethic/self_repair/sync/transaction_handler.ex
```diff
@@ -35,18 +35,44 @@ defmodule Archethic.SelfRepair.Sync.TransactionHandler do
...
       ) do
-    chain_storage_nodes = Election.chain_storage_nodes_with_type(address, type, node_list)
+    last_chain_storage_nodes = Election.chain_storage_nodes_with_type(address, type, node_list)
+    genesis_chain_storage_nodes = Election.chain_storage_nodes(genesis_address, node_list)
+    node_key = Crypto.first_node_public_key()
 
-    if Utils.key_in_node_list?(chain_storage_nodes, Crypto.first_node_public_key()) do
+    if Utils.key_in_node_list?(last_chain_storage_nodes, node_key) or
+         Utils.key_in_node_list?(genesis_chain_storage_nodes, node_key) do
       not TransactionChain.transaction_exists?(address)
```
Neylix, line 51: Here the is a mismatch for the genesis address pool. If you are genesis node of a chain you will store transaction as io transaction. So for the genesis pool, then function `TransactionChain.transaction_exists?` without the io flag will always return false.

Maybe the genesis pool check should be done in the else statement

lib/archethic/self_repair/sync/transaction_handler.ex
```diff
@@ -35,18 +35,44 @@ defmodule Archethic.SelfRepair.Sync.TransactionHandler do
...
     else
       io_node? =
```
Neylix, line 53: To optimize the request, you should previously ensure the transaction does not already exist it the DB to avoid doing IO and having a "transaction already exists" after

lib/archethic/self_repair/sync/transaction_handler.ex
```diff
@@ -35,18 +35,44 @@ defmodule Archethic.SelfRepair.Sync.TransactionHandler do
...
-        Enum.any?(mvt_addresses, fn address ->
+        mvt_addresses
+        |> Task.async_stream(fn address ->
+          case summary_version do
+            # Before AEIP-21, we fetch the genesis address to avoid multiple beacon chain's data changes
+            1 ->
```
Neylix, line 58: Here you assume that if the transaction summary is version 2, it means the resolved movement address are the genesis, but this is not true. During phase 1, transaction summary will be version 2, but resolved movement address will still be the last address
The only true condition to ensure the behavior of the resolved movement address is to check the protocol version which is not available in the transaction summary.

So we have 2 options, either have a condition for the resolved movement base on the transaction summary version, which mean we will need to have a transaction summary version 3 after phase 2 without any modification of the transaction summary. Or put the protocol version in the transaction summary

lib/archethic/self_repair/sync/transaction_handler.ex
```diff
@@ -92,7 +123,10 @@ defmodule Archethic.SelfRepair.Sync.TransactionHandler do
            acceptance_resolver: acceptance_resolver
          ) do
       {:ok, tx = %Transaction{}} ->
-        summary = TransactionSummary.from_transaction(tx)
+        summary =
+          tx
+          |> TransactionSummary.from_transaction(genesis_address)
```
Neylix, line 128: I would recommend to add a new flag or create a new function in `TransactionSummary` to select the version of the transaction summary to create

Neylix (changes requested):

lib/archethic/self_repair/sync.ex
```diff
@@ -343,6 +349,30 @@ defmodule Archethic.SelfRepair.Sync do
...
+             address: tx_address,
+             version: 1
+           }
+         },
+         download_nodes
+       ) do
+    storage_nodes = Election.chain_storage_nodes(tx_address, download_nodes)
+    {:ok, genesis_address} = TransactionChain.fetch_genesis_address(tx_address, storage_nodes)
+
+    put_in(
```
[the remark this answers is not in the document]
Neylix: As @bchamagne said, this is strange to have a transaction summary version 1 with a genesis address.
A solution could be to update to version 2 as well, but we will encounter an issue after. When we verify the signature of the attestation we will serialize the transaction summary with the genesis address, but during the validation the storage nodes signed a transaction summary version 1, and so the signature verification will fail.

I suggest to not update the transaction summary and fetch the genesis address in the `download_transaction?` function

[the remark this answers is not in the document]
samuelmanzanera: The point was to unify upfront the data structures, so less I/O logic in business logic.
For me the version is more impacting by the how the transaction summary was created and hence signed, but the data structure including a new field can de decorrelated to a version but  it's just data model transformation.

@bchamagne My intenation was to particularly have version 1 with genesis.

lib/archethic/self_repair/notifier.ex
samuelmanzanera: It was the intentation as this PR targets for now only phase1 and we don't know the protocol version of phase2.
Hence it will be simplifier once phase2 developed to adapt the notifier to include the right protocol version.

lib/archethic/self_repair/sync/transaction_handler.ex
samuelmanzanera: Yes true

lib/archethic/self_repair/sync/transaction_handler.ex
samuelmanzanera: Not false, then I would prefer to increase the summary's version in order to reduce storage size for the beacon, as it's not particularly necessary.

lib/archethic/self_repair/sync.ex
```diff
@@ -363,6 +412,31 @@ defmodule Archethic.SelfRepair.Sync do
...
+           transaction_summary: tx_summary = %TransactionSummary{
+             version: 1,
+             movements_addresses: movements_addresses
+           }
+         },
+         %Transaction{validation_stamp: %ValidationStamp{recipients: recipients}}
+       ) do
+    consolidated_movements_addresses =
+      recipients
+      |> Task.async_stream(fn recipient ->
```
Neylix, line 426: Here also you could add `on_timeout: :kill_task` and maybe some `max_concurrency`

lib/archethic/transaction_chain/transaction_summary.ex
```diff
@@ -202,57 +206,85 @@ defmodule Archethic.TransactionChain.TransactionSummary do
...
-    {address, <<timestamp::64, type::8, fee::64, rest::bitstring>>} =
-      Utils.deserialize_address(data)
+  @spec resolve_movements_addresses(t(), list(Node.t())) :: Enumerable.t() | list(binary())
+  def resolve_movements_addresses(
+        %__MODULE__{movements_addresses: addresses, version: version},
+        node_list
+      )
+      when version <= 2 do
+    addresses
+    |> Task.async_stream(fn address ->
```
Neylix, line 242: Same here with Task parameters

lib/archethic/self_repair/notifier.ex
```diff
@@ -138,18 +140,40 @@ defmodule Archethic.SelfRepair.Notifier do
...
       |> Enum.map(& &1.to)
+      |> Enum.concat(recipients)
+
+    # Before AEIP-21, resolve movements included only last addresses,
+    # then we have to resolve the genesis address for all the movements
+    resolved_addresses =
+      if protocol_version < 7 do
+        Task.async_stream(movements_addresses, fn address ->
+          {:ok, resolved_genesis_address} =
+            TransactionChain.fetch_genesis_address(address, P2P.authorized_and_available_nodes())
```
Neylix, line 167: You should do the election of the address before fetching the genesis

test/archethic/transaction_chain/transaction_summary_test.exs
```diff
@@ -3,5 +3,39 @@ defmodule Archethic.TransactionChain.TransactionSummaryTest do
 
   alias Archethic.TransactionChain.TransactionSummary
```
Neylix, line 5: You may add test for the 2 new functions of TransactionSummary (resolve_addresses and equals?)

Neylix (changes requested):

PR #1394, opened 2024-01-23
Lock chain before start mining

### LINKED ISSUE #1373: Split brain on a transaction
# Some facts
- The transaction address is generated via the `chain_size` and the `chain_seed`.
- The transaction hash is generated via the payload of the transaction.
- The election of validators is based on the transaction hash.

# Issue
It is possible today to create multiple transactions with the same address and a different hash. If these  transactions are mined at the same time, they will have different elections, and they will be mined concurrently on different nodes. We have a split-brain situation.

# Consequence
Differents nodes will have different transactions at the same address. 
Impossible to create a new transaction on the same chain. (Because the Proof of Integrity will fail in some nodes)

Neylix:
# Description

This PR adds a new security layer before to start mining a transaction.
When receiving the StartMining message for a transaction, validation nodes request the storage nodes of this transaction to lock the address. If the address is already locked for the same address but with a different transaction data, the lock is refused and the validation node does not start the mining.

This PR does not handle malicious node which could still start the transaction mining even with no lock. And so ask the replication to storage nodes.
To handle this, the storage nodes could sign the lock request, and so the storage nodes will replication the transaction only if they have the proof of the request lock being signed by all storage nodes.

Fixes #1373 

## Type of change

# How Has This Been Tested?

- Unit test
- Create multiple transaction in the same chain => no problem
- Create an invalid transaction send it and retry to validate it again after receiving the error => no problem, transaction still in error with the expected error
- Create an invalid transaction send it, update it and retry to validate it again after receiving the error => no problem, transaction still in error with the expected error
- Update the code so the node never unlock an address, and run the previous test => validation nodes return the error that the transaction is already locked with another data
- Tested with high usage with benchmark => no problem

# Checklist:

### DIFF
--- a/lib/archethic/mining.ex
+++ b/lib/archethic/mining.ex
@@ -18,7 +18,9 @@ defmodule Archethic.Mining do
 
   alias Archethic.P2P
   alias Archethic.P2P.Message
+  alias Archethic.P2P.Message.Ok
   alias Archethic.P2P.Message.ReplicationError
+  alias Archethic.P2P.Message.RequestChainLock
 
   alias Archethic.TransactionChain.Transaction
   alias Archethic.TransactionChain.Transaction.CrossValidationStamp
@@ -29,6 +31,8 @@ defmodule Archethic.Mining do
   use Retry
 
   @protocol_version 5
+
+  @lock_threshold 0.75
 
   def protocol_version, do: @protocol_version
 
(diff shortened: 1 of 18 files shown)


lib/archethic/mining.ex
```diff
@@ -96,6 +100,47 @@ defmodule Archethic.Mining do
...
+      |> Transaction.to_pending()
+      |> Transaction.serialize()
+      |> Crypto.hash()
+
+    message = %RequestChainLock{address: address, hash: hash}
+
+    aggregated_responses =
+      Task.Supervisor.async_stream_nolink(
+        Archethic.TaskSupervisor,
+        storage_nodes,
```
samuelmanzanera, line 122: I wondering if we should use locally available nodes, to save some times in the StartMining, as authorized and globally available nodes might not respond due to unavailability

Neylix: Done

bchamagne (approved):

PR #1398, opened 2024-01-29
Enhance utxo with consumed inputs

samuelmanzanera:
# Description

This PR improves the UTXO model for the AEIP21 consumed inputs logic.

## Type of change

- Enhancements

# Checklist:

### DIFF
--- a/lib/archethic/transaction_chain/transaction/validation_stamp/ledger_operations.ex
+++ b/lib/archethic/transaction_chain/transaction/validation_stamp/ledger_operations.ex
@@ -309,6 +309,8 @@ defmodule Archethic.TransactionChain.Transaction.ValidationStamp.LedgerOperation
         |> get_inputs_to_consume(
           uco_to_spend,
           tokens_to_spend,
+          uco_balance,
+          tokens_balance,
           encoded_state,
           contract_context
         )
@@ -405,6 +407,8 @@ defmodule Archethic.TransactionChain.Transaction.ValidationStamp.LedgerOperation
          inputs,
          uco_to_spend,
          tokens_to_spend,
+         uco_balance,
+         tokens_balance,
          encoded_state,
          contract_context
        ) do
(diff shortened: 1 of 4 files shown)


lib/archethic/transaction_chain/transaction/validation_stamp/ledger_operations.ex
```diff
@@ -288,15 +288,25 @@ defmodule Archethic.TransactionChain.Transaction.ValidationStamp.LedgerOperation
...
+    %{uco: uco_balance, token: tokens_balance} = ledger_balances(consolidated_inputs)
     %{uco: uco_to_spend, token: tokens_to_spend} = total_to_spend(fee, movements)
 
     if sufficient_funds?(uco_balance, uco_to_spend, tokens_balance, tokens_to_spend) do
       consumed_utxos =
-        get_inputs_to_consume(
-          inputs,
+        consolidated_inputs
+        |> Enum.map(fn
+          utxo = %UnspentOutput{from: ^change_address} ->
```
Neylix, line 300: I think this burn address modification should be done directly when we create the utxo at `LedgerOperations.get_utxos_from_transaction`

lib/archethic/utxo/loader.ex
```diff
@@ -95,16 +96,14 @@ defmodule Archethic.UTXO.Loader do
...
          },
-         phase2? \\ false
+         transaction_address
        ) do
     # Filter unspent outputs which have been consumed and updated (required in the AEIP21 Phase 1)
-    Enum.filter(unspent_outputs, fn %UnspentOutput{type: type} ->
-      phase2? or Enum.any?(consumed_inputs, &(&1.type == type))
-    end)
+    unspent_outputs
+    |> Enum.filter(&match?(%UnspentOutput{from: ^transaction_address}, &1))
```
Neylix, line 106: Since new token are coming from burn address, you should also filter the utxo with from == burn address

lib/archethic/transaction_chain/transaction/validation_stamp/ledger_operations.ex
```diff
@@ -424,8 +420,8 @@ defmodule Archethic.TransactionChain.Transaction.ValidationStamp.LedgerOperation
       {{:token, token_address, token_id}, inputs} ->
         token_used?(tokens_to_spend, token_address, token_id) or consolidate_inputs?(inputs)
 
-      {:state, _} ->
-        encoded_state != nil
+      {:state, [%UnspentOutput{encoded_payload: previous_state}]} ->
+        encoded_state != previous_state
 
       {:call, inputs} ->
         case contract_context do
```
Neylix, line 427: It looks like there is an invalid filter function here, filter should receive true or false, but here it can receive nil, an utxo or an empty list.
Also, if the return is an empty list or an utxo, the filter will be true, and then we will consume all the inputs of type call, but we want to consume only the call that has been used.
So maybe the filter could be replaced by a flat_map directly and all the function would return a list with the utxo to consume.

Neylix (changes requested):

lib/archethic/transaction_chain/transaction/validation_stamp/ledger_operations.ex
samuelmanzanera: Not completely because we need them as unspent output with the change address after , if none are spent.

lib/archethic/transaction_chain/transaction/validation_stamp/ledger_operations.ex
samuelmanzanera: The state in this case can only exist with a single unspent output. It cannot be nil or empty list as there is only one state utxo. 
Hence the nil cannot be present as it's not the encoded state we use in the group by

lib/archethic/utxo/loader.ex
samuelmanzanera: New tokens are not coming from burn address, it's the reason to not cast them by default but use the burn only in the function to list consumed inputs

samuelmanzanera: New tokens are not coming from burn address, it's the reason to not cast them by default but use the burn only in the function to list consumed inputs

lib/archethic/transaction_chain/transaction/validation_stamp/ledger_operations.ex
Neylix: I was talking about call utxo, not state

samuelmanzanera: Ah... yes, indeed it's not well used. 
It can be improved accordingly.

lib/archethic/transaction_chain/transaction/validation_stamp/ledger_operations.ex
```diff
@@ -288,15 +288,25 @@ defmodule Archethic.TransactionChain.Transaction.ValidationStamp.LedgerOperation
...
       consumed_utxos =
-        get_inputs_to_consume(
-          inputs,
+        consolidated_inputs
+        |> Enum.map(fn
+          utxo = %UnspentOutput{from: ^change_address} ->
+            # As the minted tokens are used internally during transaction's validation
+            # and doesn't not exists outside, we use the burning address
+            # to identify inputs coming from the token's minting.
+            %{utxo | from: burning_address()}
```
bchamagne, line 304: why this choice? shouldn't this be current transaction address? (or a `:mint` from ?)

lib/archethic/transaction_chain/transaction/validation_stamp/ledger_operations.ex
```diff
@@ -412,39 +408,67 @@ defmodule Archethic.TransactionChain.Transaction.ValidationStamp.LedgerOperation
...
+
+        %UnspentOutput{encoded_payload: previous_state} ->
+          encoded_state != previous_state
+      end
+
+    if include_new_state? do
+      [
+        %UnspentOutput{
+          type: :state,
+          encoded_payload: encoded_state,
```
bchamagne, line 463: is this change compatible with phase 1? This remove the state utxo if it's the same, which means current worker won't be able to find it?

lib/archethic/transaction_chain/transaction/validation_stamp/ledger_operations.ex
samuelmanzanera: The intent was to differentiate between inputs which are coming from input transaction and some produced dynamically by the transaction (minting)

For instance, on EVM, when you create a contract you send transaction to 0x0000... and the explorer shows the destination as "Contract Creation", we could then adapt this kind of logic.

lib/archethic/transaction_chain/transaction/validation_stamp/ledger_operations.ex
samuelmanzanera: Indeed, it's not wrong.
For phase2, as state might be forwarded from the genesis to the worker pool it will be ok, but for phase1, it might not, as the state is part of the %Contract{} struct coming from the transaction details (including the state as UTXO).

Neylix (approved):

PR #1413, opened 2024-02-02
do not check for reward conversion on token transaction

bchamagne:
Wondered why I need to mock Reward module when I was working with tokens. 
No need to check if a token is a reward if the transaction type is token.


bchamagne:
Invalid because we may have transfers of MUCO inside a transaction of type token.

PR #1416, opened 2024-02-05
Forbid user do send fraction of non fungible token

Neylix:
# Description

Currently there is the possibility for a use to send fraction of a NFT token which should not be possible.
This PR adds a new control in the pending transaction validation to allow user to send NFT token only by unit (1e8).

## Type of change

# How Has This Been Tested?

Create a NFT token and try to send it by fraction. It is refused (same case using token mint recipient as well).

# Checklist:

### DIFF
--- a/lib/archethic/mining/pending_transaction_validation.ex
+++ b/lib/archethic/mining/pending_transaction_validation.ex
@@ -82,13 +82,14 @@ defmodule Archethic.Mining.PendingTransactionValidation do
       ) do
     start = System.monotonic_time()
 
-    with :ok <- validate_size(tx),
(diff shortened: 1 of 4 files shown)


bchamagne (approved):
LGTM 
Capture d’écran 2024-02-05 à 17 43 04

also tried to have a fraction of nft sent when creating the nft and it's also blocked (different error msg but it's fine)

samuelmanzanera (approved):
LGTM.
Evenif probably the order of validation rules in the pending transaction could be reviewed to provide a better `fail-fast ` approach before to enter business rules (i.e : transaction exists -> previous signature -> size, etc...)

PR #1428, opened 2024-02-14
Fix GetUnspentOutputs deserialization

bchamagne:


lib/archethic/p2p/message/get_unspent_outputs.ex
```diff
@@ -3,7 +3,7 @@ defmodule Archethic.P2P.Message.GetUnspentOutputs do
   Represents a message to request the list of unspent outputs from a transaction
   """
   @enforce_keys [:address]
-  defstruct [:address, offset: 0, limit: 0, genesis?: false]
+  defstruct [:address, offset: 0, genesis?: false]
```
Neylix, line 6: You should keep limit as it will surely be used in future graphql API as this request will become the new unspent output view after AEIP-21

Neylix (changes requested):

samuelmanzanera (approved):
LGTM

Neylix (approved):

PR #1445, opened 2024-02-28
AEIP-21-P2: New genesis node SC Worker

### LINKED ISSUE #1409: AEIP-21: Smart Contract Worker from Genesis Pool
To unify smart contract triggers and to avoid problem of I/O synchronization, the genesis pools need to load contract using the worker, and trigger new transaction based on the calls.

To distribute the load and execution, we can use the validation nodes election to pick the trigger node.
This way new triggers' node will be elected for new chain's transaction

Neylix:
# Description

Implement new Genesis Worker behavior

Fixes #1409 

## Type of change

# Checklist:

### DIFF
--- a/lib/archethic/contracts/loader.ex
+++ b/lib/archethic/contracts/loader.ex
@@ -12,6 +12,7 @@ defmodule Archethic.Contracts.Loader do
 
   alias Archethic.TransactionChain
   alias Archethic.TransactionChain.Transaction
+  alias Archethic.TransactionChain.TransactionData
   alias Archethic.TransactionChain.Transaction.ValidationStamp
   alias Archethic.TransactionChain.Transaction.ValidationStamp.LedgerOperations.UnspentOutput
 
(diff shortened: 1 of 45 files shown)


lib/archethic/contracts/loader.ex
```diff
@@ -42,90 +61,163 @@ defmodule Archethic.Contracts.Loader do
...
+    |> Enum.zip(recipients)
+    |> Enum.map(fn
+      {{:ok, {:ok, genesis_address}}, recipient} -> {recipient, genesis_address}
+      {_, recipient} -> {recipient, recipient}
+    end)
+  end
+
+  defp resolve_genesis_address(recipients, _, _), do: recipients
+
+  defp contract_contains_trigger?(%Contract{triggers: triggers}) do
```
samuelmanzanera, line 218: I think it can be interesting to add this function to the Contract module.
Also this will be easier to unit test

lib/archethic/contracts/worker.ex
```diff
@@ -40,10 +43,56 @@ defmodule Archethic.Contracts.Worker do
...
+  def stop(genesis_address) do
+    case genesis_address |> via_tuple() |> GenServer.whereis() do
+      nil ->
+        :ok
+
+      pid ->
+        Logger.info("Stop smart contract at #{Base.encode16(genesis_address)}")
+        DynamicSupervisor.terminate_child(ContractSupervisor, pid)
+    end
+  end
```
samuelmanzanera, line 78: I think it would be better to decouple worker to the supervisor which holds it.
Maybe we can use a WorkerSupervisor (like for Connection), which does the action to spawn and terminate worker, or let the loader doing it.

PR #1463, opened 2024-03-14
Wrap interpreter execution in a task to isolate calls

samuelmanzanera:
# Description

Fix smart contract execution to be run inside a task to avoid intepreter's process limitations

This way the code's execution is isolated and multiple run from condition, actions and after contract renewal will be independent of 
the worker being or not in the same process.

## Type of change

# Checklist:

### DIFF
--- a/lib/archethic/contracts.ex
+++ b/lib/archethic/contracts.ex
@@ -16,8 +16,10 @@ defmodule Archethic.Contracts do
   alias __MODULE__.Loader
   alias Archethic.Crypto
   alias Archethic.TransactionChain.Transaction
+  alias Archethic.TransactionChain.Transaction.ValidationStamp
   alias Archethic.TransactionChain.TransactionData
   alias Archethic.TransactionChain.TransactionData.Recipient
+  alias Archethic.Utils
 
   require Logger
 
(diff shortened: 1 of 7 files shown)


lib/archethic/contracts.ex
```diff
@@ -255,32 +239,53 @@ defmodule Archethic.Contracts do
...
-             error: "Missing condition",
-             user_friendly_error: "Missing condition",
-             logs: [],
-             stacktrace: []
-           }}
-        end
+    conditions
+    |> Map.get(condition_key)
+    |> do_execute_condition(condition_key, contract, transaction, datetime, maybe_recipient)
+  rescue
```
Neylix, line 250: I think this rescue is not usefull anymore since the error will be rescued in the async task

Neylix (approved):

lib/archethic/contracts.ex
samuelmanzanera: I was not sure about this also, I kept in case of.
I don't know before if the rescue was for function like ` Interpreter.get_named_action_constants`or `get_condition_constants`

samuelmanzanera:
Maybe you should also wrap the  async function (now delegated to the JobCache) in try catch as it was done on f218f38, because the job cache will crash if the

samuelmanzanera:
:+1:

PR #1493, opened 2024-04-05
Ensure enough funds during smart contract call validation

samuelmanzanera:
# Description

This PR aims to fix an issue to accept a smart contract call if the contract doesn't have funds for the transaction triggered by the contract based on the movements.

## Type of change

- Enhancements

# How Has This Been Tested?

Create a smart contract which send UCO once a trigger transaction is received.
Do not funds the contract with the UCO to send in the action;
If a client call the contract it should be rejected.
Fund more the contract to satisfy the UCO to send in the action
Resent the transaction from the client to call the contract, the contract's transaction should be validated and emitted.

# Checklist:

### DIFF
--- a/lib/archethic/contracts/interpreter.ex
+++ b/lib/archethic/contracts/interpreter.ex
@@ -181,16 +181,7 @@ defmodule Archethic.Contracts.Interpreter do
           })
 
         {maybe_next_tx, next_state} =
-          case version do
-            0 ->
-              {
-                Legacy.execute_trigger(trigger_code, constants, contract_tx),
-                State.empty()
-              }
-
-            _ ->
-              ActionInterpreter.execute(trigger_code, constants, contract_tx)
-          end
+          do_execute_trigger(version, trigger_code, constants, contract_tx)
 
         logs = Scope.read_global([:logs])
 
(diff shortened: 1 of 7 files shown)


Neylix (changes requested):
I think this kind of verification should not be done in the Interpreter module.
Interpreter is intended to execute a contract code and create an output transaction, but it should not have functional verification on other domain than contract code execution.
I guess this verification could be done in `ValidateSmartContractCall` message.

Moreover, the contract could create movement based on the inputs received by the trigger. So if this verification is done during trigger transaction validation, the UTXO balance of the contract will not contains the inputs of the trigger tx so the verification might fail. But this could work for the worker execution of the contract.

Other question more functional, is the responsibility to determine if the contract has enough fund assigned to the worker or to the consensus through validation nodes ?

samuelmanzanera:
> I think this kind of verification should not be done in the Interpreter module. Interpreter is intended to execute a contract code and create an output transaction, but it should not have functional verification on other domain than contract code execution. I guess this verification could be done in `ValidateSmartContractCall` message.
> 
> Moreover, the contract could create movement based on the inputs received by the trigger. So if this verification is done during trigger transaction validation, the UTXO balance of the contract will not contains the inputs of the trigger tx so the verification might fail. But this could work for the worker execution of the contract.
> 
> Other question more functional, is the responsibility to determine if the contract has enough fund assigned to the worker or to the consensus through validation nodes ?

For the location of the code, yes I can do it in the P2P message itself no problem.

For the functional aspect, because validation do not execute the contracts locally, we have to request the workers to give us information about the feasibility of this transaction, including enough funds to execute it based on the funds retrieved by the validation nodes. 
As we should not accept a transaction where the validation nodes can know if this will be invalid just after.

PR #1504, opened 2024-04-17
API: fix transaction_chain when the paging_address == genesis_address

bchamagne:
# Description

When querying a chain with paging_state = genesis we should act as if paging_state is nil

Example of broken behaviour (testnet data)
```
query {
 transactionChain(
    address: "00006AF4C3AE64D435DA162B27918D6BE3AF8610494E3021148709BBE11A5520334C",
    pagingAddress: "00006AF4C3AE64D435DA162B27918D6BE3AF8610494E3021148709BBE11A5520334C"
  ) {
		address
    
  }
}
```

This should reply the same thing as the following query

```
query {
 transactionChain(
    address: "00006AF4C3AE64D435DA162B27918D6BE3AF8610494E3021148709BBE11A5520334C"
  ) {
		address
  }
}
```

It currently return empty (instead of 2 txs)

## Type of change

# How Has This Been Tested?

Unit test + manual test

# Checklist:

### DIFF
--- a/lib/archethic/db/embedded_impl/chain_reader.ex
+++ b/lib/archethic/db/embedded_impl/chain_reader.ex
@@ -132,18 +132,8 @@ defmodule Archethic.DB.EmbeddedImpl.ChainReader do
     filepath = ChainWriter.chain_path(db_path, genesis_address)
 
     if File.exists?(filepath) do
-      fd = File.open!(filepath, [:binary, :read])
-
       {transactions, more?, paging_address} =
-        case Keyword.get(opts, :order, :asc) do
-          :asc ->
-            process_get_chain(fd, fields, opts, db_path)
-
-          :desc ->
-            process_get_chain_desc(fd, genesis_address, fields, opts, db_path)
-        end
-
-      File.close(fd)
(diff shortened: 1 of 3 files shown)


lib/archethic/transaction_chain.ex
```diff
@@ -970,8 +970,15 @@ defmodule Archethic.TransactionChain do
           | {:error, :not_in_local}
   def resolve_paging_state(_, nil, _), do: {:ok, nil}
 
-  def resolve_paging_state(_, paging_state, _) when is_binary(paging_state),
-    do: {:ok, paging_state}
+  def resolve_paging_state(last_chain_address, paging_state, _) when is_binary(paging_state) do
+    genesis_address = get_genesis_address(last_chain_address)
```
Neylix, line 974: The node using this function may not be a storage node of this chain so the genesis address may not be retrieved.
Maybe the best place do do this check is in the `ChainReader.get_transaction_chain` function

Neylix (changes requested):

Neylix (approved):

PR #1494, opened 2024-04-08
Bootstrap: Replicate missed transaction of current summary 

bchamagne:
# Description

When a node starts and is authorized, it'll catch up the missed transaction of current summary (current day)
Fixes #1468

## Type of change

- New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Unit tested 
Manual test of this scenario: 

1. 4 nodes running
2. close one
3. run a faucet
4. restart the one
5. check that faucet transaction is in db

# Checklist:

- My code follows the style guidelines of this project
- I have performed a self-review of my own code
- I have commented my code, particularly in hard-to-understand areas
- I have made corresponding changes to the documentation
- My changes generate no new warnings
- I have added tests that prove my fix is effective or that my feature works
- New and existing unit tests pass locally with my changes
- Any dependent changes have been merged and published in downstream modules

PR #1507, opened 2024-04-25
Also fetch the inputs when fetching the transactions

### LINKED ISSUE #1501: Inputs are not written on self repair & rebalancing
### Context
Since AEIP-21 the inputs are not sent to the transactions directly but to the genesis. 
The nodes provide a graphQL query that returns the inputs of a given transaction. 
To feed the db of inputs, when replicating a transaction, we write the inputs used when validating the transaction. 

### Issue
If a node missed a transaction by being offline, it'll never do a replication; instead it'll do a self-repair. But the self-repair mechanism does not write the inputs of each transaction. Resulting in empty results in the graphQL query.
This also happen when rebalancing the transactions when topology changes

The Self Repair should write the inputs of each transaction in DB. The difficult part is to determine the inputs of each transaction.

### Epic

_No response_

bchamagne:
# Description

Based on the branch of PR #1494  (only 2 commits for this one)
- [ ] Sam told me about a new function get_node_synchronized_before, I think I should use it

Fixes #1501

## Type of change

# How Has This Been Tested?

Unit tests
- [x] Manual scenario : 
  - start a new node 
  - after self repair, check with graphql that the inputs of transactions made before are present on the new node

# Checklist:

### DIFF
--- a/lib/archethic/self_repair/sync.ex
+++ b/lib/archethic/self_repair/sync.ex
@@ -26,6 +26,7 @@ defmodule Archethic.SelfRepair.Sync do
   }
 
   alias Archethic.BeaconChain.Subset.P2PSampling
+  alias Archethic.TransactionChain
   alias Archethic.TransactionChain.Transaction
   alias Archethic.TransactionChain.Transaction.ValidationStamp
   alias Archethic.TransactionChain.TransactionSummary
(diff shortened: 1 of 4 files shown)


lib/archethic/self_repair/sync.ex
```diff
@@ -466,16 +467,23 @@ defmodule Archethic.SelfRepair.Sync do
...
       TaskSupervisor,
       attestations,
-      fn attestation ->
+      fn attestation = %ReplicationAttestation{
+           transaction_summary: %TransactionSummary{address: address}
+         } ->
+        # todo: parallelize downloads
         tx = TransactionHandler.download_transaction(attestation, download_nodes)
+
+        inputs = TransactionChain.fetch_inputs(address, download_nodes)
```
Neylix, line 476: I think downloading data should be done by the TransactionHandler.
Also here you don't run the Election to fetch the inputs

lib/archethic/self_repair/sync.ex
```diff
@@ -466,16 +467,23 @@ defmodule Archethic.SelfRepair.Sync do
...
-        {consolidated_attestation, tx}
+        {consolidated_attestation, tx, inputs}
       end,
       max_concurrency: System.schedulers_online() * 2,
       timeout: Message.get_max_timeout() + 2000
     )
-    |> Stream.each(fn {:ok, {attestation, tx}} ->
+    |> Stream.each(fn {:ok, {attestation, tx = %Transaction{address: address}, inputs}} ->
       :ok = TransactionHandler.process_transaction(attestation, tx, download_nodes)
+      :ok = TransactionChain.write_inputs(address, inputs)
```
Neylix, line 486: Same for downloading, processing data should be done by the TransactionHandler

Neylix (changes requested):

Neylix (approved):

Issue #1509, opened 2024-05-06
Analyze: storage_nonce change


bchamagne:
### Describe the solution you'd like

The storage_nonce is known to anyone that have joined the network once. 
It is used to encode data that only nodes are supposed to know.

Should we change it? When should we change it? How can we change it?

PR #1512, opened 2024-05-14
Add Crypto.decrypt_with_storage_nonce/1 in Interpreter's library

samuelmanzanera:
# Description

Add `Crypto.decrypt_with_storage_nonce/1` in the Interpreter's library.

This feature allows node to decrypt secrets during the contract's execution.

## Type of change

# How Has This Been Tested?

- Unit tests
- Integration test with the following code:
```elixir
  @version 1
  actions triggered_by: interval, at: "* * * * * *" do
    cleartext = Crypto.decrypt_with_storage_nonce(0x...)
    Contract.set_content(cleartext)
  end
```

The ciphertext is generated using the `Crypto.ec_encrypt("data to encrypt", Crypto.storage_nonce_public_key())`
# Checklist:

### DIFF
--- a/lib/archethic/contracts/interpreter/library/common/crypto.ex
+++ b/lib/archethic/contracts/interpreter/library/common/crypto.ex
@@ -97,6 +97,19 @@ defmodule Archethic.Contracts.Interpreter.Library.Common.Crypto do
     end
   end
 
+  @spec decrypt_with_storage_nonce(binary()) :: binary()
+  def decrypt_with_storage_nonce(cipher) do
+    case cipher
+         |> UtilsInterpreter.maybe_decode_hex()
+         |> Crypto.ec_decrypt_with_storage_nonce() do
+      {:ok, data} ->
+        data
+
+      {:error, :decryption_failed} ->
+        raise Library.Error, message: "Unable to decrypt data with storage nonce"
+    end
+  end
+
   @spec check_types(atom(), list()) :: boolean()
   def check_types(:hash, [first]) do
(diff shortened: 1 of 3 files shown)


bchamagne (approved):

Neylix (commented):
I'm wondering how we could handle this function if the storage nonce change.
Currently it's not planned to update it (maybe with SGX ?).
Maybe we should anticipate this if possible

samuelmanzanera:
> I'm wondering how we could handle this function if the storage nonce change. Currently it's not planned to update it (maybe with SGX ?). Maybe we should anticipate this if possible

Indeed, the storage nonce was not aim to be changed, otherwise this will result as a lot of shard movements (aka new election), and will require all the smart contracts to change their secrets, etc...
But maybe somewhere we could try to imaginate this kind of scenario (for instance, hard fork or security breach), related #1509

Neylix:
> > I'm wondering how we could handle this function if the storage nonce change. Currently it's not planned to update it (maybe with SGX ?). Maybe we should anticipate this if possible
> 
> Indeed, the storage nonce was not aim to be changed, otherwise this will result as a lot of shard movements (aka new election), and will require all the smart contracts to change their secrets, etc... But maybe somewhere we could try to imaginate this kind of scenario (for instance, hard fork or security breach), related #1509

Yes, A simple solution could be to try to decrypt cipher with all the previous storage nonce, but only encrypt with the last one and request users to change their cipher if needed

Neylix (approved):

PR #1517, opened 2024-05-22
Metrics: add P2P.nodes_connected

bchamagne:
Add a metric to count the number of nodes currently connected. 
This will be useful to detect splits or nodes that are not connected to anyone but still accept incoming connections (so they appear available, but in the end they cannot request anything outside of its shard resulting in wrong replies)

### DIFF
--- a/lib/archethic/p2p.ex
+++ b/lib/archethic/p2p.ex
@@ -103,6 +103,26 @@ defmodule Archethic.P2P do
   end
 
   @doc """
+  Called by the telemetry poller
+  """
+  def nodes_connected_count() do
+    nodes_connected =
+      authorized_and_available_nodes()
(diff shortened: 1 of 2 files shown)


lib/archethic/p2p.ex
```diff
@@ -102,6 +102,26 @@ defmodule Archethic.P2P do
     end
   end
 
+  @doc """
+  Called by the telemetry poller
+  """
+  def nodes_connected_count() do
+    nodes_connected =
+      Archethic.P2P.authorized_and_available_nodes()
```
Neylix, line 110: Wondering if it should be `list_nodes` or `authorized_and_available_nodes`.
Also you can remove `Archethic.P2P`
```suggestion
      authorized_and_available_nodes()
```

samuelmanzanera (approved):

PR #1522, opened 2024-05-29
Fix the execution's errors when cache? is false

bchamagne:
made a mistake on https://github.com/archethic-foundation/archethic-node/pull/1521. The errors on conditions were not properly transformed in the %Failure{} struct when cache? was false.

I added the tests this time.

### DIFF
--- a/lib/archethic/contracts.ex
+++ b/lib/archethic/contracts.ex
@@ -571,7 +571,7 @@ defmodule Archethic.Contracts do
     end
 
     result =
-      if Keyword.get(opts, :cache?) do
+      if Keyword.fetch!(opts, :cache?) do
         # We set the maximum timeout for a transaction to be processed before the kill the cache
         Utils.JobCache.get!(key,
           function: func,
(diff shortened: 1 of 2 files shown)


lib/archethic/contracts.ex
```diff
@@ -560,26 +560,31 @@ defmodule Archethic.Contracts do
...
+          {:error, err, __STACKTRACE__}
       end
+    end
 
-      # We set the maximum timeout for a transaction to be processed before the kill the cache
-      case Utils.JobCache.get!(key, function: func, timeout: timeout, ttl: 60_000) do
-        {:error, err, stacktrace} -> {:error, raise_to_failure(err, stacktrace)}
-        result -> result
+    result =
+      if Keyword.get(opts, :cache?) do
```
Neylix, line 574: To avoid confusion, since this opts must be set, I suggest to use `Keyword.fetch!`
```suggestion
      if Keyword.fetch!(opts, :cache?) do
```

PR #1541, opened 2024-07-19
Integrate bls signature scheme

### LINKED ISSUE #1519: Support BLS signature scheme
To support aggregated signatures, we have to support BLS signature scheme.

We might rely on external dependency as Erlang does not support it yet.

We might have to use [BLS elixir ](https://hex.pm/packages/bls) or Rust library through Rustler

### LINKED ISSUE #1542: Update node's transaction to support BLS's public key publication

samuelmanzanera:
# Description

Add support of BLS mining public key

Note: I inserted a migration task which not run by default. 
This task should be run after the upgrade have been released.

Fixes #1519
Fixes #1542 

## Type of change

# How Has This Been Tested?

Test have been released with multiple nodes with upgrades of nodes.
In memory table should be updated.
Starting from develop and then to this branch. 

Run the migration script to create a new transaction with the mining public key

Restart tests have been made to make sure the node's transaction content decoding is backward compatible

Explorer should list the new mining public key from the node info and from the node's transaction content

# Checklist:


lib/archethic/p2p/node.ex
```diff
@@ -353,7 +283,8 @@ defmodule Archethic.P2P.Node do
       DateTime.to_unix(enrollment_date)::32, available_bin::1, synced_bin::1, authorized_bin::1,
       authorization_date::32, first_public_key::binary, last_public_key::binary,
       reward_address::binary, last_address::binary, origin_public_key::binary,
-      DateTime.to_unix(last_update_date)::32, DateTime.to_unix(availability_update)::32>>
+      DateTime.to_unix(last_update_date)::32, DateTime.to_unix(availability_update)::32,
+      mining_public_key::binary>>
```
Neylix, line 287: According to typespec, mining_public_key can be nil and you cannot encode nil as a binary.
You should have a check before to determine if the value is nil.
Same for deserialize function

lib/archethic/p2p/node.ex
```diff
@@ -123,8 +86,31 @@ defmodule Archethic.P2P.Node do
...
+        port,
+        http_port,
+        transport,
+        reward_address,
+        origin_public_key,
+        key_certificate
+      )::binary, mining_public_key::binary>>
+  end
+
+  @doc false
```
Neylix, line 113: I don't get why you keep the old encode function since it is not used anymore I believe

Neylix (changes requested):

lib/archethic/p2p/node.ex
samuelmanzanera: Indeed, at first see, it doesn't look relevant I will check, thx

lib/archethic/p2p/node.ex
```diff
@@ -348,12 +257,19 @@ defmodule Archethic.P2P.Node do
 
     avg_bin = trunc(average_availability * 100)
 
+    mining_public_key_bin =
+      case mining_public_key do
+        nil -> ""
+        _ -> mining_public_key
+      end
```
Neylix, line 264: Suggestion:
I think it would be more elegant and less error prone to set a true / false byte if the node has a mining key

```suggestion
    mining_public_key_bin =
      if is_nil(mining_public_key),
        do: <<0::8>>,
        else: <<1::8, mining_public_key::binary>>
```

lib/archethic/p2p/node.ex
```diff
@@ -435,6 +302,13 @@ defmodule Archethic.P2P.Node do
     {origin_public_key, <<last_update_date::32, availability_update::32, rest::bitstring>>} =
       Utils.deserialize_public_key(rest)
 
+    {mining_public_key, rest} =
+      try do
+        Utils.deserialize_public_key(rest)
+      rescue
+        _ -> {nil, rest}
+      end
```
Neylix, line 310: Using a true/false byte avoid to use a try rescue
```suggestion
    {mining_public_key, rest} =
      case rest do
        <<1::8, rest::bitstring>> -> Utils.deserialize_public_key(rest)
        <<0::8, rest::bitstring>> -> {nil, rest}
      end
```

PR #1557, opened 2024-08-26
Mining: check for unknown origin as part of the valid pending transaction

### LINKED ISSUE #1539: Incorrect handling of not found proof of work

When an incoming transaction has an invalid origin signature, the coordinator node does not retrieve the proof of work.
The proof of work returns an empty string. Then the validation stamp is serialized replacing the empty string with `<<0::32>>`.
Then the cross validators compare the proof of work they found (empty string) with the one in the validation stamp (`<<::32>>`) which does not match while all nodes did not found the POW.

Define a better workflow for invalid proof of work

### Epic

_No response_

bchamagne:
# Description

Differentiate an invalid proof of work from an unknown origin.
Fixes #1539

## Type of change

- Improvment

# How Has This Been Tested?

I modified the CLI to send invalid origin signature. Then I sent a transaction with the CLI
Capture d’écran 2024-08-26 à 16 50 51

# Checklist:


lib/archethic/mining/validation_context.ex
```diff
@@ -1137,7 +1154,8 @@ defmodule Archethic.Mining.ValidationContext do
          transaction: tx
        }) do
     case pow do
-      "" ->
+      <<0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
```
samuelmanzanera, line 1157: Suggestion: you could probably shorten this as: `<<0::16, 0::256>>`  or even `<<0::272>>` even if the latter might a bit more confusing in terms of understanding.

lib/archethic/mining/pending_transaction_validation.ex
```diff
@@ -139,6 +141,18 @@ defmodule Archethic.Mining.PendingTransactionValidation do
     end
   end
 
+  defp valid_origin_signature(tx = %Transaction{}) do
```
samuelmanzanera, line 144: Question: I don't get why this check is done in pending transaction validation. The purpose of the proof of work is to be done by coordinator mainly. The case where the proof of work is done by cross validation nodes is when the proof of work is empty. Maybe the logic should be the same

lib/archethic/mining/validation_context.ex
```diff
@@ -1137,7 +1154,8 @@ defmodule Archethic.Mining.ValidationContext do
...
+        0, 0, 0, 0>> ->
```
samuelmanzanera, line 1158: Suggestion: Probably a check with the context's error might be useful, for example if the error is "Unknown origin" and the pow is empty, then the check is :ok.
And otherwise (proof of work empty and not errors as "Unknown origin", the proof of work should be redone.

samuelmanzanera (changes requested):

lib/archethic/mining/validation_context.ex
```diff
@@ -1146,6 +1167,10 @@ defmodule Archethic.Mining.ValidationContext do
     end
   end
 
+  def valid_stamp_origin_signature?(%ValidationStamp{proof_of_work: pow}, %__MODULE__{}) do
```
samuelmanzanera, line 1170: Maybe it's not needed as the L1123 is commented.

PR #1569, opened 2024-10-08
Replicate previous chain raise on missed tx

### LINKED ISSUE #1563: Chain replication does not fail on error

When a node replicate a transaction, it also replicate the missing previous transactions of the entire chain.
But as the retrieval of the transaction is done through a Stream, if there is a connection issue the node may not retrieve the previous transaction and continue the process of replication.
In the current state, this lead to an error as the current transaction is replicated under a wrong genesis address and broke the chain.

During replication, if we detect that the node needs to replicate previous transaction in the chain, if the stream does not returns the expected transaction, the replication should fail.

### Epic

_No response_

Neylix:
# Description

During replication, if a node has to sync previous transactions of a chain, when it downloads them it ensure all transactions are fetched from the paging address to the limit address.
It ensure all transactions using there previous address which should be the address of the previous transaction fetched.

Fixes #1563 

## Type of change

# Checklist:

### DIFF
--- a/lib/archethic/replication/transaction_context.ex
+++ b/lib/archethic/replication/transaction_context.ex
@@ -66,6 +66,7 @@ defmodule Archethic.Replication.TransactionContext do
         genesis_address
         |> TransactionChain.fetch(storage_nodes, paging_state: paging_address)
         |> Stream.take_while(&(Transaction.previous_address(&1) != limit_address))
+        |> ensure_all_tx_fetched(paging_address, genesis_address, limit_address)
     end
   end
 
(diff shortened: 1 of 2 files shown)


samuelmanzanera (approved):
LGTM

samuelmanzanera:
I'm wondering if we can even improve the security of this flow, by checking the proof of integrity of the transaction chain in the stream processing.

bchamagne (approved):

Neylix:
> I'm wondering if we can even improve the security of this flow, by checking the proof of integrity of the transaction chain in the stream processing.

As discussed, we are planning to evolve the proof of integrity so that it can be validated without the need of the previous transaction data. Combined with the proof of validation using BLS signatures, all the proofs could be validated in the message request before entering the replication process.

Issue #1553, opened 2024-08-21
AEWeb ssl certificate does not work on multiple / wildcard domain name


Neylix:
### Describe the problem you discovered

Currently when an aeweb reference transaction contains an ssl certificate, the node ensure the host domain name is certified by the certificate. To do so it search for the subjectAltName and take the last one.
It works well if the certificate only certify one domain name, but it fail if there is a wildcard or multiple domain name certified.

For exemple the domain name `guide.ucopourlesnuls.fr` will not be found as the node only take the last one which is `dashboard.ucopourlesnuls.fr`

![image](https://github.com/user-attachments/assets/da944dc9-e61f-46eb-8da3-ee82aeffd4d9)

Same for wildcard since we compare 2 strings.
`guide.ucopourlesnuls.fr` != `dashboard.ucopourlesnuls.fr` != `*.ucopourlesnuls.fr`

### Describe the solution you'd like

The check for the domain name should be more clever and find the domain name even with mutiple domain certified or a wildcard.

PR #1585, opened 2024-10-25
Fix AEWeb SSL certs with multi wildcard domains (#1553)

wassimans:
# Description

This PR addresses an issue with the SSL certificate verification logic in the node. Previously, the node validated the SSL certificate for a host domain by comparing it only to the last entry in the list of _subjectAltName_ entries within the certificate. This approach neglected other domain entries and did not consider wildcard domains.

The fix introduces enhanced matching logic that:

- Compares the host name against the entire list (if present) of _subjectAltNames_ within the certificate.

- Supports wildcard matching for more flexible and complete SSL verification.

This improvement ensures that requests to certified subdomains and wildcard domains are handled correctly according to the SSL certificate’s full set of authorized domains.

Fixes # (1553)

## Type of change

# How Has This Been Tested?

The solution has been tested by implementing and verifying the following test cases:

- Matching certified domains: Ensures that domains listed in the _subjectAltNames_ (including exact matches and wildcard patterns) are verified correctly.
- Non-certified domains: Confirms that non-certified domains are rejected, falling back to default SSL configurations when appropriate.

# Checklist:


lib/archethic_web/aeweb/domain.ex
```diff
@@ -56,9 +58,9 @@ defmodule ArchethicWeb.AEWeb.Domain do
             ownerships: [ownership = %Ownership{secret: secret} | _]
           }} <- ReferenceTransaction.fetch_last(tx_address),
          {:ok, cert_pem} <- Map.fetch(json_content, "sslCertificate"),
-         %{extensions: extensions} <- EasySSL.parse_pem(cert_pem),
-         {:ok, san} <- Map.fetch(extensions, :subjectAltName),
-         ^domain <- String.split(san, ":") |> List.last(),
+         all_domain_names <-
+           get_all_domain_names(cert_pem),
```
Neylix, line 62: The function `get_all_domain_names` could be simplier using only EasySSL:
```suggestion
         %{all_domains: all_domain_names} <- EasySSL.parse_pem(cert_pem, all_domains: true),
```

test/archethic_web/aeweb/domain_test.exs
```diff
@@ -0,0 +1,233 @@
...
+
+      result = Domain.sni("example.com")
+
+      expected_key = read_pem(fake_key_pem) |> hd()
+      expected_cert = read_pem(fake_cert_pem) |> hd() |> elem(1)
+      assert [key: expected_key, cert: expected_cert] == result
+
+      result = Domain.sni("blog.example.com")
+
+      assert [key: expected_key, cert: expected_cert] == result
```
Neylix, line 167: Here you should add another test to ensure the wildcard (*.example.com) is working well

Neylix (changes requested):

lib/archethic_web/aeweb/domain.ex
wassimans: But we need to parse cert_pem to der format before giving it to EasySSL.parse_pem/2.

Neylix: EasySSL expose 2 functions: `parse_pem` and `parse_der`. Using the first one does not require to parse the cert to der format

test/archethic_web/aeweb/domain_test.exs
```diff
@@ -0,0 +1,237 @@
...
+
+      result = Domain.sni("*.example.com")
```
Neylix, line 169: "*.example.com" is not a valid uri.
You should use something like "unknown.example.com"

lib/archethic_web/aeweb/domain.ex
```diff
@@ -56,9 +58,10 @@ defmodule ArchethicWeb.AEWeb.Domain do
             ownerships: [ownership = %Ownership{secret: secret} | _]
           }} <- ReferenceTransaction.fetch_last(tx_address),
          {:ok, cert_pem} <- Map.fetch(json_content, "sslCertificate"),
-         %{extensions: extensions} <- EasySSL.parse_pem(cert_pem),
-         {:ok, san} <- Map.fetch(extensions, :subjectAltName),
-         ^domain <- String.split(san, ":") |> List.last(),
+         [{:Certificate, certificate_der, _}] <- :public_key.pem_decode(cert_pem),
+         %{all_domains: all_domain_names} <-
+           EasySSL.parse_der(certificate_der, all_domains: true),
```
Neylix, line 63: You could directly use `EasySSL.pem_decode/2` instead of converting the pem to der format

Neylix (approved):

PR #1588, opened 2024-11-04
Add the genesis address in the validation_stamp

### LINKED ISSUE #1562: Add genesis address in Validation Stamp

Since [AEIP-21](https://github.com/archethic-foundation/aeip/blob/main/AEIP-21.md), genesis address became a major information used in many places.
Currently the genesis address of a transaction is retrieved from P2P but there is no proof if it's the real genesis address.
Also in term of code, this require to pass the genesis address as a parameter in a lot of functions.

We should add the genesis address of the transaction chain in the validation stamp. So all validation nodes will ensure this is the good one, and then whenever this transaction is used, nodes can have the proof of the validity of the genesis address.

### Additional context

In term of DB storage, there is a specific encoding system for all transaction field and we could avoid to encode the genesis address since the genesis is the name of the file in DB. So when decoding we could use the filename to fill the genesis address of the stamp.

### Epic

_No response_

bchamagne:
# Description

Added the genesis_address in the validation_stamp. It's written on disk only on IO transaction, because the chain storage is written in a file named by the genesis, so there's no point to duplicate the data there.
There's a migration script to run on all legacy IOs

- [ ] Update migration version when decided which version it'll be

Fixes #1562

## Type of change

# How Has This Been Tested?

Updated the code to store every tx also as IO then run a release upgrade on 4 nodes.
Chain & IO storage both returned the genesis address on legacy and new transactions

# Checklist:

### DIFF
--- a/lib/archethic/mining.ex
+++ b/lib/archethic/mining.ex
@@ -34,7 +34,8 @@ defmodule Archethic.Mining do
   # version 5->6 the POI changed and is now done with tx.data.recipients.args serialized with :extended mode
   # version 6->7 add Add consumed inputs in tx.validation_stamp.ledger_operations
   # version 7->8 movement resolved address are now the genesis address of the destination
-  @protocol_version 8
+  # version 8->9 genesis in the validation stamp
+  @protocol_version 9
 
   @lock_threshold 0.75
 
--- a/lib/archethic/bootstrap/network_init.ex
+++ b/lib/archethic/bootstrap/network_init.ex
@@ -200,6 +200,7 @@ defmodule Archethic.Bootstrap.NetworkInit do
 
     validation_stamp =
       %ValidationStamp{
+        genesis_address: Transaction.previous_address(tx),
         protocol_version: 1,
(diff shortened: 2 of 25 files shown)


Neylix (changes requested):
I think you could also update the grapphql schemas and the transaction details on the explorer

lib/archethic/mining.ex
```diff
@@ -34,7 +34,8 @@ defmodule Archethic.Mining do
   # version 5->6 the POI changed and is now done with tx.data.recipients.args serialized with :extended mode
   # version 6->7 add Add consumed inputs in tx.validation_stamp.ledger_operations
   # version 7->8 movement resolved address are now the genesis address of the destination
-  @protocol_version 8
+  # version 8->9 genesis in the validation stamp
+  @protocol_version 9
```
bchamagne, line 38: remainder: both this and https://github.com/archethic-foundation/archethic-node/pull/1591 use v9 so they need to go in same release

PR #1590, opened 2024-11-05
Add timestamp in node validation election

### LINKED ISSUE #1564: Add timestamp in the validation node election

Currently the validation node election is based on transaction data and the daily nonce seed. Therefore if a transaction fail to be validated and the exact same transaction is sent again the validation nodes will be the same until the daily nonce seed change.
But is one of the validation nodes (mainly the coordinator) has a wrong state or cannot retrieve some data, the stamp will be wrong and the transaction cannot be validated for the entire day.

To avoid this issue, we can add a timestamp in the election so the elected nodes will be always different.
The timestamp used could be the one used by the welcome node which is sent in the StartMining message (the ref timestamp)

### Additional context

_No response_

### Epic

_No response_

wassimans:
# Description

This pull request addresses an issue with the validation node election process. Currently, the election is based on transaction data and a daily nonce seed, which causes repeated selection of the same validation nodes for identical transactions throughout the day. If one of the validation nodes, especially the coordinator, has an incorrect state or is unable to retrieve required data, the transaction will fail validation for the entire day.

The solution implemented here introduces a timestamp into the election process to ensure that the elected nodes vary with each transaction attempt. The timestamp used is from the StartMining message, sent by the welcome node, which serves as a reference timestamp to diversify node election.

Fixes #1564

## Type of change

# How Has This Been Tested?

The following tests were conducted to validate this change:

- Ensuring the seed sorting value changes with changing transactions, and/or ref_timestamp
- Ensuring the seed sorting value doesn't change for identical parameters (transaction, ref_timestamp)
- Ensuring that validation nodes election changes with changing transactions and/or ref_timestamp
- Ensuring that validation nodes election doesn't change for identical transactions and ref_timestamp

# Checklist:

### DIFF
--- a/lib/archethic/mining.ex
+++ b/lib/archethic/mining.ex
@@ -80,11 +80,13 @@ defmodule Archethic.Mining do
   @doc """
   Elect validation nodes for a transaction
   """
-  def get_validation_nodes(tx = %Transaction{address: tx_address, validation_stamp: nil}) do
-    current_date = DateTime.utc_now()
-    sorting_seed = Election.validation_nodes_election_seed_sorting(tx, current_date)
+  def get_validation_nodes(
+        tx = %Transaction{address: tx_address, validation_stamp: nil},
+        ref_timestamp
+      ) do
+    sorting_seed = Election.validation_nodes_election_seed_sorting(tx, ref_timestamp)
 
-    node_list = P2P.authorized_and_available_nodes(current_date)
+    node_list = P2P.authorized_and_available_nodes(ref_timestamp)
 
     storage_nodes = Election.chain_storage_nodes(tx_address, node_list)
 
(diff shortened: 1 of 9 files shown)


lib/archethic/mining.ex
```diff
@@ -80,9 +80,12 @@ defmodule Archethic.Mining do
...
-  def get_validation_nodes(tx = %Transaction{address: tx_address, validation_stamp: nil}) do
+  def get_validation_nodes(
+        tx = %Transaction{address: tx_address, validation_stamp: nil},
+        ref_timestamp
+      ) do
     current_date = DateTime.utc_now()
-    sorting_seed = Election.validation_nodes_election_seed_sorting(tx, current_date)
+    sorting_seed = Election.validation_nodes_election_seed_sorting(tx, ref_timestamp)
 
     node_list = P2P.authorized_and_available_nodes(current_date)
```
bchamagne, line 90: Suggestion: you may remove `current_date` since it's the default parameters for `authorized_and_available_nodes`

bchamagne (approved):
LGTM

Neylix (changes requested):
I'm also wondering, in the ValidationContext, when a node cross validate the stamp, it ensure the timestamp is valid. This function could evolve since the timestamp should be the ref_timestamp shared by all validations node

wassimans:
> I'm also wondering, in the ValidationContext, when a node cross validate the stamp, it ensure the timestamp is valid. This function could evolve since the timestamp should be the ref_timestamp shared by all validations node

I now check for **equality** between the node's validation time and the corss validation time.

For consistency, I also modified the following test/archethic/mining/validation_context_test.exs:252, from "should validate even if validation_time and cross_validation_time are in different oracle bucket" to "should not validate if validation_time and cross_validation_time are in different oracle bucket" and made sure ":timestamp" is present in the inconsistencies table.

The test in test/archethic/mining/validation_context_test.exs:252 ("should validate with a valid validation stamp"), should handle the happy path for us without change.

Neylix (approved):

PR #1593, opened 2024-11-07
Consume the MUCO correctly

bchamagne:
# Description

A bug was introduced in 1.5.11's refactor where the UCO were consumed instead of the MUCO. This is an incorrect behaviour and people with MUCO were consuming UCO instead of MUCO when transferring MUCO.
This fix the issue by resolving MUCO only after the consume step.

## Type of change

# How Has This Been Tested?

Added a unit test

# Checklist:

### DIFF
--- a/lib/archethic/mining/validation_context.ex
+++ b/lib/archethic/mining/validation_context.ex
@@ -1139,7 +1139,9 @@ defmodule Archethic.Mining.ValidationContext do
       proof_of_integrity: fn -> valid_stamp_proof_of_integrity?(stamp, context) end,
       proof_of_election: fn -> valid_stamp_proof_of_election?(stamp, context) end,
       transaction_fee: fn -> valid_stamp_fee?(stamp, fee) end,
-      transaction_movements: fn -> valid_stamp_transaction_movements?(stamp, context) end,
+      transaction_movements: fn ->
+        valid_stamp_transaction_movements?(stamp, ledger_operations)
+      end,
       recipients: fn -> valid_stamp_recipients?(stamp, context) end,
       consumed_inputs: fn -> valid_consumed_inputs?(stamp, ledger_operations) end,
       unspent_outputs: fn -> valid_stamp_unspent_outputs?(stamp, ledger_operations) end,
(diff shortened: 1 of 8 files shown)


test/archethic/mining/validation_context_test.exs
```diff
@@ -158,6 +159,83 @@ defmodule Archethic.Mining.ValidationContextTest do
...
+            %UnspentOutput{
+              from: random_address(),
+              amount: 10,
+              type: {:token, muco_addr2, 0},
+              timestamp: timestamp
+            }
+          ]
+        )
+        | transaction:
+            Transaction.new(
```
Neylix, line 192: You should use the function `TransactionFactory.create_non_valided_transaction` and pass it in the opts field of the create_context function as you did for utxo

test/archethic/mining/validation_context_test.exs
```diff
@@ -158,6 +159,83 @@ defmodule Archethic.Mining.ValidationContextTest do
...
+      muco_addr1 = random_address()
+      muco_addr2 = random_address()
+
+      start_supervised!(RewardTokens)
+      RewardTokens.add_reward_token_address(muco_addr1)
+      RewardTokens.add_reward_token_address(muco_addr2)
+
+      validation_context = %ValidationContext{
+        create_context(timestamp,
+          utxos: [
```
Neylix, line 176: To ensure UCO are not used, you could add UCO utxo and ensure they remain the same (or less the fees)

lib/archethic/mining/validation_context.ex
```diff
@@ -800,17 +800,17 @@ defmodule Archethic.Mining.ValidationContext do
     protocol_version = Mining.protocol_version()
 
     ops =
-      %LedgerValidation{fee: fee}
+      %LedgerValidation{fee: fee, transaction_movements: movements}
```
Neylix, line 803: I think to keep a workflow like previously and to avoid errors, I think the movements could be passed in arguments to the function `validate_sufficient_funds` which will add them in the struct. Also we could set the `transaction_movements` field of the LedgerValidation struct therefore the `build_resolved_movements` function could have a guard to ensure the movements is a list. This could avoid doing misstake in ordering the functions (You did not update the order in TransactionFactory and no error is raised)

Neylix (changes requested):

PR #1603, opened 2024-11-21
Bls proof of replication

### LINKED ISSUE #1544: Leverage BLS aggregated signatures for replication's signatures

Neylix:
# Description

Handle new BLS replication signatures
- Add new ProofOfReplication module that create and validate the replication aggregated signatures
- A threshold of 75% of replication nodes is expected to consider the proof valid
- Once the proof of validation is created, validation nodes request replication nodes to create the replication signature which signs the transaction summary
- Replication signatures are sent asynchronously allowing validation nodes to create the proof as soon as the threshold is reached
- Coordinator node is responsible to create the proof of replication and share it to other validation nodes

Fixes #1544 and #1545

## Type of change

# Checklist:

### DIFF
--- a/lib/archethic/replication/transaction_pool.ex
+++ b/lib/archethic/replication/transaction_pool.ex
@@ -6,7 +6,10 @@ defmodule Archethic.Replication.TransactionPool do
 
   require Logger
 
+  alias Archethic.Crypto
+
   alias Archethic.TransactionChain.Transaction
+  alias Archethic.TransactionChain.Transaction.ProofOfValidation
 
   alias Archethic.TransactionChain.Transaction.ValidationStamp.LedgerOperations.VersionedUnspentOutput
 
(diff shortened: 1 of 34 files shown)


lib/archethic/replication/transaction_pool.ex
```diff
@@ -27,15 +30,31 @@ defmodule Archethic.Replication.TransactionPool do
...
+  end
+
   @doc """
   Dequeue the transaction and the inputs for the given transaction's address
   """
-  @spec pop_transaction(GenServer.server(), binary()) ::
+  @spec get_transaction(
+          GenServer.server(),
+          address :: Crypto.prepended_hash(),
+          opts :: Keyword.t() | [get_transaction?: boolean()]
```
samuelmanzanera, line 51: I think you mean `[pop_transaction?: boolean]`.
Suggestion: Probably this would be clearer to keep the functions separated, like: `Map.get` `Map.pop` to avoid options

samuelmanzanera (commented):
LGTM

PR #1641, opened 2025-01-20
Chralu/issue1638

### LINKED ISSUE #1638: Use genesis address stored in Tx.validationStamp

Since v1.7.0, Genesis address is available in the validation stamp.

We should refactor to remove useless `fetch_genesis_address` requests when the information is already available.

‼️ Careful ‼️ During the validation process, incoming Tx validation stamp should not be trusted. In that particular case, `fetch_genesis_address` request is legitimate.

Chralu:
# Description

Fixes #1638


lib/archethic/transaction_chain.ex
```diff
@@ -1440,10 +1440,11 @@ defmodule Archethic.TransactionChain do
...
   @spec first_transaction?(Transaction.t()) :: boolean()
-  def first_transaction?(
-        tx = %Transaction{validation_stamp: %ValidationStamp{proof_of_integrity: poi}}
-      ) do
-    poi == proof_of_integrity([tx])
+  def first_transaction?(%Transaction{
+        address: address,
+        validation_stamp: %ValidationStamp{genesis_address: genesis_address}
+      }) do
+    address == genesis_address
```
Neylix, line 1447: By definition a genesis address is not a transaction, so this equality will always fail.
You could do
```elixir
def first_transaction?(tx = %Transaction{
        validation_stamp: %ValidationStamp{genesis_address: genesis_address}
      }) do
  previous_address(tx) == genesis_address
end
```

Neylix (changes requested):

PR #1651, opened 2025-01-30
Tx replication sometimes fails. #1650

### LINKED ISSUE #1650: Tx replication sometimes fails.

Tx replication process can be started multiple times asynchronously. 

The transaction local existence is checked by the replication process. This might cause a single Tx to be concurrently replicated multiple times.

## Prevent multiple writes
`ChainWriter` write operations are sequentially executed. We could move the Tx presence check in `ChainWriter.write` to ensure a Tx is not written multiple times.

## Do not interrupt chain synchronisation on `:transaction_already_exists` error

`Archethic.Replication.sync_transaction_chain` should continue when `ChainWriter.write` returns a `: transaction_already_exists` error.

### Epic

_No response_

Chralu:
# Description

When replicating transactions, checks if Tx is already present before writing it.

Check is performed in `ChainWriter.write` to ensure they are sequentials.

Fixes #1650

### DIFF
--- a/lib/archethic/db.ex
+++ b/lib/archethic/db.ex
@@ -33,7 +33,8 @@ defmodule Archethic.DB do
                 order: :asc | :desc
               ]
             ) :: Enumerable.t()
-  @callback write_transaction(Transaction.t(), storage_type()) :: :ok
+  @callback write_transaction(Transaction.t(), storage_type()) ::
+              :ok | {:error, :transaction_already_exists}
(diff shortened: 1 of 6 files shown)


Neylix (changes requested):
You ensured the transaction has been succesfully written before ingesting it only when writting the previous transactions of the chain. But in there is 2 others call of `write_transaction` that need the same check

Neylix (approved):

